PR #9, opened 2015-01-09
Fix check for mark_out.mask in starter_cmp_conn()

maksqwe:


tobiasbrunner:
Thanks, applied to master

PR #10, opened 2015-03-10
Python 3 support

sbjorn:
This adds python 3 support and some basic unit tests of the protocol layer.

All values of keys and list items of deserialized messages are kept as `bytes` - no attempt of interpretation is done at the protocol level.

Tested on python 2.7, 3.2, 3.3, and 3.4.

To run the tests, make sure you have `pytest` installed;
```bash
$ # pip install pytest
$ py.test
```


martinwilli:
Integrated to the vici-python branch, thanks.

PR #12, opened 2015-06-23
capabilities: Handle ERANGE in user and group lookups

ebroder:
As it turns out, getpwnam_r, getgrnam_r, and friends will return ERANGE if _any_ user or group on the system is larger than will fit into the scratch buffer you pass to them.

This reworks the resolve_uid and resolve_gid methods plus init_supplementary_groups to use a variable-size buffer that is grown until the results fit.

I ran into this, not because the group that strongswan was dropping to was had a large membership list, but because a random other group did. As it stands currently, it's a pretty non-intuitive failure, since all you get in the output is "invalid uid/gid".

### DIFF
--- a/src/libstrongswan/utils/capabilities.c
+++ b/src/libstrongswan/utils/capabilities.c
@@ -288,13 +288,29 @@ METHOD(capabilities_t, resolve_uid, bool,
 
 #ifdef HAVE_GETPWNAM_R
 	struct passwd passwd;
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks for the patch. I reworked it a bit and fixed the case where the user/group name is not found (the functions return 0 but the result pointer will be NULL).

PR #13, opened 2015-08-22
vici: Handle closed sockets in the Ruby gem

ebroder:
From recvfrom(2) (which UDPSocket#recv backs into):

  The return value will be 0 when the peer has performed an orderly
  shutdown.

(i.e. it will return an empty string)

Previously in this scenario, Vici::Transport#recv_all would spin
forever trying to pull more data off the socket. I'm not entirely
clear what happened that caused strongswan to shutdown the socket, but
it probably should cause vici Ruby apps to spin.


tobiasbrunner:
Applied, thanks!

PR #16, opened 2015-11-06
VICI: Add NAT-T information on list_ike()

rbgarga:
Add 'local-nat-t' and 'remote-nat-t' keys to list_ike() to inform if
local or remote are using NAT-T

Obtained-from:	pfSense
Sponsored-by:	Rubicon Communications (Netgate)


tobiasbrunner:
Thanks. Pushed a slightly modified version to master.

PR #18, opened 2015-11-07
Add option to allow millisecond resolution in file logger

ttroy50:
This adds an option use_millisecond to the file logger configuration. If set the millisecond will be added to the end of the time_format with a format .XXX. 

The actual use of this only makes sense if time_format ends with %S e.g. %H:%M:%S or %T. Either of these would result in a time output of something like 12:52:25.123

### DIFF
--- a/src/conftest/conftest.c
+++ b/src/conftest/conftest.c
@@ -382,15 +382,17 @@ static void load_log_levels(file_logger_t *logger, char *section)
  */
 static void load_logger_options(file_logger_t *logger, char *section)
 {
-	bool ike_name;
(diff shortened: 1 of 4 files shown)


tobiasbrunner:
Good idea. I pushed a couple of modified patches to the _gh-18-file-logger-ms_ branch. Let me know if that's OK for you.

ttroy:
That's fine. Your changes seem to be functionally the same as mine just a change in the naming scheme.

ttroy50:
Sorry, just realised the last comment was from my other account.

tobiasbrunner:
OK, thanks.

PR #19, opened 2015-11-11
Radius retry timer config

ttroy50:
This change makes the RADIUS retransmissions configurable in the radius plugin.

It creates the following settings 

num_request_attempts = 4
first_request_timeout = 2000
request_backoff_timeout = 1000

This recreates the current default behaviour to preserve backwards compatibility.  

num_request_attempts is the total number of requests that will be sent.
first_request_timeout. The timeout for the first radius request in ms
request_backoff_timeout. The timeout for calculating subsequent attempt backoffs in ms, calculated as first_request_timeout + (retransmissions_attempt * request_backoff_timeout).

Using the default settings the requests could be sent as:

first @ 0
second @ 2s
third @ 5s
fourth @ 9s
final timeout @ 14s

### DIFF
--- a/conf/plugins/eap-radius.opt
+++ b/conf/plugins/eap-radius.opt
@@ -80,6 +80,15 @@ charon.plugins.eap-radius.secret =
 charon.plugins.eap-radius.server =
 	IP/Hostname of RADIUS server.
 
+charon.plugins.eap-radius.retransmit_base = 1.4
+	Base to use for calculating exponential back off.
+
+charon.plugins.eap-radius.retransmit_timeout = 2
+	Timeout in seconds before sending first retransmit.
+
+charon.plugins.eap-radius.retransmit_tries = 4
+	Number of times to retransmit a packet before giving up.
+
 charon.plugins.eap-radius.servers {}
 	Section to specify multiple RADIUS servers.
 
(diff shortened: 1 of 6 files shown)


tobiasbrunner:
I agree with making this configurable. But I wonder if it wouldn't be better and more consistent to use the same configuration parameters and algorithm we use for IKE (still configurable for each RADIUS server separately, of course).

That is, _retransmit_tries_ (int), _retransmit_timeout_ (double) and _retransmit_base_ (double). The timeout in milliseconds for each try would be calculated as `timeout = 1000.0 * retransmit_timeout * pow(retransmit_base, try)`. While the current behavior could not exactly be replicated (which I don't think is essential) the following values come pretty close `retransmit_tries = 4`, `retransmit_timeout = 2` and `retransmit_base = 1.4`.

Also, for non-trivial changes we have some [contribution requirements](https://wiki.strongswan.org/projects/strongswan/wiki/Contributions#Contributing).

ttroy50:
Following the same settings as the IKE behaviour makes sense. 

I'll look to make those changes and also add the copyright header to the changed files.

ttroy50:
I've made the recommended changes and updated the licence in all the changed .c and .h files. The defaults are as you suggested and it's very close to current behaviour. 

I also added the settings to the .opt and .conf files so that the documentation reflects the code.

tobiasbrunner:
Great, thanks a lot! I've squashed the commits to a single one and merged it with some small changes to master.

PR #21, opened 2015-12-08
Allow send_vendor_natt_ids option in charon.conf

heri16:
Some IKEv1 NAT-T implementations on hardware are buggy. Allow an easy way to disable sending of nat capabilities to allow charon to work with these implementations.


tobiasbrunner:
> Some IKEv1 NAT-T implementations on hardware are buggy.

Buggy how?

>  Allow an easy way to disable sending of nat capabilities to allow charon to work with these implementations.

One problem I see with this is that it's a global option. So if this is disabled no connections may use NAT-T. And I'd rename the option to something like _ikev1_use_nat_t_ and the default should obviously be TRUE.

PR #23, opened 2015-12-16
stroke: print DH group proposal for child-sa

gollub:


tobiasbrunner:
Thanks. I guess this was originally not logged because for IKEv2 the first CHILD_SA is created without a separate DH exchange (the key material is derived from the IKE keys). Could be that the output was not correct back then, or to avoid confusing users not familiar with this IKEv2 property. Anyway, now there is just no DH group listed until an IKEv2 CHILD_SA has been rekeyed (or the SA is created with a separate CREATE_CHILD_SA exchange). For IKEv1 we always see the DH group. Since we also report the DH group for CHILD_SAs via vici/swanctl adding it to stroke is at least consistent.

I pushed some patches to the _stroke-list-dh_ branch.

gollub:
Looks good to me. Closing my PR in favor on your proposal.

PR #24, opened 2015-12-18
configure: support systemd >= 209

cjp256:
libsystemd-journal and libsystemd-daemon are now just
part of libsystemd.

Keep original systemd checks as a fallback.

Updates charon-systemd/Makefile.am accordingly.

Tested on:
- debian wheezy (systemd v44)
- ubuntu 15.10 (systemd v255).

Signed-off-by: Chris Patterson <pattersonc@ainfosec.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -941,13 +941,17 @@ if test x$systemd = xtrue; then
 		AC_MSG_ERROR([not found (try --with-systemdsystemunitdir)])
 	fi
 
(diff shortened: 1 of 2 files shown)


cjp256:
I updated this to retain backwards compatibility with older systemd.  Is this the preferred approach, or should it just move forward?  I can change it, if desired :)

Thanks!

tobiasbrunner:
Thanks. Pushed to master with some minor whitespace changes.

PR #26, opened 2016-01-04
[RFC] stroke: print host ports

gollub:
Print host ports to gather NAT source/destination port as in pluto.

Note: this might break existing third-party application relying on parsing the statusall output:

```
peer-10.10.2.3-tunnel-1[1]: ESTABLISHED 21 seconds ago, 10.10.2.2[500][10.10.2.2]...10.10.2.3[500][10.10.2.3]
```

vs. (original):

```
peer-10.10.2.3-tunnel-1[1]: ESTABLISHED 21 seconds ago, 10.10.2.2[10.10.2.2]...10.10.2.3[10.10.2.3]
```

### DIFF
--- a/src/libcharon/plugins/stroke/stroke_list.c
+++ b/src/libcharon/plugins/stroke/stroke_list.c
@@ -114,7 +114,7 @@ static void log_ike_sa(FILE *out, ike_sa_t *ike_sa, bool all)
 		fprintf(out, " %V ago", &now, &established);
 	}
 
-	fprintf(out, ", %H[%Y]...%H[%Y]\n",
+	fprintf(out, ", %#H[%Y]...%#H[%Y]\n",
 			ike_sa->get_my_host(ike_sa), ike_sa->get_my_id(ike_sa),
 			ike_sa->get_other_host(ike_sa), ike_sa->get_other_id(ike_sa));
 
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
I don't really like the look of that as the identities are already printed in brackets. And since charon never did this it's more likely to break current third-party tools than have any benefit in providing backwards compatibility with pluto (which was removed years ago).

And listing ports when enumerating "listening" addresses makes no sense as that is a list of known addresses not a list of endpoints/sockets.

For what reason do you need to know the ports?

gollub:
For a third-party tool which used to parse/print the used NAT src/dst port.

I agree the listing address %#H change might not require.

The only information missing compared to pluto status output is the NAT src/dst port pair IMHO.

tobiasbrunner:
> For a third-party tool which used to parse/print the used NAT src/dst port.

Yeah, sure. But what's the point of that?

gollub:
Nothing special, just trying to provide in the third-party tool the same status information as was available for pluto. Just printing the used src/dst NAT port to indicate that port X and Y are in use for that connection.

Or is there another way to obtain this information via stroke today?

tobiasbrunner:
> Or is there another way to obtain this information via stroke today?

No. But you might actually want to consider using [vici](https://wiki.strongswan.org/projects/strongswan/wiki/Vici)/[swanctl](https://wiki.strongswan.org/projects/strongswan/wiki/Swanctl) instead of the legacy stroke interface. It currently does not provide this information either, but it is definitely easier to add without breaking any current users. I actually just did that in the _vici-ports_ branch.

gollub:
Switching to vici is on my (wish) list. Hopefully with the next (third-party) release.

Ok, let's close that PR for now. I'll propose internally to carry that as an downstream patch for now - to not break other users tools.

Thanks!

PR #28, opened 2016-01-20
android: add CFLAG depending on abi

jberrenberg:
Because HAVE_PTHREAD_COND_TIMEDWAIT_MONOTONIC is not working for pthreads on
arm64-v8a, compilation fails. This commit makes adding the flag conditional.
All non arm64-v8a build still include the flag.

### DIFF
--- a/src/frontends/android/app/src/main/jni/Android.mk
+++ b/src/frontends/android/app/src/main/jni/Android.mk
@@ -39,7 +39,6 @@ strongswan_CFLAGS := \
 	-DHAVE_ALLOCA \
 	-DHAVE_CLOCK_GETTIME \
 	-DHAVE_DLADDR \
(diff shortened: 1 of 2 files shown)


tobiasbrunner:
This is not a problem with the arm64-v8a ABI per se, but with API level 21. For all 64-bit ABIs the NDK automatically uses level 21 and this function is deprecated there (as it's a proprietary extension, however, older levels don't support `pthread_condattr_setclock`). We currently don't support level 21 as building against it will break the app on older Android versions (which is not a problem for 64-bit ABIs that are not supported by these anyway). The proper fix is to replace `pthread_cond_timedwait_monotonic` with `pthread_condattr_setclock`, which was added with level 21, when building against this API level. I did that in the _android-64-bit_ branch.

jberrenberg:
In that case, i am sorry for bothering you. 
Thanks for the detailed explanation and the great work.

PR #29, opened 2016-01-20
android: make loose identity configurable

jberrenberg:
To allow the android client wrapper to set the IDr when necessary this commit
adds a new settings key. This key may be used in the CharonVpnService.java and
is used in android_service.c.

* new key "connection.loose_identity"
* default value still TRUE (as before)

related to Issue #1268


tobiasbrunner:
See https://wiki.strongswan.org/issues/1268

PR #30, opened 2016-02-03
feat: updown hander to include usage stats

willsteel:
Adds new environment variables to the updown script hook:

 - ${PLUTO_BYTES_IN}
 - ${PLUTO_BYTES_OUT}
 - ${PLUTO_PACKETS_IN}
 - ${PLUTO_PACKETS_OUT}
 - ${PLUTO_DURATION}

When a connection is terminated, these new values give indication about how it has been used. The ${PLUTO_DURATION} variable is only set if there has been any traffic on the connection.

### DIFF
--- a/src/libcharon/plugins/updown/updown_listener.c
+++ b/src/libcharon/plugins/updown/updown_listener.c
@@ -17,6 +17,7 @@
 #define _GNU_SOURCE
 #include <stdio.h>
 #include <unistd.h>
+#include <time.h>
 
 #include "updown_listener.h"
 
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Unfortunately, this won't work. The usage data you'd get this way is only for the last pair of SAs because with each rekeying new IPsec SAs are created and the counters, thus, are reset (the updown script is not called when SAs are rekeyed). Please have a look at the [_eap-radius_ plugin](https://github.com/strongswan/strongswan/blob/master/src/libcharon/plugins/eap_radius/eap_radius_accounting.c) to see an example of what's necessary to get at least somewhat accurate numbers for a complete session.

willsteel:
Thanks for the Update. I'm currently developoing a plugin called 'quota' that is made of code in parts from 'updown' and 'eap-radius'. Doing only the tracking and not the radius massaging from the eap-radius module and then making shell hooks just like updown but for accounting purposes. 

This also solves the issue that the original updown script (routing stuff) is not called.

I will make a new pull request when ready...

PR #32, opened 2016-02-15
fix: gitignore

willsteel:
Adds stuff that is created by automake/configure:
    strongswan.config
    strongswan.files
    strongswan.includes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ coverage/
 .metadata/
 .settings/
 test-driver
+strongswan.config
+strongswan.files
+strongswan.includes


tobiasbrunner:
I've never seen these files. What version of automake/configure creates them (and at what point)?

willsteel:
Sorry, turns out they were created by qt-creator IDE. I thought it was from the gcc-toolchain. Nevertheless its a popular IDE and you might want to consider ignore its created files.

If not, doesn't matter.

tobiasbrunner:
I see.

PR #33, opened 2016-02-18
feat: call scripts in IPSEC_CONFDIR as pwd

m-schmoock:
Enables relative script links in strongswan.conf and ipsec.conf.
Currently only absolute paths are working, which is unhandy for deployment.

This change does a 'chdir' to IPSEC_CONFDIR after forking and before
executing a shell.

### DIFF
--- a/src/libstrongswan/utils/process.c
+++ b/src/libstrongswan/utils/process.c
@@ -158,6 +158,7 @@ process_t* process_start(char *const argv[], char *const envp[],
 		return NULL;
 	}
 
+	char* ipsec_confdir = getenv("IPSEC_CONFDIR");
 	this->pid = fork();
 	switch (this->pid)
 	{
@@ -194,6 +195,10 @@ process_t* process_start(char *const argv[], char *const envp[],
 			if (close_all)
 			{
 				closefrom(3);
+			}
+			if (ipsec_confdir)
+			{
+				chdir(ipsec_confdir);
 			}
 			if (execve(argv[0], argv, envp ?: empty) == -1)
 			{
--- a/src/libstrongswan/utils/process.c
+++ b/src/libstrongswan/utils/process.c
@@ -196,9 +196,10 @@ process_t* process_start(char *const argv[], char *const envp[],
 			{
 				closefrom(3);
 			}
-			if (ipsec_confdir)
+			if (ipsec_confdir && chdir(ipsec_confdir) == -1)
 			{
-				chdir(ipsec_confdir);
+				DBG1(DBG_LIB, "chdir to IPSEC_CONFDIR=%s failed: %s",
+						ipsec_confdir, strerror(errno));
 			}
 			if (execve(argv[0], argv, envp ?: empty) == -1)
 			{


willsteel:
Obviously, the test scripts need to have a certain PWD...

willsteel:
Is `WIN32` consistency required for such a change?

tobiasbrunner:
While I don't really see a problem with configuring absolute paths, the ability to configure relative paths might simplify configuration in some situations. Unfortunately, there are several issues with this patch:

1. `process_start()` is a generic library function, just always changing to `IPSEC_CONFDIR` (which usually translates to `/etc`) is probably not expected by callers of that function. So the working directory would definitely have to be passed as argument, if that approach would be used. And yes, Windows support would be required.
2. You read `IPSEC_CONFDIR` from an environment variable, however, only the `ipsec` script defines this. If the charon-systemd daemon is started via systemd it wouldn't be defined (unless it is explicitly added to the service file). So that might result in inconsistent behavior.
3. Using `IPSEC_CONFDIR` as working directory might not always be expected. For ipsec.conf-based configurations, where ipsec.conf is usually in that directory, this might work. However, swanctl.conf is located in `IPSEC_CONFDIR/swanctl` so for updown scripts defined there relative paths might better be resolved relative to that directory. And if the configuration is passed via VICI directly (e.g. via Python bindings) there wouldn't even be a specific root directory (the same is true for updown scripts defined in an SQL database).
4. The locations of the config files are not necessarily static, for instance, the path to ipsec.conf can be set via the `--conf` argument for `ipsec start` or via the _starter.config_file_ configuration option. Similarly, the path to strongswan.conf can be changed with the `STRONGSWAN_CONF` environment variable. So if paths defined in these files should be relative to the config files they would have to be adapted dynamically.
5. The _updown_ plugin, which executes the updown scripts, does not know where a script's path originally came from (i.e. from what config file). So it wouldn't actually be able to pass the correct working directory to `process_create_shell()`. Paths would definitely have to be resolved before.

So unless we specify that paths for all scripts are always resolved relative to `IPSEC_CONFDIR` (i.e. `sysconfdir`, which is usually `/etc`, but is `/usr/local/etc` by default and e.g. `/etc/strongswan` on Fedora) - which I'm not sure is the best approach - the only way to do this properly is to let the individual components that read the configuration files resolve the relative paths. So _starter_, which parses ipsec.conf, would do so relative to the actual location of that file, and _swanctl_, which parses swanctl.conf, would do so relative to that file's location. For scripts in strongswan.conf the components that execute those (e.g. the _ext-auth_ plugin) would have to do that individually, probably based on `lib->conf`, if it is defined.
Making the paths relative to the config files could be a bit confusing, though, for one because the location of these files could change, but also if config files in other directories are included. That's because the code that would resolve the relative paths only sees the complete configuration file. It wouldn't know the location of an included file that configured a specific script but only that of the main file. So a simpler, more consistent, option might be to use the approach we use for credentials, that is, we define that relative paths configured in ipsec.conf are relative to `IPSEC_CONFDIR/ipsec.d/updown` (similar to file names of certificates set in _left|rightcert_, which are resolved relative to `IPSEC_CONFDIR/ipsec.d/certs`) and those in `swanctl.conf` are relative to `IPSEC_CONFDIR/swanctl/updown`. For scripts in strongswan.conf we could e.g. use `IPSEC_CONFDIR/strongswan.d/scripts` (if that's even necessary - since all components would have to handle script paths themselves they could also use separate directories or insist on absolute paths).

What do you think?

PR #34, opened 2016-02-25
Include required headers for FreeBSD

denisvm:
duplicheck.c code fails to compile in FreeBSD due to lack of a include file. Also there are two warnings for undeclared functions.

```
duplicheck.c: In function 'make_connection':
duplicheck.c:34: error: field 'in' has incomplete type
duplicheck.c:42: error: 'INADDR_LOOPBACK' undeclared (first use in this function)
duplicheck.c:42: error: (Each undeclared identifier is reported only once
duplicheck.c:42: error: for each function it appears in.)
duplicheck.c:49: warning: incompatible implicit declaration of built-in function 'strcpy'
duplicheck.c:51: warning: incompatible implicit declaration of built-in function 'strlen'
```

### DIFF
--- a/src/libcharon/plugins/duplicheck/duplicheck.c
+++ b/src/libcharon/plugins/duplicheck/duplicheck.c
@@ -19,8 +19,10 @@
 #include <stdlib.h>
 #include <stddef.h>
 #include <stdio.h>
+#include <string.h>
 #include <errno.h>
 #include <arpa/inet.h>
+#include <netinet/in.h>
 
 #include "duplicheck_msg.h"
 


tobiasbrunner:
Why do you want to build the _duplicheck_ plugin on FreeBSD? Did you read [the wiki page](https://wiki.strongswan.org/projects/strongswan/wiki/Duplicheck)?

denisvm:
Hello Tobias, I've had few problemas with duplicated security associations being negotiated without dropping the older ones, which caused the tunnel traffic be dropped after reach the peer's margin time. This plugin seems to have solved the problem.

tobiasbrunner:
Sounds like you should rather fix the underlying problem instead of using this plugin (it deletes all SAs when a duplicate is detected, not sure if that's what you really want).

denisvm:
@tobiasbrunner I thought it would delete only the old SA, not the current one. At least after I enable the plugin, it solved the problem.
I've running with the same problem in some cases, and without this plugin, the security associations keep duplicating if the other peer doesn't delete them on restart for instance. `uniqueids` option seems to have no effect here.

tobiasbrunner:
> I thought it would delete only the old SA, not the current one.

It does if the old SA is not active anymore, but if it is then both are deleted.

> the security associations keep duplicating if the other peer doesn't delete them on restart for instance.

And why is that? Seems you should try to find out what exactly happens here.

denisvm:
>> I thought it would delete only the old SA, not the current one.

> It does if the old SA is not active anymore, but if it is then both are deleted.

So if I have an old SA that is not active anymore, it is going to be deleted, right? That's exactly the behaviour that I'm expecting.

>> the security associations keep duplicating if the other peer doesn't delete them on restart for instance.

> And why is that? Seems you should try to find out what exactly happens here.

I don't have control over the other peer, so I have to make sure that my peer will do all the magic needed to remove old/inactive security associations to avoid conflicts.

Would you have any other suggestion for this scenario? Thanks

tobiasbrunner:
>>> I thought it would delete only the old SA, not the current one.

>> It does if the old SA is not active anymore, but if it is then both are deleted.

> So if I have an old SA that is not active anymore, it is going to be deleted, right? That's exactly the behaviour that I'm expecting.

But why would there be an old inactive SA in the first place? And why not get rid of inactive SAs with DPD? I'm also not sure how the _duplicheck_ plugin interacts with make-before-break reauthentication (i.e. rekeying in IKEv1).

>>> the security associations keep duplicating if the other peer doesn't delete them on restart for instance.

>> And why is that? Seems you should try to find out what exactly happens here.

> I don't have control over the other peer, so I have to make sure that my peer will do all the magic needed to remove old/inactive security associations to avoid conflicts.

Again, why would there be "old" SAs? And what conflicts are they causing exactly?

denisvm:
@tobiasbrunner looks like DPD isn't being triggered at all for this SA which was causing some trouble. That connection is an IKEv2

Also, I have other connections with older IKEv1 peers which have DPD disabled or that doesn't support it, which disables DPD on strongSwan entirely for those peers, where I get this output:
```
charon: 16[IKE] DPD not supported by peer, disabled
charon: 16[CFG] detected duplicate IKE_SA for 'A.B.C.D', triggering delete for old IKE_SA
```

On the IKEv2 troubled connection, my peer sometimes starts new associations without replacing or timing out the older SAs, even with DPD enabled. I have those flags already also configured: `make_before_break = yes` and `uniqueids = yes`.

denisvm:
@tobiasbrunner If you prefer, I can open a ticket on Redmine regarding those issues. Anyways, I think this patch could be useful for others who may want to use/try this plugin on FreeBSD.

tobiasbrunner:
Sure, I've no problem with the patch as such. Do you maybe want to use a different email address for your commit (i.e. an actual address instead of _denis@Denis-MacBook-Pro.local_)?

Those other issues seem to be problems of these other peers/implementations. If you want you may open a ticket in our issue tracker but it might be difficult to debug this, in particular, with proprietary implementations, which often also have bad logs.

denisvm:
@tobiasbrunner Sorry about the address, I've just fixed that. Thanks!
I'll collect more data about this issue then I'll open a ticket about that. Thanks for your attention :)

PR #36, opened 2016-03-09
added some missing returns to the python vici wrapper

Sebubu:
According to [the vici readme](https://github.com/strongswan/strongswan/tree/master/src/libcharon/plugins/vici) quiet every vici method returns a value. The python wrapper has missed some of these returns. This pull request fixes this.

Gruss Severin

### DIFF
--- a/src/libcharon/plugins/vici/python/vici/session.py
+++ b/src/libcharon/plugins/vici/python/vici/session.py
@@ -30,8 +30,11 @@ class Session(object):
 
     def reload_settings(self):
         """Reload strongswan.conf settings and any plugins supporting reload.
+
+        :return: generator for logs emitted as dict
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
That's on purpose. For all the VICI commands the Python bindings don't return a value no log messages are expected and the VICI response message looks like this:
```
{
    success = <yes or no>
    errmsg = <error string on failure>
}
```
Instead of returning this data and letting the caller analyze it the bindings check the response and raise a `CommandException` that includes `errmsg` if `success` is not _yes_, see [session.py](https://github.com/strongswan/strongswan/blob/819da83fccf99acf7af1ed2bf61a498425c375e1/src/libcharon/plugins/vici/python/vici/session.py#L287).

PR #37, opened 2016-03-12
Get value for PLUTO_MARK_{IN,OUT} from child_sa

sorah:
Or the invoked script will get broken value when `mark=%unique` is used in a configuration.


tobiasbrunner:
Makes sense. Thanks!

PR #39, opened 2016-03-29
Some cleanups

lkundrak:
Split off from pull reques #15 for ease of review.


tobiasbrunner:
Looks good, up to the last commit. The problem there is that there is still a call to `gnome_keyring_set_network_password_sync()` which is now obviously not defined anymore.  I tried migrating that to `secret_password_store_sync()`, which seems to compile, not sure if it would actually work.  I pushed your changes (with some minor tweaks to the commit messages) and that (squashable) fix to the _nm-updates_ branch (it also includes a fix regarding the LIBS).

PR #40, opened 2016-03-30
Fix dictionary keys

cameronmccord2:
The dictionary keys for the requests need to have a (-) instead of a (_). They should follow the same style as swanctl.


tobiasbrunner:
Applied to master, thanks!

PR #41, opened 2016-04-14
appdata: add AppStream metadata

lkundrak:
This will ensure the strongSwan NetworkManager plugin will be easily
installable from the app stores such as GNOME Software.

### DIFF
--- a/src/frontends/gnome/.gitignore
+++ b/src/frontends/gnome/.gitignore
@@ -9,3 +9,4 @@ nm-strongswan-service.name
 stamp-h1
 config.guess.cdbs-orig
 config.sub.cdbs-orig
+NetworkManager-strongswan.appdata.xml
--- a/src/frontends/gnome/Makefile.am
+++ b/src/frontends/gnome/Makefile.am
@@ -10,6 +10,11 @@ nmvpnservice_DATA = nm-strongswan-service.name
 
 @INTLTOOL_DESKTOP_RULE@
 
+appdatadir = $(datadir)/appdata
+appdata_DATA = $(appdata_in_files:.xml.in=.xml)
+appdata_in_files = NetworkManager-strongswan.appdata.xml.in
+@INTLTOOL_XML_RULE@
+
 nm-strongswan-service.name: $(srcdir)/nm-strongswan-service.name.in
 	$(AM_V_GEN) \
 	sed -e 's|[@]LIBEXECDIR[@]|$(libexecdir)|' \
@@ -17,11 +22,13 @@ nm-strongswan-service.name: $(srcdir)/nm-strongswan-service.name.in
 
 EXTRA_DIST = nm-strongswan-service.name.in \
     $(dbusservice_DATA)  \
+    $(appdata_in_files)  \
+    $(appdata_DATA)      \
     intltool-extract.in  \
     intltool-merge.in    \
     intltool-update.in
 
-CLEANFILES = $(nmvpnservice_DATA) *~
+CLEANFILES = $(nmvpnservice_DATA) $(appdata_DATA) *~
 DISTCLEANFILES = intltool-extract intltool-merge intltool-update
 
 ACLOCAL_AMFLAGS = -I m4
--- a/src/frontends/gnome/NetworkManager-strongswan.appdata.xml.in
+++ b/src/frontends/gnome/NetworkManager-strongswan.appdata.xml.in
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright Lubomir Rintel 2016 -->
+<component type="addon">
+  <id>network-manager-strongswan</id>
+  <project_license>GPL-2.0+</project_license>
+  <metadata_license>CC0-1.0</metadata_license>
+  <extends>nm-connection-editor.desktop</extends>
+  <extends>gnome-control-center.desktop</extends>
+  <_name>strongSwan VPN client</_name>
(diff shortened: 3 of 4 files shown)


tobiasbrunner:
Isn't this something distributions should provide? In particular regarding the bugtracker and contact email addresses. Or do they generally patch this (or doesn't that matter anyway)?

The screenshots should probably be updated (or at least served from more stable URLs like e.g. strongswan.org/images). And the second screenshot shows the libgnomeui password dialog, so that will look quite differently I guess with NM 1.2 (maybe we should just drop it?).

Another thing is that the plugin does not support IKEv1. So those two references should be removed.

lkundrak:
> Isn't this something distributions should provide? In particular regarding the bugtracker and contact email addresses. Or do they generally patch this (or doesn't that matter anyway)?

They do patch it. AppStream is cross-distribution: The point is that the upstreams fill in as much as possible (and relevant) and distribution fills in the rest (the actual distribution package name, icons, rating, etc.) while generating a feed from the packages they provide.

> The screenshots should probably be updated (or at least served from more stable URLs like e.g. strongswan.org/images). And the second screenshot shows the libgnomeui password dialog, so that will look quite differently I guess with NM 1.2 (maybe we should just drop it?).

Yeah, better screenshots and on a better URL would be perfect. The appstream tooling is picky about the screenshots and complains loudly now:

[lkundrak@belphegor gnome]$ appstream-util validate NetworkManager-strongswan.appdata.xml.in 
• attribute-invalid     : <screenshot> width too small [https://wiki.strongswan.org/attachments/download/27/nm-strongswan-config.png]
• style-invalid         : <image> has vertical padding [https://wiki.strongswan.org/attachments/download/27/nm-strongswan-config.png]
• style-invalid         : <image> has horizontal padding [https://wiki.strongswan.org/attachments/download/27/nm-strongswan-config.png]
• attribute-invalid     : <screenshot> width too small [https://wiki.strongswan.org/attachments/download/26/nm-strongswan-auth.png]
• attribute-invalid     : <screenshot> height too small [https://wiki.strongswan.org/attachments/download/26/nm-strongswan-auth.png]
• style-invalid         : <image> has vertical padding [https://wiki.strongswan.org/attachments/download/26/nm-strongswan-auth.png]
• style-invalid         : <image> has horizontal padding [https://wiki.strongswan.org/attachments/download/26/nm-strongswan-auth.png]

If you need a compliant screenshot to host, you can use https://people.freedesktop.org/~lkundrak/strongswan.png

It's probably a bit wider than usual, but makes the tools happy and it looks better in the actual app store.

Note that the screenshot URLs is something that the distributions tooling patches in the XML: they fetch the screenshots and cache them in their own infrastructure and add a couple of links to mirrors.

> Another thing is that the plugin does not support IKEv1. So those two references should be removed.

Fixed.

tobiasbrunner:
Thanks for the clarifications.

> If you need a compliant screenshot to host, you can use https://people.freedesktop.org/~lkundrak/strongswan.png
> 
> It's probably a bit wider than usual, but makes the tools happy and it looks better in the actual app store.

Looks great, however, I'd prefer an (additional) example that uses EAP or certificate authentication (we don't promote PSK authentication generally - even though that PSK looks comparatively strong :-). Could you provide one?

lkundrak:
> Looks great, however, I'd prefer an (additional) example that uses EAP or certificate authentication (we don't promote PSK authentication generally

Updated. If you host the picture somewhere else. please amend the patch when applying it.

It doesn't matter too much where the picture is (it's going to be mirrored, so it won't consume bandwidth or be latency-sensitive); it's just important that it doesn't go away too quickly. A web site is okay; people also tend to use raw.github.com or their cgit instances.

> even though that PSK looks comparatively strong :-).

Well, that password was copied from an example in the documentation :)

tobiasbrunner:
> Updated. If you host the picture somewhere else. please amend the patch when applying it.

Thanks a bunch. I did so in the _nm-appstream_ branch.

I saw that according to the [docs](https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Addon.html) the files for Addons should be located at `/usr/share/metainfo/%{id}.metainfo.xml` (`/usr/share/appdata/` is listed as [legacy path](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html)). Should we change that? I also wondered whether `appdata_DATA` should actually be added to the distribution, or if `appdata_in_files` is enough (at least the [docs](https://www.freedesktop.org/software/appstream/docs/sect-Quickstart-Translation.html) indicate as much).

lkundrak:
> I saw that according to the docs the files for Addons should be located at /usr/share/metainfo/%{id}.metainfo.xml (/usr/share/appdata/ is listed as legacy path). Should we change that?

The /usr/share/metainfo is too new to be used by actual distribution or appdata-glib. I've asked the appstream upstream and they recommend to start using it no sooner than in a year. Thus we need to stick with /usr/share/appdata at least for now.

> I also wondered whether appdata_DATA should actually be added to the distribution, or if appdata_in_files is enough (at least the docs indicate as much).

I don't know. I've cargo-culted it from somewhere. GNOME projects don't seem to ship the generated file. On the other hand, generating the file requires intltool-update which is not guaranteed to be around (the user could do a ./configure --disable-nls). For that reason I would prefer to keep it.

tobiasbrunner:
> I've asked the appstream upstream and they recommend to start using it no sooner than in a year. Thus we need to stick with /usr/share/appdata at least for now.

Ah, OK. Thanks for the clarification.

> For that reason I would prefer to keep it.

OK, no problem.

PR #43, opened 2016-04-25
Support dmdName RDN

y-cann:
Add dmdName attribute for RDN (oid: 2.5.4.54).


tobiasbrunner:
Is anybody actually using this? It was removed with [RFC 4519](https://tools.ietf.org/html/rfc4519#appendix-A).

y-cann:
It is still used by some governmental organisation. At least, one of our client.

tobiasbrunner:
OK, applied to master. Thanks.

PR #44, opened 2016-05-03
When Android connection cannot identify the wildcard domain name

coolcol001:
When Android connection cannot identify the wildcard domain name

my server config with a wildcard domain name like *.myvpn.com

when i connnect to one of vpn server like  us1.myvpn.com,the android client report:

SSL cert's Common Name is : *.myvpn.com
constraint check failed: indentity 'us1.myvpn.com' required.


tobiasbrunner:
Wildcard certificates are not supported by strongSwan (see e.g. [#629](https://wiki.strongswan.org/issues/629 on our issue tracker).  However, in the current version of the Android app it is possible to set the server's identity explicitly, which allows you to use a different identity than the server's IP/FQDN (so if you have multiple servers with different hostnames they all could use the same identity).

coolcol001:
but we connect  by win7 or win8,the Wildcard certificates works fine with Strongswan on centos version 5.3.4.
if there have some way to support on android ?

PR #47, opened 2016-05-24
New plugin named dead-peer-notify-plugin

Void-p:
After the last retransmit has been sent and the peer is considered dead ( plugin hooks ALERT_RETRANSMIT_SEND_TIMEOUT alert ), plugin sends an email with notification, which contains a name and an ip address of the dead peer, or( and ) executes external command with two arguments: dead peer name and IP address. System administrators may find it useful for configuring email alerts and customizing other actions.​

### DIFF
--- a/conf/Makefile.am
+++ b/conf/Makefile.am
@@ -32,6 +32,7 @@ plugins = \
 	plugins/bliss.opt \
 	plugins/certexpire.opt \
 	plugins/coupling.opt \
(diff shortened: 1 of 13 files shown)


jia3ep:
@tobiasbrunner this plugin is widely used in products of Infotecs company. Is there a chance it will be accepted?

tobiasbrunner:
No, I have no intention to include this in our repository.

PR #48, opened 2016-06-21
IKEV1: Force delete IKE_SA during quick mode. Issue 479

prathyks:


prathyks:
Changes in this patch are to ensure IKEv1 IKE-SA delete happens irrespective of the tasks in the lists 'queued' and 'active'. Details about the fix are explained in the issue:

https://wiki.strongswan.org/issues/1537

PR #49, opened 2016-08-04
An option to save IKE_SA  keys for wireshark

superCodrut:
This is the first patch series for feature  #1557.


neeraj9:
This is definitely a very useful feature for debugging.
https://wiki.strongswan.org/issues/1557

superCodrut:
Indeed, this feature is for debugging and the cool part about it is that this feature will not kill performance.

superCodrut:
The second patch series contains the code for saving CHILD_SA keys.

superCodrut:
@tobiasbrunner , @strongX509 and @martinwilli can you review this code?

tobiasbrunner:
> This is definitely a very useful feature for debugging.

@neeraj9 Debugging what exactly? @superCodrut You also never responded to my similar question in [#1557](https://wiki.strongswan.org/issues/1557).

I had a quick look and there are several issues:

 * Several fixups could be squashed into other commits.
 * In some commits are changes that are clearly not related to the rest, so they should be split up (and parts possibly squashed in other commits).
 * There is test code in several commits that should be removed.
 * The indention in lots of commits is incorrect as is the [code style](https://wiki.strongswan.org/projects/strongswan/wiki/ProgrammingStyle).
 * Please don't use camel case, the plugin should probably be called `save-keys/save_keys` (look at other plugins with two-part names).
 * I don't really like the additional `save_*_keys` hooks (in general and in particular because of their arguments). I'd rather see you using the existing `*_keys` hooks and doing the key derivation in the plugin (as there is no logging required and the error handling could be simplified this shouldn't be that much code).
 * The `send_spis` hook is not necessary as the SPIs can be retrieved from the IKE_SA directly (it is passed to the `ike_keys` hook).
 * Not your fault but the format used by Wireshark is just the worst (algorithm identifiers can't even be passed numerically). To avoid compatibility problems and to simplify the code it might be better to just define the complete strings in a map (identifier -> string).
 * The contents of chunk_t instances can be printed as plain hex strings with the `%+B` printf specifier (no need to call `chunk_to_hex()`).
 * The path for the files to which keys etc. are written should be configurable already when they are first introduced (basically an extension of the test code comment above). If the setting is not configured the plugin should be disabled (or if writing to `/tmp` is preferred in that case the files should at least be created with e.g. `mkstemp()` or written to a directory created with `mkdtemp()`). Since this format is very Wireshark specific it might be useful to have different schemes to store the keys later. So the current option should perhaps be called `wireshark_directory` or similar (and only if it is set will files in this format and with specific names be created/extended).

There must be some safeguards that prevent this plugin from getting loaded automatically when it is built. For instance, it should not be included in the default plugin list used by the daemon when it is enabled (`charon` in `ADD_PLUGIN` in `configure.ac`) so it would have to be loaded explicitly (also needs some work regarding the config snippets in `conf/plugins`, which you currently ignored). Doing nothing until a path is configured would help too.

> the cool part about it is that this feature will not kill performance

I don't really see how that's relevant for a debugging feature that should never be enabled outside of specific test scenarios.

strainu:
@tobiasbrunner, debugging interop with a different device that does not have capture and/or proper logging capabilities, for instance.

tobiasbrunner:
@strainu Again debugging what exactly? IKE? What aspects of it? strongSwan can also log a lot of details about the IKE packet encoding if the debug level is increased.

strainu:
Both IKE and encrypted traffic. For IKE, when working with a very high number of tunnels, extended logging is not really feasible. Logging only the keys significantly reduces the strain on the performance. For ESP, NULL encryption is not acceptable for all use-cases.

tobiasbrunner:
That does not really sound like a debugging test scenario.

vamposdecampos:
Thanks @tobiasbrunner for reviewing!

On the use-cases:

1. There's the old adage of "I don't trust your implementation".  Verification with a 3rd-party implementation (in this case, wireshark) can be enlightening. This also applies to in-kernel ESP, which strongswan can't log.  Captures + keys can be more convenient than logs (e.g. correlations with other network traffic), particularly if you already have a packet capture tool running somewhere in the network, and you realize post-factum that you'd like to investigate something.

1. We're thinking of enabling this on the automated test suite, and exporting captures + keys for all tests.  This would produce a ton of [sample captures](https://wiki.wireshark.org/SampleCaptures) for all kinds of scenarios (both IKE and ESP), which would be a great learning resource.

On the patches:

- the one point to debate would be how exactly to extract the keys.  IMHO for use-case 1) above, it would be beneficial to log the exact keys used later on by the strongswan proper, rather than deriving them again -- but I'll admit that it does make for a more hairy implementation.  In the end, of course, The Maintainer Is Always Right ;)

tobiasbrunner:
> There's the old adage of "I don't trust your implementation". Verification with a 3rd-party implementation (in this case, wireshark) can be enlightening. This also applies to in-kernel ESP, which strongswan can't log. 

Doesn't this apply way more to proprietary implementations/devices than open-source software like strongSwan or the Linux kernel? And while strongSwan obviously can't log information about the ESP packets sent by the kernel it already can log the keys of these SAs.

> Captures + keys can be more convenient than logs (e.g. correlations with other network traffic), particularly if you already have a packet capture tool running somewhere in the network, and you realize post-factum that you'd like to investigate something.

That's one of the reasons I'm reluctant to include a plugin like this in our distribution. Storing the keys just to have them on hand when needed later (for whatever unspecified reason) defeats the whole purpose of dynamic keying and rekeying (PFS). Of course, it can be done already by writing a custom plugin/script but shipping a ready-made plugin might convey the unintended message that we somehow endorse this kind of behavior.

> We're thinking of enabling this on the automated test suite, and exporting captures + keys for all tests.

Are you referring to our test suite? That would have to be hidden somehow as we already struggle with users literally copying the configs of our [test scenarios](https://www.strongswan.org/testing/testresults/) (despite [our warnings](https://wiki.strongswan.org/projects/strongswan/wiki/ConfigurationExamplesNotes)) and we definitely wouldn't want anybody using this plugin unintentionally.

superCodrut:
>@neeraj9 Debugging what exactly? @superCodrut You also never responded to my similar question in [#1557](https://wiki.strongswan.org/issues/1557).

@strainu  and @vamposdecampos responded this one for me.
> - Several fixups could be squashed into other commits.
> - In some commits are changes that are clearly not related to the rest, so they should be split up (and parts possibly squashed in other commits).

Fixed it by rebasing the code.
> - There is test code in several commits that should be removed.

Done.
> - The indention in lots of commits is incorrect as is the [code style](https://wiki.strongswan.org/projects/strongswan/wiki/ProgrammingStyle).

Done.
> - Please don't use camel case, the plugin should probably be called save-keys/save_keys (look at other plugins with two-part names).

Done.
> - I don't really like the additional save_*_keys hooks (in general and in particular because of their arguments). I'd rather see you using the existing *_keys hooks and doing the key derivation in the plugin (as there is no logging required and the error handling could be simplified this shouldn't be that much code).

In a lab enviroment, if we have tens of thousands of connections and we do the key derivation in the plugin then this scenario will take much longer than the scenario when we use the save_*_keys hooks.
> - The send_spis hook is not necessary as the SPIs can be retrieved from the IKE_SA directly (it is passed to the ike_keys hook).

If we don't use the ike_keys hook and we keep the save_*_keys hooks, is there another way to receive the SPIs?
> - Not your fault but the format used by Wireshark is just the worst (algorithm identifiers can't even be passed numerically). To avoid compatibility problems and to simplify the code it might be better to just define the complete strings in a map (identifier -> string).

Done. We used the map model used in the [kernel_netlink](https://github.com/strongswan/strongswan/blob/408282196453b3f610c758d6f96236ac1e9da3c7/src/libcharon/plugins/kernel_netlink/kernel_netlink_ipsec.c#L176)
> - The contents of chunk_t instances can be printed as plain hex strings with the %+B printf specifier (no need to call chunk_to_hex()).

Done.
> - The path for the files to which keys etc. are written should be configurable already when they are first introduced (basically an extension of the test code comment above). If the setting is not configured the plugin should be disabled (or if writing to /tmp is preferred in that case the files should at least be created with e.g. mkstemp() or written to a directory created with mkdtemp()). Since this format is very Wireshark specific it might be useful to have different schemes to store the keys later. So the current option should perhaps be called wireshark_directory or similar (and only if it is set will files in this format and with specific names be created/extended).
There must be some safeguards that prevent this plugin from getting loaded automatically when it is built. 

Done. Now even if you compile the plugin, the plugin will be loaded by charon only when the path is set.
> (also needs some work regarding the config snippets in conf/plugins, which you currently ignored). 

Done.

> >We're thinking of enabling this on the automated test suite, and exporting captures + keys for all tests.

>Are you referring to our test suite? That would have to be hidden somehow as we already struggle with users literally copying the configs of our test scenarios (despite our warnings) and we definitely wouldn't want anybody using this plugin unintentionally.

Fair enough, we will drop this idea.

superCodrut:
@tobiasbrunner  is there a chance for this plugin to be accepted?

tobiasbrunner:
I've refactored this and pushed a bunch of commits to the _save-keys_ branch. The new hooks providing the derived keys will most likely end up in our next release. Regarding the plugin, we've not  decided yet whether it stays like that or even gets merged. Perhaps we'll maintain it in a separate branch or provide it as an out-of-tree plugin.

PR #50, opened 2016-09-30
Compatibility with Windows 2012R2: accept certificate requests with ID BMPSTRING 

ramon-garcia:
Hello,

While testing with Windows 2012R2 I found that strongswan could not estabish an IPSEC connection in transport mode because it makes a certificate request with the CA in ASN1_BMPSTRING format. The ID comparison function in libstrongswan/utils/identification.c do not accept matching an DN in ASN1_BMPSTRING with another in ASN1_UTF8STRING.

So I implemented a string function to match a BMPSTRING with a UTF8STRING in strongswan/src/libstrongswan/utils/utils/string.c . I regard BMPSTRING as UTF-16, including handling of characters out of the basic multilingual plane represented. This should be the most general and safe support.

The code tries to be as safe as possible. It does not finish early in case of mismatch, in order to avoid helping an attacker to guess the value of stored IDs of certificates by timing attacks. However, it cannot be timing independent of secret data, because the information about which ID is secret, and which one is provided by the peer is not known.

Best regards.


ramon-garcia:
Need to create again with proper tags

PR #51, opened 2016-09-30
Compatibility with Windows 2012R2: accept certificate requests with ID BMPSTRING

ramon-garcia:
Hello,

While testing with Windows 2012R2 I found that strongswan could not estabish an IPSEC connection in transport mode because it makes a certificate request with the CA in ASN1_BMPSTRING format. The ID comparison function in libstrongswan/utils/identification.c do not accept matching an DN in ASN1_BMPSTRING with another in ASN1_UTF8STRING.

So I implemented a string function to match a BMPSTRING with a UTF8STRING in strongswan/src/libstrongswan/utils/utils/string.c . I regard BMPSTRING as UTF-16, including handling of characters out of the basic multilingual plane represented. This should be the most general and safe support.

The code tries to be as safe as possible. It does not finish early in case of mismatch, in order to avoid helping an attacker to guess the value of stored IDs of certificates by timing attacks. However, it cannot be timing independent of secret data, because the information about which ID is secret, and which one is provided by the peer is not known.

Best regards.


ramon-garcia:
I see that there is a compile error because OS X does not have uchar.h that defines char16_t
Let us see what can be done.

tobiasbrunner:
> While testing with Windows 2012R2 I found that strongswan could not estabish an IPSEC connection in transport mode because it makes a certificate request with the CA in ASN1_BMPSTRING format.

Are you saying the CA certificate's subject is not actually encoded with ASN1_BMPSTRING and that the Windows client uses not the actual subject DN but something similar that uses ASN1_BMPSTRING RDNs? Because if the CA certificate actually used ASN1_BMPSTRING and you have that installed on both ends why wouldn't there be a match? Could you provide more information on this (example certificates, logs etc.).

> The code tries to be as safe as possible. It does not finish early in case of mismatch, in order to avoid helping an attacker to guess the value of stored IDs of certificates by timing attacks.

I don't think that's necessary. The server already sends certificate requests for the CAs it has installed or configured in the fourth Main Mode message (unless that's disabled with `rightsendcert=never`, but do you really consider the subject of CA certificates private?). 

I haven't looked at the code that closely yet, but you should definitely squash the fixup commits. And also have a look at our [contribution requirements](https://wiki.strongswan.org/projects/strongswan/wiki/Contributions).

ramon-garcia:
Yes, during my test I created a test self signed as CA certificate with openssl, then created two machine certificates and assigned one to the Windows 2012R2 server and other to strongswan. And the connection did not work. After troubleshooting this problem I implemented this patch, and now the IPSec connection is established.

I am going to fix the continuus integration issues.

tobiasbrunner:
> Yes, during my test I created a test self signed as CA certificate with openssl, then created two machine certificates and assigned one to the Windows 2012R2 server and other to strongswan. And the connection did not work. After troubleshooting this problem I implemented this patch, and now the IPSec connection is established.

Could you provide more information on this. Like example certificates and logs (_enc_ and _asn_ log levels increased to 3). Changing the encoding for the CA certificate's subject when sending the certificate request doesn't make much sense. In particular to BMPString as [RFC 5280](https://tools.ietf.org/html/rfc5280) says:

> When encoding attribute values of type DirectoryString, conforming
   CAs MUST use PrintableString or UTF8String encoding, with the
   following exceptions

The exceptions are for pre-existing legacy CAs that use other encodings. Anyway, to compare the DNs properly, what should be implemented is probably what's described in [section 7.1 of RFC 5280](https://tools.ietf.org/html/rfc5280#section-7.1). But that'd be quite a lot of work.

And please re-read the last paragraph of my reply above.

ramon-garcia:
Attached requested.log.txt.gz : debug log with log level enc and ans = 3
IP Address 192.168.56.1 is Linux machine with strongswan, IP Address 192.168.56.2 is Windows 2012 R2

[requested.log.txt.gz](https://github.com/strongswan/strongswan/files/508263/requested.log.txt.gz)

Certificates:

CA [cacert.pem.txt](https://github.com/strongswan/strongswan/files/508270/cacert.pem.txt)
Strongswan side 
[linux_cert.pem.txt](https://github.com/strongswan/strongswan/files/508273/linux_cert.pem.txt)
Windows 2012R2 
[windows2012r2_cert.pem.txt](https://github.com/strongswan/strongswan/files/508275/windows2012r2_cert.pem.txt)

With regard to being careful of avoiding depending on early match, during debugging, I think that al the certificates where tested, the code in identification.c:compare_dn was invoked comparing the dn in the request with each one of the certificates, not just CA certificates. Maybe I overlooked something.

tobiasbrunner:
Hm, yeah looks like Windows, for no apparent reason, changes the encoding of the CA certificate's subject when sending a certificate request for it (the original encoding uses UTF8String). This seems wrong to me. [RFC 4945, section 3.2.7.1](https://tools.ietf.org/html/rfc4945#section-3.2.7.1), for example, states:
> Implementations SHOULD populate the Certification Authority field
   with the Subject field of the trust anchor, populated such that
   binary comparison of the Subject and the Certification Authority will
   succeed.

I'm a bit reluctant to add this code. I'd rather we refactor the whole DN comparison some time in the future as mentioned earlier.

The simplest workaround for you is probably to configure _leftsendcert=always_ so strongSwan sends its certificate also when not receiving a matching certificate request.

(By the way, you posted a private key above in `windows2012r2_cert.pem.txt`)

ramon-garcia:
So, what options do we have?
What changes should I do so that this patch is accepted?

Note that Windows has a very large amount of users. An issue with Windows 2012R2 is also likely to happen with Windows 8.1 (essentially the same system) and later versions. Microsoft might have incentives to be incompatible with Strongswan different from Strongswan ones.

Is there any more general way (but a reasonable work) that you would prefer?

tobiasbrunner:
> So, what options do we have?
> What changes should I do so that this patch is accepted?

You'd definitely need to change the license first (again, I refer to the link above). Otherwise, we can't start integrating any part of your code, if we decide to do so.

But doesn't the workaround I described above work for you?

> Note that Windows has a very large amount of users.

Does Windows not also support IKEv2? At least via _SC-VMM_ it seems possible to configure this (see e.g. [this technet article](https://blogs.technet.microsoft.com/networking/2014/12/26/vpn-interoperability-guide-for-windows-server-2012-r2/))

> An issue with Windows 2012R2 is also likely to happen with Windows 8.1 (essentially the same system) and later versions.

You are the first one to report this issue and these systems have been out there for quite a while. And I guess on client systems IKEv2 is generally used. Which client/function are you using in Windows? Do you configure it via _Windows Firewall with Advanced Security_?

ramon-garcia:
With regard to licensing, I hereby grant to to the Strongswan team, permission to use this code under the MIT license.

Is there any procedure that I should follow? I did not add a license text becuase I added no file.

ramon-garcia:
With regard to the IPSEC Windows configuration, I used the normal method for configuring IPSEC in LAN (as opposed to VPN tunnels): one opens with "Windows firewall with advanced security" -> Connection security rules -> New Rule.

Nowhere I found anything that allows one to select IKEv2.

I guess that the reason is compatibility with Windows 7 desktops and Windows 2008 R2 servers (the most common deployed type) that do not support IKEv2 (see here https://technet.microsoft.com/en-us/library/hh831365(v=ws.11).aspx )

In order to configure IPSEC policies with IKEv2, one has to use command line tools https://technet.microsoft.com/en-us/library/hh831807 .

The use case that I am working with is this one, placing Linux and Windows machines in the same LAN and encrypt traffic with IPSEC in transport model.

ramon-garcia:
Also note, that this patch is not just for me. The purpose is to make it easy for someone, with a normal Windows installation, a normal Windows network with IPSec policies, and a normal Strongswan configuration, that it works, without having to use complicated tricks.

ramon-garcia:
Hello?

tobiasbrunner:
> Is there any procedure that I should follow? I did not add a license text becuase I added no file.

Read the docs again. We still require you to add it in case a file is significantly modified. Also, please rebase your branch (squash those fixup commits etc.).

> that do not support IKEv2

They do, but only via the Agile VPN client (RAS) I guess. Not when configured via firewall.

> Also note, that this patch is not just for me.

Well, you are so far the first and only one who has this problem (but I suppose using Windows clients with Linux servers with transport mode SAs is a rare setup).

> without having to use complicated tricks.

What's complicated about adding `leftsendcert=always` to the server config?

ramon-garcia:
Thank you for your answers. I will look at it as soon as possible.

ramon-garcia:
> Well, you are so far the first and only one who has this problem (but I suppose using Windows clients with Linux servers with transport mode SAs is a rare setup).

Perhaps I did not explain well the use case. One has a network with Linux and Windows machines, and would like to secure the communications among them. Windows provides a simple to use mechanism, with IPSec rules that can be applied to workstations via active directory. In a very straightforward way, one can encrypt all the network traffic among Windows machines.

The purpose of this patch is to make Linux machines take part in this encrypted communications.

PR #52, opened 2016-10-06
Fix crash in case of empty LDAP response for CRL

y-cann:
In case of empty LDAP result during CRL fetch (for example, due to wrong filter attribute in LDAP URI , or invalid LDAP configuration), call to ldap_result2error() with NULL value for "entry" lead to Strongswan crash.   
See assertion on http://www.openldap.org/devel/gitweb.cgi?p=openldap.git;a=blob;f=libraries/libldap/error.c;hb=HEAD#l256


tobiasbrunner:
Applied, thanks!

PR #54, opened 2016-11-14
send local certificate subject to ext_auth script.

acd16:
Needed local certificate subject for my ext_auth script, added required code and upstreaming it. Please let me know if the changes are acceptable or would need any changes.


tobiasbrunner:
Can't you set the local identity to the subject DN of the certificate?

PR #56, opened 2016-12-13
vici: Check for strongswan closing connection in Python

optix2000:
The Python vici library does not check if the socket is closed.
If strongswan closes the connection, recv_all() spins forever.

### DIFF
--- a/src/libcharon/plugins/vici/python/vici/protocol.py
+++ b/src/libcharon/plugins/vici/python/vici/protocol.py
@@ -33,7 +33,10 @@ class Transport(object):
         """Ensure to read count bytes from the socket"""
         data = b""
         while len(data) < count:
-            data += self.socket.recv(count - len(data))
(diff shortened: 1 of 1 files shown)


src/libcharon/plugins/vici/python/vici/protocol.py
```diff
@@ -34,6 +34,8 @@ def _recvall(self, count):
         data = b""
         while len(data) < count:
             data += self.socket.recv(count - len(data))
+            if not data:
```
tobiasbrunner: Checking this after a `+=` operation seems not ideal. It's possible that a small amount was read before the socket was closed resulting in the same spinning you intended to prevent.

optix2000:
Sorry my mistake.
Silly oversight on my part.

Fixed.

tobiasbrunner:
Applied, thanks!

PR #57, opened 2016-12-14
NetworkManger-strongswan: user can custom remote identery when use CA certificate

browsnet:
Android VPN Client can custom server identity, but linux users cannot do it.

Some iKEv2 VPN Service need this, if user not custom server identity,  error occurred.

"constraint check failed: identity * required"


Asgoret:
@martinwilli HI! Can you plz look at this?)

tobiasbrunner:
I pushed something to the _gh57-nm-remote-id_ branch.

PR #59, opened 2017-01-26
child-sa: Do not install mark on inbound kernel SA

ebirger:
The SA ID (src, dst, proto, spi, encaps) is unique on ingress.

As such, explicit inbound marking is not needed to match an SA.

On the other hand, requiring inbound SAs to use marks forces the
installation of a mechanism for marking traffic (e.g. iptables) based
on some criteria.

Defining the criteria becomes complicated, for example when required to
support multiple SAs from the same src, especially when traffic is UDP
encapsulated.

This commit removes the assignment of the child_sa mark_in to the inbound SA.

Policies can be arbitrated by existing means - e.g, via netfilter policy
matching or using VTI interfaces - without the need to classify the flows prior
to state matching.

Since the reqid allocator regards the mark value, there is no risk of matching
the wrong policy.

And as explicit marking was required for route-based VPN to work before this
change, it should not cause regressions in existing setups.

### DIFF
--- a/src/libcharon/sa/child_sa.c
+++ b/src/libcharon/sa/child_sa.c
@@ -479,7 +479,7 @@ static status_t update_usebytes(private_child_sa_t *this, bool inbound)
 				.dst = this->my_addr,
 				.spi = this->my_spi,
 				.proto = proto_ike2ip(this->protocol),
-				.mark = this->mark_in,
+				.mark = {},
 			};
 			kernel_ipsec_query_sa_t query = {};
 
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks, looks good. I guess the expectation might initially have been that the kernel would actually **set** the mark on inbound packets after decryption. Sadly that's not how it works, instead, the kernel requires the mark on the ESP packets in order to match them. So I suppose adding the marks really doesn't serve much purpose and only complicates things. I changed the patch a bit (no need to set the marks except in `install()`). Will probably be included in the next release.

ebirger:
Thank you very much for the quick response!

ebirger:
Accidentally closed PR. need to keep open for your tracking?

tobiasbrunner:
I'll close it when I push the commit to master.

PR #61, opened 2017-02-07
Add option to force remote host as remote ts

tanzeeb:
This option forces the remote traffic selectors to be converted from subnet
ranges to the remote host when initiating a connection. It is useful for
making connections to the Windows IKEv2 implementation, which does not
support subnet ranges as traffic selectors in transport mode.

### DIFF
--- a/conf/options/charon.opt
+++ b/conf/options/charon.opt
@@ -106,6 +106,13 @@ charon.flush_auth_cfg = no
 charon.follow_redirects = yes
 	Whether to follow IKEv2 redirects (RFC 5685).
 
+charon.force_host_ts = no
+	Force remote traffic selector to be the remote host when initiating a
+	CHILD_SA (IKEv2 only).
+
+	This is useful for transport mode connections with implementations that do
+	not support subnet range negotiation (eg. Windows).
+
 charon.fragment_size = 1280
 	Maximum size (complete IP datagram size in bytes) of a sent IKE fragment
 	when using proprietary IKEv1 or standardized IKEv2 fragmentation, defaults
--- a/src/libcharon/sa/ikev2/tasks/child_create.c
+++ b/src/libcharon/sa/ikev2/tasks/child_create.c
@@ -202,6 +202,11 @@ struct private_child_create_t {
 	 * whether we are retrying with another DH group
 	 */
 	bool retry;
+
+	/**
+	 * Force host as remote traffic selector
+	 */
+	bool force_host_ts;
 };
 
 /**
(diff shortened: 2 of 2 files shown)


tobiasbrunner:
> which does not support subnet ranges as traffic selectors in transport mode.

What exactly do you mean? Transport mode is inherently host-to-host. So how did you configure this? Or doesn't it even support narrowing from `0.0.0.0/0` i.e. `%dynamic` to single IPs? Is there a NAT involved?

tanzeeb:
@tobiasbrunner yes, the windows IKEv2 implementation doesn't support traffic selector narrowing in transport mode. We sent a trace to Microsoft support, they sent us this snippet from the logs:

```
[3] 04C4.0F10::01/19/17-23:55:38.8773304 [ikeext] IkeValidateIkeV2Address() - 110|10.0.16.4|TS processing failure: IkeV2 transport mode implementation can only handle a single IP address. IP Ranges not handled
```

In our test network, we have a few Linux hosts (10.x.x.x) and Windows hosts (172.31.x.x), and want to enable host-to-host encryption between them on demand. We also have some hosts that are excluded (eg. `10.0.16.1`).

Linux `ipsec.conf`:
```
conn no-ipsec-10.0.16.1/32
    auto=route
    type=passthrough
    right=10.0.16.1/32
    rightsubnet=10.0.16.1/32

conn ipsec-10.0.0.0/8
   # auth and other config snipped
    auto=route
    keyexchange=ike
    type=transport
    right=10.0.0.0/8
    rightsubnet=10.0.0.0/8

conn ipsec-172.31.0.0/16
   # auth and other config snipped
    auto=route
    keyexchange=ike
    ike=aes128-sha1-modp1024!
    esp=aes128-sha1!
    type=transport
    right=172.31.0.0/16
    rightsubnet=172.31.0.0/16
```
Linux `strongswan.conf`:
```
charon {
  # other config snipped
  ignore_acquire_ts = yes
  force_host_ts = yes
}
```

On Windows, we setup one host with StrongSwan Windows and another with `New-NetIPsecRule -KeyModule=IKEv2`. The firewalls have a rule to allow all traffic between the two subnets. We have tried configurations with both the remote subnet ranges (eg. 10.0.16.0/24) and specific IPs (eg. 10.0.16.4) and the results were the same.

The behaviour is as follows:
* When we initiate a ping from Windows to Linux, the ping goes through.
* When we initiate a ping from Linux to Windows, the ping hangs, From the syslog on Linux and Wireshark on Windows, we can see the IKE_SA being established, but the CHILD_SA cannot be established because Windows rejects the traffic selector (`172.31.0.0/16`) and does not send a TSr payload back. 

We've also tried the following configuration changes:
* `rightsubnet=%dynamic` on Linux: This works because StrongSwan sends the specific remote IP as the TSr, but the trap it installs captures *all* traffic to and from the machine and assumes it should be encrypted. We want some connections to be `passthrough` and others to use stronger encryption algorithms if supported (eg. linux-to-linux).
* `keyexchange=ikev1` and `New-NetIPsecRule -KeyModule=IKEv1`: This works fine, but we would prefer to use IKEv2

We have opened a support ticket with Microsoft, and they confirmed that their IKEv2 transport mode implementation does not support traffic selector narrowing, and provided the logs above. With the patch, initiating a connection from Linux-Windows works with both Strongswan Windows and `New-NetIPsecRule`.

tanzeeb:
@tobiasbrunner We've tried the following config with the subnets being `%dynamic`:

```
conn no-ipsec-10.0.16.1/32
    auto=route
    type=passthrough
    right=10.0.16.1/32

conn ipsec-10.0.0.0/8
   # auth and other config snipped
    auto=route
    keyexchange=ike
    type=transport
    right=10.0.0.0/8

conn ipsec-172.31.0.0/16
   # auth and other config snipped
    auto=route
    keyexchange=ike
    ike=aes128-sha1-modp1024!
    esp=aes128-sha1!
    type=transport
    right=172.31.0.0/16
```

But this results in `ip xfrm state`s that trap *all* the traffic, not traffic within the range specified in the `right` field.

Is there a better way for us to approach the issue?

tobiasbrunner:
Sorry, for the delay.

Usually, with transport mode the traffic selectors are not specified explicitly (i.e. `left|rightsubnet` is implicitly set to _%dynamic_), which means that the source and destination traffic selectors are replaced with the actual endpoint addresses.

What you use, though, are wildcard trap policies, i.e. you specify a remote subnet with _auto=route_, which causes the installation of trap policies, and only when the kernel triggers an acquire based on matching traffic will the actual endpoints be known. But since the remote TS is not _%dynamic_ but a subnet it currently does not get replaced with the peer's IP address.

With strongSwan at the remote end that's not a problem if the peer's `leftsubnet` is _%dynamic_ as that gets replaced with the remote's IP address and that in turn will narrow the proposed traffic selector.

But I think proposing a subnet for a transport mode SA does not really make sense in the first place. So a new option might not actually be required, instead we could simply always do what we already do for dynamic traffic selectors, i.e. set them to the endpoint's addresses.

I pushed a couple of commits to the _gh61-transport-narrow_ branch. Would be great if you could test them.

tanzeeb:
The commits in `gh61-transport-narrow` work perfectly with our `rightsubnet`/`auto=route` configuration.

Thank you!

tobiasbrunner:
OK, great. Thanks for testing the changes.

PR #62, opened 2017-02-08
forecast: mark correct port in UDP NAT-T

abrasive:
The current code sets up a mark rule with rport:lport as the source port range, which breaks routing (unless rport < lport).


tobiasbrunner:
Applied, thanks!

PR #63, opened 2017-02-14
Add Tea-CI

stephengroat:
Add Tea-CI to build strongswan in mingw shells (32 and 64 bit)

Could also add windows command line tests in the future

example build: https://tea-ci.org/stephengroat/strongswan

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,12 @@
+build:
+  image: teaci/msys$$arch
+  pull: true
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
> Could also add windows command line tests in the future

Without that I don't really see an advantage over the current build on Travis-CI. I guess you could check for `DRONE` in `scripts/test.sh` and don't reset `TARGET` in that case to run the tests?

PR #64, opened 2017-03-02
Add option to install optionally-enforced polices

tanzeeb:
Add an option to kernel-netlink to install policies with the level 'use'
instead of 'required'.

This is useful for introducing IPsec to an existing cluster.

### DIFF
--- a/conf/plugins/kernel-netlink.opt
+++ b/conf/plugins/kernel-netlink.opt
@@ -38,6 +38,9 @@ charon.plugins.kernel-netlink.parallel_xfrm = no
 charon.plugins.kernel-netlink.policy_update = no
 	Whether to always use XFRM_MSG_UPDPOLICY to install policies.
 
+charon.plugins.kernel-netlink.policy_optional = no
+	Whether the policy level is 'use' instead of 'required'
+
 charon.plugins.kernel-netlink.port_bypass = no
 	Whether to use port or socket based IKE XFRM bypass policies.
 
--- a/src/libcharon/plugins/kernel_netlink/kernel_netlink_ipsec.c
+++ b/src/libcharon/plugins/kernel_netlink/kernel_netlink_ipsec.c
@@ -333,6 +333,11 @@ struct private_kernel_netlink_ipsec_t {
 	bool policy_update;
 
 	/**
+	 * Whether the policy level is 'use' instead of 'required'
+	 */
+	bool policy_optional;
+
+	/**
 	 * Installed port based IKE bypass policies, as bypass_t
 	 */
 	array_t *bypass;
@@ -2377,8 +2382,9 @@ static status_t add_policy_internal(private_kernel_netlink_ipsec_t *this,
 			tmpl->id.proto = protos[i].proto;
 			tmpl->aalgos = tmpl->ealgos = tmpl->calgos = ~0;
 			tmpl->mode = mode2kernel(proto_mode);
-			tmpl->optional = protos[i].proto == IPPROTO_COMP &&
-							 policy->direction != POLICY_OUT;
+			tmpl->optional = this->policy_optional ||
+							 (protos[i].proto == IPPROTO_COMP &&
+							 policy->direction != POLICY_OUT);
 			tmpl->family = ipsec->src->get_family(ipsec->src);
 
 			if (proto_mode == MODE_TUNNEL || proto_mode == MODE_BEET)
@@ -3204,6 +3210,8 @@ kernel_netlink_ipsec_t *kernel_netlink_ipsec_create()
 							  "kernel_netlink_get_priority_custom"),
 		.policy_update = lib->settings->get_bool(lib->settings,
 					"%s.plugins.kernel-netlink.policy_update", FALSE, lib->ns),
+		.policy_optional = lib->settings->get_bool(lib->settings,
+					"%s.plugins.kernel-netlink.policy_optional", FALSE, lib->ns),
 		.install_routes = lib->settings->get_bool(lib->settings,
 							"%s.install_routes", TRUE, lib->ns),
 		.proto_port_transport = lib->settings->get_bool(lib->settings,


tobiasbrunner:
Could you write more about the use case for this?

tanzeeb:
@tobiasbrunner Sure! 

Given we've got a cluster of machines (`machine-1`, `machine-2`, and `machine-3`) and we want to add transport-mode IPsec to all of them, we can use the optional policy flag to add IPsec with minimal downtime.

Currently, the procedure for adding IPsec to the cluster is as follows:

1. Add strongSwan to `machine-1`. It won't be able to talk to `machine-2` and `machine-3`.
2. Add strongSwan to `machine-2`. It will be able to talk to `machine-1` but wont be able to talk to `machine-3`.
3. Add strongSwan to `machine-3`. It will be able to talk to `machine-1` and `machine-2`.

As you can see, the cluster won't be able to communicate fully until IPsec is enabled on all machines.

With the `policy_optional` flag, the procedure would be as follows:

1. Add strongSwan to `machine-1` with `policy_optional = yes` . It will be able to talk to `machine-2` and `machine-3`.
2. Add strongSwan to `machine-2` with `policy_optional = yes`. It will be able to talk to `machine-1` and `machine-3`.
3. Add strongSwan to `machine-3` with `policy_optional = yes`. It will be able to talk to `machine-1` and `machine-2`.
4. Set `policy_optional = no` to `machine-1`. It will still be able to talk to `machine-2` and `machine-3`.
5. Set `policy_optional = no` to `machine-2`. It will still be able to talk to `machine-1` and `machine-3`.
6. Set `policy_optional = no` to `machine-3`. It will still be able to talk to `machine-1` and `machine-2`.

Even though there are more steps to the deployment, we are able to do it with minimal downtime in communication between the machines.

tobiasbrunner:
Couldn't you achieve the same thing, and even more controlled, with passthrough policies? Either configured with strongSwan or manually installed via `ip xfrm policy`. It would also not require a restart or a global option.

tanzeeb:
Do you mean something like this?

1. Add strongSwan to `machine-1`, passthrough to `machine-2` and `machine-3`. It will be able to talk to `machine-2` and `machine-3`.
2. Add strongSwan to  `machine-2`, IPsec to `machine-1` and passthrough to `machine-3`. It will not be able to `machine-1`, but will be able to talk to `machine-3`.
3. On `machine-1`, switch from passthrough to IPsec for `machine-2`. It will now be able to talk to `machine-2` again.
4. Add strongSwan to `machine-3`, IPsec to `machine-1` and `machine-2`. It will be not be able to talk to `machine-1` or `machine-2`.
5. On `machine-1`, switch from passthrough to IPsec for `machine-3`. It will now be able to talk to `machine-3` again.
6. On `machine-2`, switch from passthrough to IPsec for `machine-3`. It will now be able to talk to `machine-3` again.

Is there a way to do this without incurring the downtime between steps 2-3, and 4-6?

tobiasbrunner:
If you install the passthrough policies manually they will just be replaced by the actual IPsec policies installed by strongSwan once the connections get established (given the selectors match). You'd see a message like `policy already exists, try to update it` when this happens (strongSwan will then just replace and "adopt" the manually installed policies).

tanzeeb:
Manually installing the passthrough policy before starting strongSwan is not working for us. We are using `auto=route`, which makes strongSwan install the IPsec policy as soon as it starts. Is there something we're missing?

tanzeeb:
Also, would it be better to configure this through `ipsec.conf` / `swanctl.conf` instead of `strongswan.conf`?

tobiasbrunner:
> We are using auto=route, which makes strongSwan install the IPsec policy as soon as it starts. Is there something we're missing?

Aren't you using the wildcard trap policies you used before anymore? That is, do you now have specific host-host trap policies that would match the passthrough policies?

Would it be an option to use `auto=add` instead of `auto=route` while you install strongSwan on all hosts and then use `ipsec route` once that's done to actually install the trap policies? (Or change the config to `auto=route` and run `ipsec update`.) Since you won't actually have any protection until you disable `policy_optional` (optional trap policies just act like passthrough policies) that would have about the same effect.

tanzeeb:
Thanks for your suggestion, we will try `auto=add`. If this makes strongSwan respond to incoming IKE requests but not try to initiate outgoing IKE requests, it can work for us. Will report back.

(Following is reference for anyone else reading this thread)

We're using the following configuration:

```
# specific hosts that will never have IPsec
conn no-ipsec-10.0.16.1/32
    auto=route
    type=passthrough
    right=10.0.16.1/32
    rightsubnet=10.0.16.1/32
    leftsubnet=10.0.16.4/32    # ip of current host

# Subnets that should have host-to-host ipsec
conn ipsec-10.0.16.0/24
    # <auth snipped>
    auto=route
    keyexchange=ike
    type=transport

    right=10.0.16.0/24
    rightid=%any
    rightsubnet=10.0.16.0/24
    leftsubnet=10.0.16.4/32  # ip of current host
```

The crux of the problem is that we aren't able to go on to each machine and change the configuration (ie. enforce IPsec) before the software on the machines try to talk to each other.

tanzeeb:
@tobiasbrunner `auto=add` worked perfectly, we didn't even need the passthrough policies. Thanks for your help!

tobiasbrunner:
OK, great.

tanzeeb:
Hi @tobiasbrunner ,

We've run into an issue where `auto=add` doesn't suffice. 

Given `optional-machine` is configured with `auto=add` and `mandatory-machine` is configured with `auto=route`:

If `optional-machine` pings `mandatory-machine` first, the ping fails because the outgoing connection will be attempted without IPsec. On failure, no attempt is made to establish an IPsec connection.

The reverse, pinging from `mandatory-machine` to `optional-machine` first, works as expected. The IPsec connection gets established, and subsequent pings from `optional-machine` to `mandatory-machine` use the IPsec connection.

With our patch, we configure both machines with `auto=route`, and `optional-machine` also gets `charon.plugins.kernel-netlink.policy_optional = yes`:

If `optional-machine` pings `mandatory-machine` first, it will attempt to send the packets unencrypted and also try to establish an IPsec connection. Once the IPsec connection is established, the packets go through successfully.

The main difference we see between `auto=add` and creating an optional policy is that in the latter, `charon` is notified and will attempt to establish a secure connection, but in the former, `charon` is not notified at all.

Is there another way to make this work without using optional policies?

PR #65, opened 2017-03-23
Avoid O(n^2) copy operations when concatenating netlink responses to one result struct

JiriHorky:
When constructing the result, all respones from netlink were concatenated
iterativelly, i.e. for each response, the previously acquired result was
copied to newly allocated place and current response appended to it.
This results in O(n^2) copy operations. Instead, we now check for the
total final length of the result struct and copy the individual
responses to it in one pass, i.e. in O(n) copy operations.

In particular, this issue caused very high CPU usage in memcpy()
function as the result struct is copied over and over. Common way
how to hit the issue is when having 1000+ ipsec routes and 5+
connecting clients a second. In that case, the memcpy() function can
take 50%+ of one CPU thread on a decent CPU and the whole charon daemon
is stuck just reading routes and concatenating them together (connecting
clients are blocked in that particular case as this is done under mutex).

After applying the patch, the CPU consumption of charon daemon dropped from 100% to 4% on some of our servers.  Similar issues were described in https://wiki.strongswan.org/issues/2055 and also mentioned in https://wiki.strongswan.org/issues/1491. Disabling insert_routes helps in some cases (where possible), but the trouble is not obtaining the rules from kernel, but rather very suboptimal way of handling multiple responses from netlink.

### DIFF
--- a/src/libcharon/plugins/kernel_netlink/kernel_netlink_shared.c
+++ b/src/libcharon/plugins/kernel_netlink/kernel_netlink_shared.c
@@ -306,6 +306,9 @@ static status_t send_once(private_netlink_socket_t *this, struct nlmsghdr *in,
 	struct nlmsghdr *hdr;
 	chunk_t result = {};
 	entry_t *entry;
+	size_t total_len = 0;
+	u_char * ptr;
+	int i;
 
 	in->nlmsg_seq = seq;
 	in->nlmsg_pid = getpid();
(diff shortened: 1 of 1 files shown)


JiriHorky:
It seem the travis-ci tests timed out on running unrelated tls-mult test suite. I am not sure what I can do about it.

tobiasbrunner:
Nice, thanks. I pushed a slightly cleaned up version to the _netlink-concat_ branch.

> It seem the travis-ci tests timed out on running unrelated tls-mult test suite. I am not sure what I can do about it.

Yes, the certificates used in these tests expired yesterday.

PR #66, opened 2017-04-14
Fix type error exception in Python3

odi79:
Line 66 yields "TypeError: can't concat bytes to str" using Python 3.4. Changing "request" to "requestdata" which is already prepared in line 65 fixes this. After the change, the code is still working in Python 2.7

### DIFF
--- a/src/libcharon/plugins/vici/python/vici/protocol.py
+++ b/src/libcharon/plugins/vici/python/vici/protocol.py
@@ -63,7 +63,7 @@ class Packet(object):
     @classmethod
     def _named_request(cls, request_type, request, message=None):
         requestdata = request.encode("UTF-8")
-        payload = struct.pack("!BB", request_type, len(request)) + request
+        payload = struct.pack("!BB", request_type, len(request)) + requestdata
         if message is not None:
             return payload + message
         else:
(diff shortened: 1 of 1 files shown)


src/libcharon/plugins/vici/python/vici/protocol.py
```diff
@@ -63,7 +63,7 @@ class Packet(object):
     @classmethod
     def _named_request(cls, request_type, request, message=None):
         requestdata = request.encode("UTF-8")
-        payload = struct.pack("!BB", request_type, len(request)) + request
+        payload = struct.pack("!BB", request_type, len(request)) + requestdata
```
tobiasbrunner: I think `len(request)` should also be changed. But an easier fix would be to just change `requestdata` to `request`. As it was before 22f08609f1b6aabdc1177e0570c943842478de72, which introduced `requestdata` but without actually using it (not sure why @martinwilli introduced the new variable, the original `request` isn't used in the function anywhere so reassigning the result of `encode()` to it should be fine - he even did so in other places, maybe just an oversight).

odi79: I suppose you're right. It seems just like an oversight. Thanks, Tobias.

tobiasbrunner:
Thanks. I squashed the two commits and applied the result to master.

PR #67, opened 2017-04-24
Enable AES_GCM_ICV16, PRF_HMAC_SHA2_512, ECP_256_BIT in default proposal ESP.

defunctio:
I'm one of the developers for [Algo VPN](https://github.com/trailofbits/algo).

> Algo VPN is a set of Ansible scripts that simplify the setup of a personal IPSEC VPN. It uses the most secure defaults available, works with common cloud providers, and does not require client software on most devices. 

We recently ran into the issue (https://github.com/trailofbits/algo/issues/263) that charon-nm does not permit custom cipher proposals. Unfortunately the default proposals for libcharon do not enable ciphers that meet our requirements.

Due to this it makes it impossible to use the NetworkManager strongswan plugin to establish connections to Algo deployments resulting in a poor user experience on Linux desktop clients. While I would like to implement custom proposal selection in charon-nm and the network manager strongswan plugin in the future; this initial PR focuses on a minimally invasive approach to solving the issue of supporting stronger proposal defaults.

### DIFF
--- a/src/libcharon/config/proposal.c
+++ b/src/libcharon/config/proposal.c
@@ -752,9 +752,6 @@ static bool proposal_add_supported_ike(private_proposal_t *this, bool aead)
 			switch (encryption)
 			{
 				case ENCR_AES_GCM_ICV16:
-					add_algorithm(this, ENCRYPTION_ALGORITHM, encryption, 128);
-					add_algorithm(this, ENCRYPTION_ALGORITHM, encryption, 256);
-					break;
 				case ENCR_AES_CCM_ICV16:
 				case ENCR_CAMELLIA_CCM_ICV16:
 					/* we assume that we support all AES/Camellia sizes */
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
The first few changes make no sense as they just copy code that is already there (you might want to look up how `switch` statements work in C :-).

Regarding the change to `proposal_create_default_aead`. Did you read the comment there? (ECP-256 is also only supported if the _openssl_ plugin is enabled). So I won't apply this change. You are free to submit that part of the patch downstream (Ubuntu, Fedora etc.) as they know what algorithms their kernel supports and whether they enabled the _openssl_ plugin. But if you do it might be preferable to change how the proposal is built in `nm_service.c` (i.e. don't use `proposal_create_default*` but create custom proposals, like we do for the Android app).

> While I would like to implement custom proposal selection in charon-nm and the network manager strongswan plugin in the future

This is definitely the better approach. I guess to simplify the implementation this could be a global option in strongswan.conf  (e.g. `charon-nm.esp_proposals`). See the _nm-proposals_ branch for some prototypical code. Making this configurable per connection in the GUI would be more flexible, but is also significantly more work.

defunctio:
Ack, that is clearly my mistake with the dup code. :man_facepalming:

I did notice the comment about kernel support and I agree the best place for this quick patch is downstream where these conditions are already known. I expected this would be rejected due to compatibility issues but thought it best that it's reviewed members of strongswan. 

I do have unfinished changes towards charon-nm that implement custom ike/esp proposals in `nm_service.c` via `nm_setting_vpn_get_data_item` but not changes to the strongwan nm plugin. So it is only configurable via nmcli, etc but is functional and does not introduce the same compatibility problems. As you said though, this will take more time and effort.

Appreciate the feedback and information.

PR #68, opened 2017-04-26
swanctl: Add --list_conn_reqid command

Ferryworld:
list currently active connection name and reqid map

### DIFF
--- a/src/swanctl/Makefile.am
+++ b/src/swanctl/Makefile.am
@@ -8,6 +8,7 @@ swanctl_SOURCES = \
 	commands/redirect.c \
 	commands/install.c \
 	commands/list_sas.c \
+	commands/list_conn_reqid.c \
 	commands/list_pols.c \
 	commands/list_authorities.c \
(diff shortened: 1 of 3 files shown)


tobiasbrunner:
What exactly is the purpose of this command? Why not just use one of the bindings and directly do this via VICI.

Ferryworld:
connection name and reqid mapping is important for fastpath design.
reqid is the keyword to find a special connection in fastpath, then useful information can be located for statistics in communications.

tobiasbrunner:
But I don't really see the point in adding such a specialized command, which only outputs information already available to users via `--list-sas`, to swanctl. And when some kind of automated querying is the goal then you have to parse this output again anyway, when you can much more easily use one of the VICI bindings and integrate this kind of query directly into your application or workflow.

Ferryworld:
Indeed, it is duplicate in the swanctl point of view.
What this request accomplished is to facilitate application.

PR #69, opened 2017-04-30
charon-nm IKE/ESP proposal support

defunctio:
Adds support for IKE/ESP proposals passed via vpn data.

vpn data parameters;
* `proposal` 
  * [ yes / no ]
* `esp` / `ike`
  *  A `;` delimited list of proposals, supporting the `!` suffix

Similar to `ipsec.conf` a `!` is accepted as a suffix to restrict proposals.

_Note: A `;` is used instead of a `,` as a delimiter due to nmcli's inability to pass `,` in values._

### DIFF
--- a/src/charon-nm/nm/nm_service.c
+++ b/src/charon-nm/nm/nm_service.c
@@ -283,9 +283,10 @@ static gboolean connect_(NMVPNPlugin *plugin, NMConnection *connection,
 	NMStrongswanPluginPrivate *priv;
 	NMSettingConnection *conn;
 	NMSettingVPN *vpn;
+	enumerator_t *enumerator;
 	identification_t *user = NULL, *gateway = NULL;
 	const char *address, *str;
-	bool virtual, encap;
+	bool virtual, encap, proposal, restricted=FALSE;
 	ike_cfg_t *ike_cfg;
 	peer_cfg_t *peer_cfg;
 	child_cfg_t *child_cfg;
@@ -540,8 +541,23 @@ static gboolean connect_(NMVPNPlugin *plugin, NMConnection *connection,
 							 charon->socket->get_port(charon->socket, FALSE),
 							(char*)address, IKEV2_UDP_PORT,
 							 FRAGMENTATION_YES, 0);
-	ike_cfg->add_proposal(ike_cfg, proposal_create_default(PROTO_IKE));
-	ike_cfg->add_proposal(ike_cfg, proposal_create_default_aead(PROTO_IKE));
+
+	str = nm_setting_vpn_get_data_item(vpn, "proposal");
+	proposal = str ? streq(str, "yes") : FALSE;
+
+	if(proposal) {
+		if((str = nm_setting_vpn_get_data_item(vpn, "ike"))) {
+			restricted = (bool)g_str_has_suffix(str, "!");
+			enumerator = enumerator_create_token(str, ";", "!");
+			while (enumerator->enumerate(enumerator, &str)) {
+				ike_cfg->add_proposal(ike_cfg, proposal_create_from_string(PROTO_IKE, str));
+			}
+		}
+	}
+	if(!restricted) {
+		ike_cfg->add_proposal(ike_cfg, proposal_create_default(PROTO_IKE));
+		ike_cfg->add_proposal(ike_cfg, proposal_create_default_aead(PROTO_IKE));
+	}
 
 	peer_cfg = peer_cfg_create(priv->name, ike_cfg, &peer);
 	if (virtual)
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
> Similar to ipsec.conf a ! is accepted as a suffix to restrict proposals.

I'd actually prefer if this was changed to the semantics of [swanctl.conf](https://wiki.strongswan.org/projects/strongswan/wiki/Swanctlconf). That is, a custom proposal is always exclusive, only if `default` is explicitly added to the list will the default proposal be added (that option might not even be necessary). Because usually if a custom proposal is configured it's done to actually use exactly that proposal (and ipsec.conf is a legacy interface anyway). Also, make sure `proposal_create_from_string()` does not return NULL and that a proposal was added (and if not add the default proposal), otherwise, if the configured proposals are invalid no proposals would be sent (see e.g. the _nm-proposals_ branch). It will also allow users to only change one of the two proposals by leaving the other text field empty (which raises the question whether the check box is required, but perhaps it makes it clearer to the user).

There are also some code style issues (see [our wiki](https://wiki.strongswan.org/projects/strongswan/wiki/ProgrammingStyle)).

> Note: A ; is used instead of a , as a delimiter due to nmcli's inability to pass , in values.

That's unfortunate. Is this documented somewhere? (Because it seems that e.g. DNS servers and search domains may be passed as comma separated list in the NM IP settings).

defunctio:
>I'd actually prefer if this was changed to the semantics of swanctl.conf. That is, a custom proposal is always exclusive, only if default is explicitly added to the list will the default proposal be added (that option might not even be necessary). Because usually if a custom proposal is configured it's done to actually use exactly that proposal (and ipsec.conf is a legacy interface anyway). 

Sounds good to me, especially if it's considered legacy and will eventually be phased out.

>Also, make sure proposal_create_from_string() does not return NULL and that a proposal was added (and if not add the default proposal), otherwise, if the configured proposals are invalid no proposals would be sent (see e.g. the nm-proposals branch).

Would it not be better to generate an error instead of falling back to the default proposal? Accepting the default proposal in this event would be undesirable from my perspective. It could even result in issues where a typo or mistake is made in the user provided proposal not being caught and falling back to defaults instead of properly being enforced. 

> It will also allow users to only change one of the two proposals by leaving the other text field empty (which raises the question whether the check box is required, but perhaps it makes it clearer to the user).

I considered the necessity of the checkbox as well. Originally I didn't add one but it seemed like a bad UX decision as it could lead users to think it is a requirement or just add to confusion. I assume the likelihood that someone would want to change only one of the proposals was probably not that high. We could add a default to the proposal if the box is checked and the input is blank. But this is just my perspective, let me know what you think is best. 

> There are also some code style issues (see our wiki).

:+1: 

> That's unfortunate. Is this documented somewhere? (Because it seems that e.g. DNS servers and search domains may be passed as comma separated list in the NM IP settings).

I don't believe it is documented. `ipv4.dns` is a simple text input while `vpn.data` is a csv key/value store for vpn options. Unfortunately since it uses a comma to delimit it, one can't be used in the values. From testing it didn't appear that values could be escaped either.

defunctio:
@tobiasbrunner updated, awaiting review.

tobiasbrunner:
> Would it not be better to generate an error instead of falling back to the default proposal? Accepting the default proposal in this event would be undesirable from my perspective. It could even result in issues where a typo or mistake is made in the user provided proposal not being caught and falling back to defaults instead of properly being enforced.

Agreed. There is some cleanup code missing, though, when you return there (e.g. destroying the ike_cfg_t and later the child/peer_cfg_t objects).

> Originally I didn't add one but it seemed like a bad UX decision as it could lead users to think it is a requirement or just add to confusion.

Yes, the checkbox definitely makes it clearer to the users that this is optional.

> I assume the likelihood that someone would want to change only one of the proposals was probably not that high.

Well, the default IKE proposal includes all the (relatively strong) algorithms supported by the daemon (or rather by the plugins it has loaded). But that's not the case for ESP (we can't query the kernel for the supported algorithms via Netlink). So it could be that someone only wants to change the ESP proposal (in particular because the default does not include any DH groups).

> We could add a default to the proposal if the box is checked and the input is blank.

Yeah, just add the default proposals, i.e. what you do in the update look fine.

> I don't believe it is documented. ipv4.dns is a simple text input while vpn.data is a csv key/value store for vpn options. Unfortunately since it uses a comma to delimit it, one can't be used in the values. From testing it didn't appear that values could be escaped either.

Actually, while the input in the DNS text field is separated by commas, the plugin (or whoever) stores it separated with semicolons (i.e. if you look at the stored settings in `/etc/NetworkManager/system-connections`). Back in the GUI the values are separated by commas again. That's at least the case in the older version I'm using (perhaps they changed something with NM 1.2). Maybe we could somehow do the same. That would allow users to copy proposal strings from other configs without having to replace all commas with semicolons themselves.

defunctio:
>Agreed. There is some cleanup code missing, though, when you return there (e.g. destroying the ike_cfg_t and later the child/peer_cfg_t objects).

I think that should take care of it, though I haven't tested it yet.

_Also to note; [nm_service.c#L643](https://github.com/defunctio/strongswan/blame/c460dc63f12008c864d6d35f0fd8269d5ed4a39f/src/charon-nm/nm/nm_service.c#L643
) and [nm_service.c#L670](https://github.com/defunctio/strongswan/blame/c460dc63f12008c864d6d35f0fd8269d5ed4a39f/src/charon-nm/nm/nm_service.c#L670)_

>Actually, while the input in the DNS text field is separated by commas, the plugin (or whoever) stores it separated with semicolons (i.e. if you look at the stored settings in /etc/NetworkManager/system-connections). Back in the GUI the values are separated by commas again. That's at least the case in the older version I'm using (perhaps they changed something with NM 1.2). Maybe we could somehow do the same. That would allow users to copy proposal strings from other configs without having to replace all commas with semicolons themselves.

I'll change to alter that on the plugin side so the underlying processing in charon-nm will remain the same. This makes it still possible to function with nmcli but transparent to users of the plugin so they can just paste the original proposal strings from elsewhere.

tobiasbrunner:
> I think that should take care of it, though I haven't tested it yet.
> 
> Also to note; nm_service.c#L643 and nm_service.c#L670

Actually, these objects (e.g. _gateway_, _user_ or _auth_) are passed to other objects and are their responsibility afterwards. So depending on where you report the error you really only have to clean up the root object of that hierarchy (e.g. _peer_cfg_). Otherwise, you are causing double frees.

> I'll change to alter that on the plugin side so the underlying processing in charon-nm will remain the same. This makes it still possible to function with nmcli but transparent to users of the plugin so they can just paste the original proposal strings from elsewhere.

Sounds good.

tobiasbrunner:
Thanks for your work on this. I've pulled the changes here and in #70 to master (with some squashing and cleanups) so this will be included in our next release.

PR #70, opened 2017-04-30
network manager strongswan IKE/ESP proposal support

defunctio:
Adds custom IKE/ESP proposal support for the strongswan network manager plugin.
This PR relies on changes made in PR #69.

image

### DIFF
--- a/src/frontends/gnome/properties/nm-strongswan-dialog.ui
+++ b/src/frontends/gnome/properties/nm-strongswan-dialog.ui
@@ -148,6 +148,9 @@
                 <property name="column_spacing">6</property>
                 <property name="row_spacing">6</property>
                 <child>
+                  <placeholder/>
+                </child>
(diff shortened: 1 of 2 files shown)


src/frontends/gnome/properties/nm-strongswan-dialog.ui
```diff
@@ -148,6 +148,9 @@
                 <property name="column_spacing">6</property>
                 <property name="row_spacing">6</property>
                 <child>
+                  <placeholder/>
```
tobiasbrunner: How did that block get moved?  Is this necessary at all?

tobiasbrunner:
Looks pretty good. Requires some changes once #69 is updated (there are also some minor code style issues, e.g. missing spaces around `if`, but as you can see from the existing code the style is not enforced as strictly here).

PR #71, opened 2017-05-11
updown: Add rules (ipv4) to fix masquerading issues

sch-m:
This patch implements the feature already documented by the
left|rightfirewall option to exclude the ipsec related traffic from
masquerading.

Signed-off-by: Martin Schiller <ms@dev.tdt.de>

### DIFF
--- a/src/_updown/_updown.in
+++ b/src/_updown/_updown.in
@@ -255,6 +255,11 @@ up-host:iptables)
 	      -s $PLUTO_PEER -d $PLUTO_ME $IPSEC_POLICY_IN -j ACCEPT
 	fi
 	#
+	# do not masquerade IPsec traffic
+	iptables -t nat -I POSTROUTING 1 -p $PLUTO_PEER_PROTOCOL \
+	    -s $PLUTO_MY_CLIENT $S_MY_PORT \
+	    -d $PLUTO_PEER_CLIENT $D_PEER_PORT $IPSEC_POLICY_OUT -j ACCEPT
+	#
 	# log IPsec host connection setup
 	if [ $VPN_LOGGING ]
 	then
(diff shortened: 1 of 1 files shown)


sch-m:
I don't know what's the problem with this codecov/project check, but I think this has nothing to do with my commit. :-)

tobiasbrunner:
One problem I see with this is that it creates such a rule for every connection, where it is way easier and efficient to just add [one simple rule](https://wiki.strongswan.org/projects/strongswan/wiki/ForwardingAndSplitTunneling#Hosts-on-the-Internet) that covers all clients (which, in this scenario, are usually roadwarriors and may connect in large numbers). While the same applies to the other rules added by the script, these here are not really security relevant and only needed if an SNAT is in place. So there isn't any benefit in making them so specific and adding/removing them with each client connection. I'd actually rather remove the note in the man page than applying this patch.

sch-m:
I chose the way to add a rule for every connection analogous to the filter rules because i thought this would be the "cleanest" way . But yes, you are right. At least on the host side it would be much more efficient to just add one simple rule that covers all clients.

But I still think that the updown script would be the best solution to set this rule(s), so that the user just have to set the left|rightfirewall option to yes to get the vpn connection working without having to handle any firewall rules.

what about handling this by-connection based rules in the "up|down-client:iptables" cases like I did in my patch and simple add the generic rule in the "up-host:iptables" case?

Or maybe add the generic rule for client and host side?

tobiasbrunner:
> what about handling this by-connection based rules in the "up|down-client:iptables" cases like I did in my patch and simple add the generic rule in the "up-host:iptables" case?

The difference between `*-client` and `*-host` is that the latter is called if the local traffic selector is a single host. It's not necessarily a distinction between client and server (but in the usual roadwarrior scenario, with virtual IPs for clients and tunneling everything to the server, `*-host` is called on the client and `*-client` on the server).

And adding generic rules (i.e. rules that are only needed once for all connections) with the updown script is not easy as you'd have to add some kind of reference counting (the script is called for every combination of local and remote subnet), otherwise the first client disconnecting will remove the rule again.

sch-m:
Oh, ok. so I missunderstood this *-client and *-host design.

But I'm still searching for a solution where the user doesn't have to set any kind of ipsec-related firewall rules, because strongswan should do it for him when the left|rightfirewall option is used.

What about just adding the generic rule without removing it at all. I know this seems not a really clean way, but the rule shouldn't hurt anything else and is only added if the left|rightfirewall option is used. So no refcounting would be needed.

PR #72, opened 2017-05-11
af_alg: fix crypt() definition conflict

baruchsiach:
Rename the crypt() method to avoid conflict with POSIX crypt(). Fixes the
following build failure with musl libc:

In file included from ../../../../src/libstrongswan/utils/utils.h:53:0,
                 from ../../../../src/libstrongswan/library.h:101,
                 from af_alg_ops.h:24,
                 from af_alg_ops.c:16:
af_alg_ops.c:110:22: error: conflicting types for 'crypt'
 METHOD(af_alg_ops_t, crypt, bool,
                      ^
../../../../src/libstrongswan/utils/utils/object.h:99:13: note: in definition of macro 'METHOD'
  static ret name(union {iface *_public; this;} \
             ^
In file included from af_alg_ops.c:18:0:
.../host/usr/x86_64-buildroot-linux-musl/sysroot/usr/include/unistd.h:144:7: note: previous declaration of 'crypt' was here
 char *crypt(const char *, const char *);
       ^

### DIFF
--- a/src/libstrongswan/plugins/af_alg/af_alg_ops.c
+++ b/src/libstrongswan/plugins/af_alg/af_alg_ops.c
@@ -107,7 +107,7 @@ METHOD(af_alg_ops_t, hash, bool,
 	return TRUE;
 }
 
-METHOD(af_alg_ops_t, crypt, bool,
+METHOD(af_alg_ops_t, crypt_, bool,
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Applied, thanks!

PR #73, opened 2017-05-19
Store configure arguments in libstrongswan and extend swanctl to print it out

Thermi:
This PR implements the following changes:

- Declare the configuration arguments as a preprocessor definition in config.h
- Store the arguments in an attribute to the private interface of libstrongswan
- Declare a method to get the configure arguments from libstrongswan
- Extend swanctl to print out the configure arguments using "--get-config-args" (-o)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -39,9 +39,8 @@ PKG_PROG_PKG_CONFIG
 
 # set configure arguments
 AC_CONFIGURE_ARGS=`echo "${ac_configure_args}" | sed "s/ //"`
-# | sed "s/'/\\\'/g"`
-AC_DEFINE_UNQUOTED(AC_CONFIGURE_ARGS, "$AC_CONFIGURE_ARGS", "")
 
+AC_DEFINE_UNQUOTED(AC_CONFIGURE_ARGS, "$AC_CONFIGURE_ARGS", "")
 
 m4_include(m4/macros/split-package-version.m4)
 SPLIT_PACKAGE_VERSION
--- a/configure.ac
+++ b/configure.ac
@@ -37,6 +37,12 @@ AC_CONFIG_HEADERS([config.h])
 AC_DEFINE([CONFIG_H_INCLUDED], [], [defined if config.h included])
 PKG_PROG_PKG_CONFIG
 
+# set configure arguments
+AC_CONFIGURE_ARGS=`echo "${ac_configure_args}" | sed "s/ //"`
+# | sed "s/'/\\\'/g"`
+AC_DEFINE_UNQUOTED(AC_CONFIGURE_ARGS, "$AC_CONFIGURE_ARGS", "")
+
+
 m4_include(m4/macros/split-package-version.m4)
 SPLIT_PACKAGE_VERSION
 
--- a/src/swanctl/commands/get_config_args.c
+++ b/src/swanctl/commands/get_config_args.c
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C) 2017 Noel Kuntze
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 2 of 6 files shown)


src/swanctl/commands/get_config_args.c
```diff
@@ -0,0 +1,51 @@
...
+ */
+static void __attribute__ ((constructor))reg()
+{
+	command_register((command_t) {
+		get_configure_arguments, 'o', "get-config-args", "prints the ./configure arguments",
+		{"[--raw|--pretty]"},
+		{
+			{"help",		'h', 0, "show usage information"},
+			{"raw",			'r', 0, "dump raw response message. Doesn't do anything here."},
+			{"pretty",		'P', 0, "dump raw response message in pretty print. Doesn't do anything here."},
```
tobiasbrunner: "raw" and "pretty" make no sense here as there are no VICI messages involved.

src/swanctl/commands/get_config_args.c
```diff
@@ -0,0 +1,51 @@
...
+	return 0;
+}
+
+/**
+ * Register the command.
+ */
+static void __attribute__ ((constructor))reg()
+{
+	command_register((command_t) {
+		get_configure_arguments, 'o', "get-config-args", "prints the ./configure arguments",
```
tobiasbrunner: `config-args` should be enough, no need for the  `get-` (also rename the file etc.).

configure.ac
```diff
@@ -37,6 +37,12 @@ AC_CONFIG_HEADERS([config.h])
 AC_DEFINE([CONFIG_H_INCLUDED], [], [defined if config.h included])
 PKG_PROG_PKG_CONFIG
 
+# set configure arguments
+AC_CONFIGURE_ARGS=`echo "${ac_configure_args}" | sed "s/ //"`
+# | sed "s/'/\\\'/g"`
```
Thermi: Quoting of variables is just learned best practice. The intention is that the args can be used directly without any mangling to allow using it with scripts.

configure.ac
```diff
@@ -37,6 +37,12 @@ AC_CONFIG_HEADERS([config.h])
...
+AC_DEFINE_UNQUOTED(AC_CONFIGURE_ARGS, "$AC_CONFIGURE_ARGS", "")
```
Thermi: Will do.

configure.ac
```diff
@@ -37,6 +37,12 @@ AC_CONFIG_HEADERS([config.h])
 AC_DEFINE([CONFIG_H_INCLUDED], [], [defined if config.h included])
 PKG_PROG_PKG_CONFIG
 
+# set configure arguments
+AC_CONFIGURE_ARGS=`echo "${ac_configure_args}" | sed "s/ //"`
```
Thermi: Will do.

src/swanctl/commands/get_config_args.c
Thermi: The intention is to make this command behave similiar to the others to avoid having to handle this one in any special way, if swanctl is called programmatically using some scripting language.

src/swanctl/commands/get_config_args.c
Thermi: Will do.

Thermi:
Sure, we could store them in /usr/share instead. The question is if the distros that ship strongSwan would need to make any changes to their files to support this single command. I decided to store it in libstrongswan, because that's what was configured with the args. Storing it there would also make it possible to write a VICI command for that later, if it's required, and adding the functionality for `ipsec stroke`. Should I use strdup() instead of returning the `char *`?

src/swanctl/commands/get_config_args.c
tobiasbrunner: I don't agree, adding useless options is just confusing (here they could e.g. be understood mean printing the arguments either "pretty" or in some raw state, if that was the case it would at least make some sense, but it would still be confusing as this command has really nothing to do with VICI and its messages so you still don't get the output you'd expect from the other commands).

tobiasbrunner:
> The question is if the distros that ship strongSwan would need to make any changes to their files to support this single command.

So? On Debian/Ubuntu you'd e.g. get a warning if an installed file is not packed up.

> I decided to store it in libstrongswan, because that's what was configured with the args.

Not really. Many options have effects on other libraries or programs (e.g. all the plugins that are not in libstrongswan or CFLAGS passed to the script or cross-compile options).

> Storing it there would also make it possible to write a VICI command for that later, if it's required, and adding the functionality for ipsec stroke.

All other commands should easily be able to read a file in `/usr/share`.

Thermi:
>> The question is if the distros that ship strongSwan would need to make any changes to their files to support this single command.

> So? On Debian/Ubuntu you'd e.g. get a warning if an installed file is not packed up.

Okay, so that's not a problem.

>> I decided to store it in libstrongswan, because that's what was configured with the args.

>Not really. Many options have effects on other libraries or programs (e.g. all the plugins that are not in libstrongswan or CFLAGS passed to the script or cross-compile options).

Well, sure, but libstrongswan is the central lib of strongswan and I think the arguments that were used to configure strongSwan should be stored in something that directly relates to strongSwan. So I chose the lib. 

>> Storing it there would also make it possible to write a VICI command for that later, if it's required, and adding the functionality for ipsec stroke.

> All other commands should easily be able to read a file in /usr/share.

Yes. VICI can be spoken about TCP though (So from other systems) and I am concerned that it could cause confusion about what's configuration line is actually printed if one executes get_stored_arguments() from a remote system, if it's just read from a file by the local binary. I'd expect the remote peer's arguments to be printed.

configure.ac
Thermi: That variable isn't populated on my system. It's empty. For now, I use `sed` directly.

PR #74, opened 2017-06-19
Update task_manager_v1.c

avinoamm:
Instead of  deleting  redundent CHILD SA on rekey, just ignore the rekey and wait for hard expiration time.


tobiasbrunner:
Handled in [2358](https://wiki.strongswan.org/issues/2358) on the issue tracker. Since 083208e805e67aad2ff58bb0d297b46dfff4573c this depends on the `charon.delete_rekeyed` option.

PR #75, opened 2017-06-20
Add coverage flags for openssl plugin

securitykernel:
We want to look into the coverage of the openssl plugin.


tobiasbrunner:
I guess we could also extend `PLUGIN_CFLAGS` to enable coverage in all plugins if compiled with `--enable-coverage`. I did so in the _plugins-coverage_ branch. Let me know if that works for you too.

securitykernel:
Works fine for us, thanks! Closing here.

PR #77, opened 2017-07-07
make nopmtudisc function configurable

Ferryworld:


tobiasbrunner:
Why would you want to disable PMTUD?

Ferryworld:
PMTUD will fragment the package and some parts will be lost in some case.

tobiasbrunner:
> PMTUD will fragment the package and some parts will be lost in some case.

No, it's actually disabling PMTUD (which is what that flag indirectly does) that might potentially cause the ESP (or ESP-in-UDP) packet to get fragmented (as the DF flag will not be set, if it was set on the original packet, which might depend on different `sysctl` options in the first place). Disabling it will only work in cases where ICMPs are dropped (i.e. PMTUD won't work) but packets are then actually fragmented on the way if they are too big (not just dropped, which some routers/ISPs do even if the DF flag is not set) and these fragments are then also properly forwarded (and not dropped, which seems to happen a lot, judging by the reports we see regarding too large IKE_AUTH messages).

PR #78, opened 2017-07-28
child-sa: allow requesting different unique marks for in/out

ebirger:
When requiring unique flags for CHILD_SAs, allow the configuration to
request different marks for each direction by adding an exclamation point
to the %unique keyword.

This is useful when different marks are desired for each direction (e.g.
when 0.0.0.0/0 - 0.0.0.0/0 traffic selectors are used) but the number
of peers is not predefined.

### DIFF
--- a/conf/format-options.py
+++ b/conf/format-options.py
@@ -217,7 +217,7 @@ class TagReplacer:
 		return re.compile(r'''
 			(^|\s|(?P<brack>[(\[])) # prefix with optional opening bracket
 			(?P<tag>''' + tag + r''') # start tag
-			(?P<text>\w|\S.*?\S) # text
+			(?P<text>\S|\S.*?\S) # text
 			''' + tag + r''' # end tag
 			(?P<punct>([.,!:)\]]|\(\d+\))*) # punctuation
 			(?=$|\s) # suffix (don't consume it so that subsequent tags can match)
--- a/man/ipsec.conf.5.in
+++ b/man/ipsec.conf.5.in
@@ -1037,7 +1037,10 @@ mask of
 .B 0xffffffff
 is assumed. The special value
 .B %unique
-assigns a unique value to each newly created IPsec SA.
+assigns a unique value to each newly created IPsec SA. To additionally
+make the mark unique for each IPsec SA direction (in/out) the special value
+.B %unique-dir
+may be used.
 .TP
 .BR mark_in " = <value>[/<mask>]"
 sets an XFRM mark in the inbound IPsec SA and
--- a/src/libcharon/sa/child_sa.c
+++ b/src/libcharon/sa/child_sa.c
@@ -1719,7 +1719,7 @@ child_sa_t * child_sa_create(host_t *me, host_t* other,
 {
 	private_child_sa_t *this;
 	static refcount_t unique_id = 0, unique_mark = 0;
-	refcount_t mark;
+	refcount_t mark = 0;
 
 	INIT(this,
 		.public = {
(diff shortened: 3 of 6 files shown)


tobiasbrunner:
Thanks for the patch. Could you please describe a possible use case for this in more detail?

The _!_ suffix is not really ideal (too easy to overlook). I pushed a modified version to the _gh78-mark-unique-dir_ branch.

ebirger:
Hi Tobias,

Thanks for the review and changes! I definitely find 'unique-dir' to be more understandable.
I noticed you hadn't changed the 'MARK_UNIQUE_STRICT' macro name to reflect that change,
Would you like me to rename it (et al.) to MARK_UNIQUE_DIR?

Re use-case, I will add the following description to the commit message:

When implementing a site-to-site route-based VPN without VTI devices, the use of 
0.0.0.0/0-0.0.0.0/0 traffic selector with identical marks for the in and out directions
leads to packet drops.

The reason is that the 'out' and 'fwd' policies overlap (same TSs and marks) but the
underlay 'template' mismatches - the 'fwd' template is from 'remote' to 'local' whereas
the 'out' template is from 'local' to 'remote. When outbound traffic is matched against the
'fwd' policy, it is dropped.

Using different marks for each direction provides a way to distinguish between the
'out' and 'fwd' policies.

Thanks!
Eyal.

ebirger:
I have altered the original commit message and added an additional commit for renaming MARK_UNIQUE_STRICT to MARK_UNIQUE_DIR - to be squashed or discarded of at your liking.

tobiasbrunner:
> I noticed you hadn't changed the 'MARK_UNIQUE_STRICT' macro name to reflect that change

Yeah, I didn't think it was necessary because STRICT in the constant seemed to work equally well. But I've no problem with changing it either.

> The reason is that the 'out' and 'fwd' policies overlap (same TSs and marks)

Hm, but that's not true. The FWD policy looks like the IN policy, not the OUT policy (with regards to the traffic selectors and templates).

> When outbound traffic is matched against the 'fwd' policy, it is dropped.

There is the _policies_fwd_out_ option in [swanctl.conf](https://wiki.strongswan.org/projects/strongswan/wiki/Swanctlconf) to explicitly install FWD policies that look like OUT policies (i.e. two FWD policies will be installed). Does that work for your use case too (i.e. without separate marks)?

Also, usually the missing outbound FWD policy is only an issue when there is a FWD drop policy that would match the traffic (the additional FWD policy will then just act as passthrough policy). Because if there is no drop policy, the Linux kernel only cares for them when forwarding after decryption, hence the IN-FWD policy, but not when forwarding plain traffic (which will just match the OUT policy). I guess that could be due to the VTI device if it enforces FWD policies. What kernel version do you use?

ebirger:
> > The reason is that the 'out' and 'fwd' policies overlap (same TSs and marks)

> Hm, but that's not true. The FWD policy looks like the IN policy, not the OUT policy (with regards to the traffic selectors and templates).

Indeed the templates are different, and since the traffic selectors of the FWD policy do match the outbound traffic, the traffic is dropped.
I can see the 'XfrmInTmplMismatch' counter going up in /proc/net/xfrm_stat

> There is the policies_fwd_out option in swanctl.conf to explicitly install FWD policies that look like OUT policies (i.e. two FWD policies will be installed). Does that work for your use case too (i.e. without separate marks)?

I can check, though as the traffic selectors will still match both FWD policies, will the correct one be picked?

> Also, usually the missing outbound FWD policy is only an issue when there is a FWD drop policy that would match the traffic (the additional FWD policy will then just act as passthrough policy). Because if there is no drop policy, the Linux kernel only cares for them when forwarding after decryption, hence the IN-FWD policy, but not when forwarding plain traffic (which will just match the OUT policy). I guess that could be due to the VTI device if it enforces FWD policies. What kernel version do you use?

I'm using kernel 4.10. I think it also happened on 4.8 and 4.4.

ebirger:
Note I'm not using a VTI device, so can't disable policy check on FWD.

tobiasbrunner:
> I can check, though as the traffic selectors will still match both FWD policies, will the correct one be picked?

No, with 0.0.0.0/0 on both ends (or with any equal TS) only one FWD policy is installed anyway (looking like the IN policy).

> Note I'm not using a VTI device, so can't disable policy check on FWD.

I see. Yeah, with VTI devices policy matching is usually disabled (for that device) as it enforces the IN/OUT policy directly (via e.g. `sysctl -w net.ipv4.conf.<vtiname>.disable_policy=1`).

I'll apply the change to master. Thanks!

PR #79, opened 2017-08-30
Add basic stream service for systemd sockets

aszlig:
This systemd socket activation by passing URIs such as systemd://foo to plugins such as VICI.

For example setting `charon.plugins.vici.socket = systemd://vici`, a systemd socket file descriptor with the name `vici` will be picked up.

So these would be the corresponding unit options:

```ini
[Socket]
FileDescriptorName=vici
Service=strongswan.service

ListenStream=/run/openswan.vici
```

The implementation currently is very basic and right now only the first file descriptor for a particular identifier is picked up if there are multiple socket units with the same FileDescriptorName.

### DIFF
--- a/src/libstrongswan/Makefile.am
+++ b/src/libstrongswan/Makefile.am
@@ -96,6 +96,7 @@ fetcher/fetcher_manager.h eap/eap.h pen/pen.h ipsec/ipsec_types.h \
 networking/host.h networking/host_resolver.h networking/packet.h \
 networking/tun_device.h networking/streams/stream.h \
 networking/streams/stream_unix.h networking/streams/stream_service_unix.h \
+networking/streams/stream_service_systemd.h \
 networking/streams/stream_tcp.h networking/streams/stream_service_tcp.h \
 networking/streams/stream_service.h networking/streams/stream_manager.h \
 resolver/resolver.h resolver/resolver_response.h resolver/rr_set.h \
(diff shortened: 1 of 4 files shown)


tobiasbrunner:
Thanks for the patch.

Is the main point of this the service activation by sending a VICI message to that socket? If so, when exactly is that useful (compared to just starting the daemon on system boot)? Or what is the use case for this?

The example in the commit message uses the wrong *swan, and the default path (which clients like `swanctl` use by default) is `/var/run/charon.vici` (otherwise this has to be configured via `swanctl.socket` or `--uri`). Ideally, there would be an easy way for clients to lookup the socket via name as well, so a single configuration via `libstrongswan.plugins.vici.socket = systemd://vici` would work in the plugin and swanctl (that'd require a `stream_create_systemd()` constructor).

Please have a look at our [wiki page regarding contributions](https://wiki.strongswan.org/projects/strongswan/wiki/Contributions). Would be great if you considered resubmitting under the mentioned license.

aszlig:
> Is the main point of this the service activation by sending a VICI message to that socket? If so, when exactly is that useful (compared to just starting the daemon on system boot)? Or what is the use case for this?

It is useful for privilege separation. For example I'm running `charon` as normal user who has no network access other than UDP and netlink and the user also doesn't own the socket.

> The example in the commit message uses the wrong *swan

Oh, right, I wanted to write `/run/charon.vici` (`/var/run` is symlinked to `/run` on systemd systems).

> Please have a look at our wiki page regarding contributions. Would be great if you considered resubmitting under the mentioned license.

Sure, I'll do.

aszlig:
@tobiasbrunner: Amended the commit with the points you mentioned.

tobiasbrunner:
> It is useful for privilege separation. For example I'm running charon as normal user who has no network access other than UDP and netlink and the user also doesn't own the socket.

I see, nice. Pushed to the _gh79-stream-service-systemd_ branch, with some minor changes. I'll close the PR once it's applied to master. Thanks again!

PR #80, opened 2017-08-31
libipsec: Fix "packet too short" error when parsing fragmented packets

rossoneri91:
Should only parse the transport header of an IP packet if it's not fragmented or is the first fragment


tobiasbrunner:
Hm, wouldn't the `ip_off` field be non-zero in the first fragment, because the MF flag is set (which is part of the 16-bit field)? I pushed an alternative patch to the _gh80-libipsec-ip-frag_ branch (also includes a change for IPv6, where there was no such error but the next header was not correct for fragmented packets). Let me know if that works for you.

PR #83, opened 2017-11-28
Version field in TBSRequest section of OSCP Request.

HenadziMatuts:
**RFC 6960** describes _TBSRequest_ section of OSCP Request as:
```
TBSRequest ::= SEQUENCE {
       version             [0]     EXPLICIT Version DEFAULT v1,
       requestorName       [1]     EXPLICIT GeneralName OPTIONAL,
       requestList                 SEQUENCE OF Request,
       requestExtensions   [2]     EXPLICIT Extensions OPTIONAL }
```
This pull request adds missing _version_ field.

### DIFF
--- a/src/libstrongswan/plugins/x509/x509_ocsp_request.c
+++ b/src/libstrongswan/plugins/x509/x509_ocsp_request.c
@@ -239,7 +239,8 @@ static chunk_t build_requestExtensions(private_x509_ocsp_request_t *this)
  */
 static chunk_t build_tbsRequest(private_x509_ocsp_request_t *this)
 {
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
It's not missing. Default values are usually omitted in ASN.1 encodings.

HenadziMatuts:
My bad. Thought it should be defined explicitly.

tobiasbrunner:
No problem. The `EXPLICIT` (if that's what confused you) refers to the encoding/tagging *if* the value is given, as opposed to `IMPLICIT` (where the type of the value is not explicitly encoded).

PR #85, opened 2017-11-30
charon-nm: port to libnm

lkundrak:
libnm-glib is deprecated for several years and reaching the end of its
life. Let's switch to the more up-to-date library.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1174,10 +1174,7 @@ if test x$eap_sim_pcsc = xtrue; then
 fi
 
 if test x$nm = xtrue; then
-	PKG_CHECK_EXISTS([libnm-glib],
-		[PKG_CHECK_MODULES(nm, [NetworkManager gthread-2.0 libnm-util libnm-glib libnm-glib-vpn])],
-		[PKG_CHECK_MODULES(nm, [NetworkManager gthread-2.0 libnm_util libnm_glib libnm_glib_vpn])]
-	)
+	PKG_CHECK_MODULES(nm, [gthread-2.0 libnm])
 	AC_SUBST(nm_CFLAGS)
 	AC_SUBST(nm_LIBS)
 fi
--- a/src/charon-nm/nm/nm_backend.c
+++ b/src/charon-nm/nm/nm_backend.c
@@ -55,7 +55,7 @@ struct nm_backend_t {
 static nm_backend_t *nm_backend = NULL;
 
 /**
- * NM plugin processing routine, creates and handles NMVPNPlugin
+ * NM plugin processing routine, creates and handles NMVpnServicePlugin
  */
 static job_requeue_t run(nm_backend_t *this)
 {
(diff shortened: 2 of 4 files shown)


lkundrak:
Please note that this might not be ready for merging yet. I basically just compile-tested it, didn't give it any real test.

I eventually plan to do some testing, but I figured out it just might be a good idea to share this early.

tobiasbrunner:
Great, there already is a [bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=862885) for the Debian package concerning this. Looking forward to your update.

lkundrak:
Gave it some testing now.

Worked fine, apart from a hang on connection error -- the disconnect() method is now called on failure even when the plugin is not yet successfully connected.

I've added a condition in the disconnect() method which effectively restores the previous behaviour, but I'm not familiar with the code base enough to understand why the hang happened. Given the behaviour is same as previously, I guess that shouldn't be considered a blocker.

tobiasbrunner:
> Worked fine, apart from a hang on connection error -- the disconnect() method is now called on failure even when the plugin is not yet successfully connected.

When exactly is it called (i.e. what triggers it)? If it's triggered via e.g. `ike_state_change()` and `signal_failure()` then this could create a deadlock if the SA is still checked out and can't again by `disconnect()` (e.g. if any of the calls in `signal_failure()` are blocking until `disconnect()` returns). If that's the problem we could perhaps queue a job that does what `signal_failure()` does in a different thread so the caller can release the SA.

lkundrak:
> When exactly is it called (i.e. what triggers it)? If it's triggered via e.g. ike_state_change() and signal_failure() then this could create a deadlock if the SA is still checked out and can't again by disconnect() (e.g. if any of the calls in signal_failure() are blocking until disconnect() returns). If that's the problem we could perhaps queue a job that does what signal_failure() does in a different thread so the caller can release the SA.

Yes, that's indeed the case.

I've modified the patch to enqueue the actual disconnection and it works well now.

tobiasbrunner:
Great, thanks. I've pushed this to the _nm-charon-libnm_ branch. Can `disconnect()` be called multiple times? Even then, I guess `disconnect_id` is not strictly necessary. Calling `do_disconnect()` multiple times should not be a problem as the calls are queued to the main loop i.e. they will run sequentially and `terminate_ike()` will actually block until the IKE_SA is terminated, i.e. should there be a second call the SA will simply not found anymore, and since `disconnect_id` is reset early before calling `terminate_ike` that might also happen with the current code (unless the `disconnect()` calls happen in quick succession before any of them run).

lkundrak:
> I guess disconnect_id is not strictly necessary.

Correct, not necessary. I thought makes the a bit more idiomatic while potentially avoiding extra work. I'm not sure if you're asking me to remove it, but I don't mind removing it.

tobiasbrunner:
Applied to master. Thanks again!

PR #86, opened 2017-11-30
nm: allow disabling libnm-glib

lkundrak:
The distros are eventually going to drop it, allow omitting it.

### DIFF
--- a/src/frontends/gnome/Makefile.am
+++ b/src/frontends/gnome/Makefile.am
@@ -12,6 +12,7 @@ appdata_DATA = $(appdata_in_files:.xml.in=.xml)
 appdata_in_files = NetworkManager-strongswan.appdata.xml.in
 @INTLTOOL_XML_RULE@
 
+if WITH_LIBNM_GLIB
 # Install a file with full path to plugins for an old gnome-shell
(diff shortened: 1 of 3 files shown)


tobiasbrunner:
I guess this could already be applied? Or should we wait until #85 is finalized? When exactly are the two plugins used (libnm-vpn-plugin-strongswan vs. libnm-strongswan-properties)? Does that depend on the NM version? Or the GUI used?

lkundrak:
> I guess this could already be applied? Or should we wait until #85 is finalized?

No dependency on #85.

> When exactly are the two plugins used (libnm-vpn-plugin-strongswan vs. libnm-strongswan-properties)? Does that depend on the NM version? Or the GUI used?

Mostly on the GUI used.

The GNOME Control Center and the nm-connection-editor (both for some time already, no idea for how long) use only the new .so file for the editor widget. (There's no other known users of this -- KDE reimplements the functionality).

The GNOME Shell and the nm-applet also use libnm or libnm-glib for locating the authentication dialog. GNOME Shell is not yet ported, but that will hopefully happen for GNOME 3.28. The migration not being 100% complete is the reason this leaves the libnm-glib version enabled by default.

tobiasbrunner:
OK, great. Applied to master, thanks!

PR #87, opened 2017-12-03
ipsec_types: do not mask the configured mark value if it is one of th…

ebirger:
…e 'unique' values

Support for mark=%unique/%unique-dir is implemented by using designated
magic mark values.

Use of masks is orthogonal to the 'unique' feature, as it is useful to be
able to designate portions of the packet mark for other purposes, while
still using different marks for different connections.

When these magic values are masked, their magic meaning is lost.

Perform masking only on explicit mark values.

### DIFF
--- a/src/libstrongswan/ipsec/ipsec_types.c
+++ b/src/libstrongswan/ipsec/ipsec_types.c
@@ -104,7 +104,10 @@ bool mark_from_string(const char *value, mark_t *mark)
 	{
 		mark->mask = 0xffffffff;
 	}
-	/* apply the mask to ensure the value is in range */
-	mark->value &= mark->mask;
+	if (!MARK_IS_UNIQUE(mark->value))
+	{
+		/* apply the mask to ensure the value is in range */
+		mark->value &= mark->mask;
+	}
 	return TRUE;
 }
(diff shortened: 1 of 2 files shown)


tobiasbrunner:
Looks good, thanks! I pushed this, slightly modified, to the _gh87-mark-mask-unique_ branch. While I appreciate the macro hack to remove the braces it's really not that clear (i.e. why `_STRINGIFY s` and not `_STRINGIFY(s)`) and I'm not sure how portable it is. We don't gain that much with it anyway.

ebirger:
Thanks! indeed I wasn't too sure about the STRINGIFY hack. I wanted the test to be indifferent to the keyword value.

Looking at it a bit more, perhaps we should pass the keyword in the test string? something like:
{"%unique/0x0000ffff",	TRUE, { MARK_UNIQUE, 0x0000ffff }}

What do you think?

tobiasbrunner:
> Looking at it a bit more, perhaps we should pass the keyword in the test string? something like:
> {"%unique/0x0000ffff", TRUE, { MARK_UNIQUE, 0x0000ffff }}
> 
> What do you think?

Now that you mention it, that would definitely make sense. I guess doing both can't hurt either, I've updated the branch. Looks like nobody checked the test coverage of that function in a while (or ever).

ebirger:
Looks excellent. Thanks!

tobiasbrunner:
Applied to master. Thanks again.

PR #88, opened 2018-01-26
Update SA based on change in priority routing rule

kristrev:
On multihomed hosts, it is common to prioritize one network over the
other(s). For example, an LTE modem over a 2G modem, or a fixed
connection over a satellite connection. Typically, the system has no way
of knowing which interface to prioritize, as the actual network
interfaces looks the same (naming follows the same pattern, etc.). Also,
policies for selecting the preferred interface can be complex and use
multiple inputs (metadata, active checks, configuration options, etc.,
etc.)

Policy routing is commonly used on  multihomed hosts. The routes for the
different interfaces are placed in separate tables, and rules are used
to select which routing table(s) to check for a packet. This PR adds
support for specifying the priority of what I have called a preferred
rule. This rule will always point to the routing table of the preferred
network interface, and the rule is assumed to be updated by the tool
responsible for enforcing the network selection policy.

The feature is currently only available on Linux, and requires MOBIKE to
be enabled and "prefer_best_path" to be set to yes. When a rule with the
given priority is updated (i.e., a NEWRULE is received), a roam event is
triggered. Before the preferred route lookup is performed, a lookup for
the current priority table is done. Routes in this table is placed at
the front of the list that is used to select the preferred route.

The reason for not acting on DELRULE messages, is that I have assumed
that there can only be one preferred rule. Thus, if the rule is deleted,
then there is no preferred network and we can stay with the current one.

Signed-off-by: Kristian Evensen <kristian.evensen@gmail.com>

### DIFF
--- a/conf/plugins/kernel-netlink.opt
+++ b/conf/plugins/kernel-netlink.opt
@@ -116,3 +116,19 @@ charon.plugins.kernel-netlink.xfrm_acq_expires = 165
 	allocated by the kernel. The default value equals the total	retransmission
 	timeout for IKE messages, see IKEv2 RETRANSMISSION in
 	**strongswan.conf**(5).
+
+charon.plugins.kernel-netlink.preferred_rule_prio =
+	Priority of routing rule pointing to the routing table of the current
+	preferred network
+
+	On multihomed hosts, it is common to prioritize one network over the
+	other(s) (for example a modem with LTE over one with 2G). If policy
+	routing is used, this option can be set to the priority of the preferred
+	routing rule. I.e., the rule that points to the routing table of the
+	current preferred interface. When the rule is updated, the SAs will try
+	to be updated with MOBIKE using the preferred path. In order for this
+	option to have an effect, MOBIKE must be supported and enabled, and
+	charon.prefer_best_path set to yes.
+
+	Implementing the policy for which interface should be preferred and
+	updating the rule, is up to other applications.
--- a/src/libcharon/plugins/kernel_netlink/kernel_netlink_net.c
+++ b/src/libcharon/plugins/kernel_netlink/kernel_netlink_net.c
@@ -526,6 +526,12 @@ struct private_kernel_netlink_net_t {
 	 * MSS to set on installed routes
 	 */
 	uint32_t mss;
+
+	/**
+	 * Priority of rule pointing to the routing table for the current
+	 * preferred network
+	 */
+	uint32_t preferred_rule_prio;
 };
 
 /**
@@ -1452,6 +1458,37 @@ static void process_route(private_kernel_netlink_net_t *this, struct nlmsghdr *h
 }
 
 /**
+ * process RTM_NEWRULE from kernel
+ */
+static void process_rule(private_kernel_netlink_net_t *this,
+			 struct nlmsghdr *hdr)
+{
+	struct fib_rule_hdr* fib_hdr = NLMSG_DATA(hdr);
+	struct rtattr *rta = (struct rtattr*) (((uint8_t*) fib_hdr) +
+			     NLMSG_ALIGN(sizeof(struct fib_rule_hdr)));
+	size_t rtasize = NLMSG_PAYLOAD(hdr, sizeof(struct fib_rule_hdr));
+	uint32_t priority = 0;
+	bool prio_rule_found = false;
+
+	while (RTA_OK(rta, rtasize)) {
+		if (rta->rta_type == FRA_PRIORITY) {
+			priority = *(uint32_t*) RTA_DATA(rta);
+
+			if (priority == this->preferred_rule_prio) {
+				prio_rule_found = true;
+			}
+
+			break;
+		}
+		rta = RTA_NEXT(rta, rtasize);
+	}
+
+	if (prio_rule_found) {
+		fire_roam_event(this, FALSE);
+	}
+}
+
+/**
  * Receives events from kernel
  */
 static bool receive_events(private_kernel_netlink_net_t *this, int fd,
@@ -1507,6 +1544,9 @@ static bool receive_events(private_kernel_netlink_net_t *this, int fd,
 				{
 					process_route(this, hdr);
 				}
+				break;
+			case RTM_NEWRULE:
+				process_rule(this, hdr);
 				break;
 			default:
 				break;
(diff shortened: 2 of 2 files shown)


tobiasbrunner:
Hm, this seems kinda hackish :) I'd rather use all the rules and their priorities to order the routes (obviously a lot more work than this, though).

However, there is also a simple workaround as you can simply use the kernel's route lookup, with full support for routing rules, by setting _charon.plugins.kernel-netlink.fwmark = !mark_ (with an arbitrary mark not used for anything else on the system). That sets up the routing rule for table 220 (where strongSwan installs it's routes) to exclude traffic with that mark and we can therefore set the mark in our routing lookups to ignore our own routes and therefore can avoid dumping the routing table and having to do the lookup for the best route ourselves. If you set it up like that is the roam job for rule updates still necessary?

kristrev:
Hi!

Thanks for your detailed feedback, it is much appreciated.

My goal with this PR was to add support for controlling which network is used for IPsec (and change it on the fly) according to some policy, when multiple networks are available. The reason I chose to monitor a rule, was to have a solution that integrates well with what seems like the most common way of configuring policy routing. I will not make any claims to have seen all policy routing configurations in the world, but, for example, on my OpenWRT router the output of ip rule looks as follows (with two active internet connections):

```
0:      from all lookup local 
10000:  from 10.0.0.11 lookup 257 
10000:  from 10.0.0.66 lookup 10000 
20000:  from all to 10.0.0.11/24 lookup 257 
20000:  from all to 10.0.0.66/24 lookup 10000 
32766:  from all lookup main 
32767:  from all lookup default 
70001:  from all fwmark 0x3000000/0xff000000 lookup 220 
90020:  from all iif lo lookup 257 
90034:  from all iif lo lookup 10000
```

The first rules match on traffic to/from the local networks of my two internet-connections, the main rule contains the route to my LAN, and then there are two "default" rules for traffic not bound to any interface. If we assume that there is one IPsec tunnel configured, the tunnel is not bound to a specific interface, and that there are no firewalls blocking IPsec, then the tunnel will currently be established over either of the two interfaces.

If we add a rule with say priority 90001 and configure this as the preferred rule, then we can change which network the tunnel uses by replacing this rule. There are other ways to accomplish what I want, for example changing route metrics also seems to work. However, at least in my opinion, replacing a rule is the easiest and "cleanest" way. With for example the metric solution, the tool implementing the policy would need to be aware of which metric is set for different routes, and then update the metric for several routes.

Also, all rules/routes are still considered and prioritized as before. The "only" change made is that rules in the table pointed to by the preferred rule is placed first in the list (but sorted internally).

I have to admit I don't fully understand your proposed work-around. This is probably due to me not understanding something, but isn't table 220 only used to control which traffic will go into the tunnel? I don't quite get how I can do some tricks with this table + rule to control which network is used by Strongswan?

tobiasbrunner:
> but isn't table 220 only used to control which traffic will go into the tunnel?

Not really, that is determined by the IPsec policies. Routes are only needed if virtual IPs are used (to force a specific source IP so packets match the policies), or if there are no routes otherwise that would match the tunneled traffic (as the kernel won't forward the traffic otherwise). So they might not actually be needed at all. And the manual route lookup in the _kernel-netlink_ plugin is only done so we can avoid the routes we install ourselves when determining the source IP we use for IKE packets or when deciding whether we need to trigger MOBIKE updates because the current path can't be used anymore etc.

In certain situations we don't need to do this lookup manually and can simply let the kernel decide which route/source IP we should use. One is the described workaround where we set a negative mark match on the routing rule for table 220, which allows us to easily avoid the table with our own routes when doing lookups (by setting that mark in the lookup request the kernel will exclude our own table which will only be used for packets that don't have that mark set). And there is actually an even easier way to use this more efficient and powerful lookup, disable route installation completely (i.e. set _charon.install_routes = no_). Then strongSwan won't install any routes in table 220 and always use the kernel's route lookup.

I suppose, if the only thing changing is the priority of the routing rule then we'd need the roam event trigger based on such changes to properly trigger a MOBIKE update (depends on the events actually triggered by the kernel during such a switch).

kristrev:
Thanks for another detailed explanation!

> And there is actually an even easier way to use this more efficient and powerful lookup, disable route installation completely (i.e. set charon.install_routes = no). Then strongSwan won't install any routes in table 220 and always use the kernel's route lookup.

This explains why I sometimes saw that handover based on priority sometimes seemed to work. If I disconnected and then connected the link with the highest priority (according to my policy), the tunnel would most of the time switch back. This is of course because the order of the routes returned by kernel happened to match.

> I suppose, if the only thing changing is the priority of the routing rule then we'd need the roam event trigger based on such changes to properly trigger a MOBIKE update (depends on the events actually triggered by the kernel during such a switch).

Yes, at least in my case, the only change is that the routing rule is replaced. Both networks are still up and the routing tables, etc. are not modified, so there is only DELRULE/NEWRULE messages. How should we progress with this PR (besides fixing Travis-complaints), if a feature like what I propose is acceptable? In my initial solution I triggered a roam event on every NEWRULE, but that is a bit excessive, and the "filter" that made most sense to me was priority.

tobiasbrunner:
> In my initial solution I triggered a roam event on every NEWRULE, but that is a bit excessive, and the "filter" that made most sense to me was priority.

Are updates to these rules that common that this would be a problem? (Kernel events triggered within ROAM_DELAY do not trigger more than one roam event, defaults to 100 ms.)

I've pushed an optional roam trigger based on rule updates to the _2518-mobike-nat_ branch (works only with the kernel-based route lookup), together with other commits that try to address MOBIKE the issues from your [wiki ticket](https://wiki.strongswan.org/issues/2518) (see there for details).

kristrev:
> Are updates to these rules that common that this would be a problem? (Kernel events triggered within ROAM_DELAY do not trigger more than one roam event, defaults to 100 ms.)

No, not really, at least not on my system. While the rules are updated fairly frequently, it is not hundreds of times per seconds. So unconditionally triggering an event on rule-updates should be fine.

I tested the code you added in the 2518-branch (will test the MOBIKE-part properly later today) and it looks fine. However, there is still the issue of configuring which network to prioritize when multiple networks are active. With your change, we depend on the ordering of the routes returned by the kernel. I have to admit that I don't exactly know how the kernel sorts routes before/while sending them to userspace, but when checking this earlier at least, it did not for example depend on the order of the rules.

The only way I found to solve the ordering problem, was to instruct Strongswan to prioritize routes in one table over the other. And the only way I found to dynamically get which table to prioritize, was to have Strongswan check one specific rule.

tobiasbrunner:
> However, there is still the issue of configuring which network to prioritize when multiple networks are active. With your change, we depend on the ordering of the routes returned by the kernel.

As mentioned above, in the commit message and the documentation of the new option it only works with the kernel-based route lookup. So you either have to disable route installation by strongSwan or let strongSwan set an inverted mark its own routing rule (see above for details).

kristrev:
> As mentioned above, in the commit message and the documentation of the new option it only works with the kernel-based route lookup. So you either have to disable route installation by strongSwan or let strongSwan set an inverted mark its own routing rule (see above for details).

Yes, I saw that. However, I have assumed that the policy is implemented outside of the routing demon, so that order of routes/rules does not necessarily reflect the priority of interfaces. Though, if I make sure that the priority rule is always placed earlier than the other "default"-rules, I guess I should be fine.

tobiasbrunner:
> However, I have assumed that the policy is implemented outside of the routing demon, so that order of routes/rules does not necessarily reflect the priority of interfaces.

What do you mean?

> Though, if I make sure that the priority rule is always placed earlier than the other "default"-rules, I guess I should be fine.

Wouldn't that be necessary anyway to control over which interface traffic is routed?

kristrev:
> What do you mean?

For example, I am using my OpenWRT router in a car that is moving. That means that modems can change mode, for example from LTE to 2G, without disconnecting. If the modem does not disconnect, looses adress, etc., then the routing and so on remains constant.

However, you are right that my priority rule would anyway have to be placed before the other default rules to have the desired effect (this is in fact how it works on my router). I just didn't tjink about it. Thus, my patch is too complicated and your change should be sufficient.

I will do some tests just in case and let you know.

kristrev:
I was finally able to test the rule-change you made in the branch connected to my ticket in the wiki. It seems to do the job, when changing my priority interface then the tunnel follows.

tobiasbrunner:
Thanks for testing. I'm closing this PR and will merge the new trigger together with the MOBIKE fixes, once we have these finalized.

PR #89, opened 2018-02-12
Fix up one typo mistake

liuqun:
- Fix up a `pki` command in section [Generating a Host or User End Entity Certificate](https://github.com/strongswan/strongswan#generating-a-host-or-user-end-entity-certificate) of the top level README
- `/etc/swanctl/priv` --> `/etc/swanctl/private`


tobiasbrunner:
Applied to master, thanks!

PR #90, opened 2018-02-14
increase recv buffer side on the ha_socket

lofvir:
otherwise some message are truncated and thus incomplete.


tobiasbrunner:
Already in the _ha-buflen_ branch.

PR #91, opened 2018-03-06
perl/Vici-Session: fix recv code and improve error handling

lofvir:
the recv code now considers the situation of incomplete reads
(e.g. due to some interrupts).

error handling is improved to avoid termination of perl programs that
use the vici perl module.

### DIFF
--- a/src/libcharon/plugins/vici/perl/Vici-Session/lib/Vici/Message.pm
+++ b/src/libcharon/plugins/vici/perl/Vici-Session/lib/Vici/Message.pm
@@ -107,13 +107,15 @@ sub parse {
                  }
                 else
                 {
-                    die "message parsing error: ", $type, "\n"
+                    $hash->{success} = 'no';
+                    $hash->{errmsg}  = 'message parsing error: '. $type;
                 }
             }
         }
         else
         {
-            die "message parsing error: ", $type, "\n"
+            $hash->{success} = 'no';
+            $hash->{errmsg}  = 'message parsing error: '. $type;
         }
     }
     return $data;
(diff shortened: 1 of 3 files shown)


lofvir:
I suspect that the python binding and maybe the ruby binding is also affected by the problems reading the STREAM based unix socket.

tobiasbrunner:
Thanks. I pushed a fix for the `recv()` handling to the _gh91-vici-perl_ branch. This is basically how we already do this in the other two bindings.

I'm no Perl guy at all, but exception handling seems easily possible using `eval {}` and `$@` (see [die](http://perldoc.perl.org/functions/die.html)).

d12fk:
Thanks for accepting the recv patch.

However, I feel the urge to contribute my 2¢ on topic of die()ing in Perl modules. =) Here we go:
While it is possible to use eval here, I would consider it a shortcoming of a module that I have to. Especially because there is an established way to signal failure. The die/eval combination is Perl's way of doing "exception handling", but it is rather a poor man's solution in my opinion. I think modules should not force this construct on their users, when there's better ways of communicating failure. While die()ing is elegant for the author of the the module, it's not so elegant for its users.

Also take into consideration that use of eval is considered harmful in interpreted languages in general (not so much for catching errors in Perl, though). So, there could be situations where use of eval is restricted by company coding guidelines.

tobiasbrunner:
It just looks nicer using exceptions and we can avoid returning magic values (or multiple values), there is also no need to check return values of every called method. Using other functions from the [Carp](http://perldoc.perl.org/Carp.html) module might be even more useful, or using actual exception objects. Searching around some also showed that there are modules that provide easy to use `try {} catch {}` semantics that are well-known from other languages (see e.g. [Error.pm](https://www.perl.com/pub/2002/11/14/exception.html/) or [Try::Tiny](http://search.cpan.org/~ether/Try-Tiny-0.30/lib/Try/Tiny.pm)). And Perl 6 actually [has exception handling built-in](https://docs.perl6.org/language/exceptions).

lofvir:
@tobiasbrunner is there any plan to merge gh91-vici-perl into master?

tobiasbrunner:
Yes, will definitely be included in the next release.

PR #92, opened 2018-03-19
libipsec: fix ip_packet_create_from_data version field in IPv6 header

matfabia:
Version field of IPv6 header in ip_packet_create_from_data should be set as "ip6_flow = htonl(6 << 28)" instead of "ip6_flow = htonl(6)"


tobiasbrunner:
Applied to master. Thanks!

PR #93, opened 2018-03-20
child-delete: avoid duplicate calls of child_updown

lofvir:
if a CHILD_SA is rekeyed after the beginning of a ike_reauth()
there is a chance that the rekeyed CHILD_SA ends up in state DELETING.
This is due to charon.delete_rekeyed_delay being set to a high
value (like 100) in combination with make_before_break = yes.

In the course of completing ike_reauth() the old IKE_SA
gets destroyed via a call to ike_updown(). This implies calls to
child_updown() but in this situation the fact that the CHILD_SA
is already rekeyed cannot be identified anymore because the state
is already DELETING.

### DIFF
--- a/src/libcharon/bus/bus.c
+++ b/src/libcharon/bus/bus.c
@@ -827,7 +827,7 @@ METHOD(bus_t, ike_updown, void,
 		enumerator = ike_sa->create_child_sa_enumerator(ike_sa);
 		while (enumerator->enumerate(enumerator, (void**)&child_sa))
 		{
-			if (child_sa->get_state(child_sa) != CHILD_REKEYED)
+			if (child_sa->get_state(child_sa) != CHILD_REKEYED && !child_sa->get_already_rekeyed(child_sa))
 			{
 				child_updown(this, child_sa, FALSE);
 			}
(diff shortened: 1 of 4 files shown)


tobiasbrunner:
That's an ugly hack I won't apply. Please find the root cause for this and fix that.

lofvir:
The root cause is explained in the commit description. This can happen because child rekeyeing can happen while ike_reauth is ongoing. Therefore such situations can occur if mbb is configured together with delete_rekeyed_delay. An alternative could be to not change the state to DELETING when sending out the child delete message but later on when the delayed job executes.

tobiasbrunner:
Both your descriptions are inaccurate. The problem is not that the rekeying happens during a reauthentication, or setting the state to `CHILD_DELETING` when deleting the SA.

The problem is that _after_ deleting the rekeyed CHILD_SA and with _charon.delete_rekeyed_delay_ set to a higher value the CHILD_SA may remain in state `CHILD_DELETING` until the IKE_SA is deleted (no matter whether it happens during a reauthentication or not). So instead of adding a new flag with getter and setter, we could e.g. add a new state (e.g. `CHILD_DELETED`) for such SAs (SAs that were deleted via IKE but have not yet been destroyed), or e.g. check the state of the outbound SA to see whether we should ignore SAs in state `CHILD_DELETING` when triggering `child_updown()` (as e.g. done in child_rekey). I think the former is a bit cleaner so I implemented that in the _gh93-child-deleted_ branch.

lofvir:
Introducing a new state provides the same functionality in a clean way. Thanks, will test this soon.

PR #94, opened 2018-03-26
wincapi: Added plugin to access Windows certificate store private keys

rdlr:
Wincapi is a PRIVKEY plugin for libstrongswan that uses the Windows CryptoAPI to provide access to private keys in the Windows local machine certificate store.

### DIFF
--- a/src/libstrongswan/plugins/wincapi/wincapi_private_key.c
+++ b/src/libstrongswan/plugins/wincapi/wincapi_private_key.c
@@ -1,0 +1,596 @@
+/*
+ * Copyright (C) 2018 Robert de la Rey, Francois ten Krooden
+ * Copyright (C) 2018 Nanoteq (Pty) Ltd.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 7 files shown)


src/libstrongswan/plugins/wincapi/wincapi_private_key.h
```diff
@@ -0,0 +1,64 @@
...
+	 */
+	private_key_t key;
+};
+
+
+/**
+ * Get a handle to a private key stored in the Windows Certificate Store.
+ *
+ * The function takes BUILD_SUBJECT and optionally a BUILD_PUBLIC_KEY
+ * to select a specific key loaded in the store.
```
tobiasbrunner: That's not what's implemented currently.

src/libstrongswan/plugins/wincapi/wincapi_private_key.c
```diff
@@ -0,0 +1,596 @@
...
+METHOD(private_key_t, sign, bool,
+		private_wincapi_private_key_t *this, signature_scheme_t scheme, void *params,
+		chunk_t data, chunk_t *signature)
+{
+	HCRYPTHASH hash_obj;
+	unsigned int alg_id;
+	BYTE *buf;
+	DWORD buf_len;
+
+	switch (scheme)
```
tobiasbrunner: You might want to make sure the key is actually an RSA key.

src/libstrongswan/plugins/wincapi/wincapi_private_key.c
```diff
@@ -0,0 +1,596 @@
...
+	chunk_t data = chunk_empty;
+
+	while (TRUE)
+	{
+		builder_part_t build_part = va_arg(args, builder_part_t);
+		switch (build_part)
+		{
+			case BUILD_BLOB_PEM:
+				data = va_arg(args, chunk_t);
+				subject = identification_create_from_string(data.ptr);
```
tobiasbrunner: Hm, passing an identity in `BUILD_BLOB_PEM` is definitely not ideal. You might want to use `BUILD_PKCS11_KEYID` (and optionally the other `BUILD_PKCS11_*` parameters), which is usually used for keys that are not directly accessible. Then e.g. pass and use the SHA-1 hash of the public key for the lookup via `CERT_FIND_SHA1_HASH`.

src/libstrongswan/plugins/wincapi/wincapi_private_key.c
```diff
@@ -0,0 +1,596 @@
...
+
+
+METHOD(private_key_t, decrypt, bool,
+		private_wincapi_private_key_t *this, encryption_scheme_t scheme,
+		chunk_t crypto, chunk_t *plain)
+{
+	BYTE *buf;
+	DWORD buf_len;
+
+	if (scheme != ENCRYPT_RSA_PKCS1)
```
tobiasbrunner: Similar to `sign()`, the key type should probably be checked.

rdlr:
We used CryptoAPI functions because our implementation executes Charon as a local machine service that is initialised before user login, on system start. According to the MSDN documentation for [NCryptOpenKey](https://msdn.microsoft.com/en-us/library/windows/desktop/aa376284(v=vs.85).aspx):

> A service must not call this function from its StartService Function. If a service calls this function from its StartService function, a deadlock can occur, and the service may stop responding.

Thank you for the recommendations, I will review and update our code.

tobiasbrunner:
> We used CryptoAPI functions because our implementation executes Charon as a local machine service that is initialised before user login, on system start.

Do you need to call it in StartService? (How exactly would that even happen exactly with how the daemon is started and configs are loaded?)

PR #95, opened 2018-03-27
unify child_updown calls when having duplicate QMs

lofvir:
If a Quick mode is initated for a CHILD_SA that is already installed
we can identify this situation and rekey the already installed CHILD_SA.

Otherwise we end up with several CHILD_SAs in state INSTALLED which
means multiple calls of child_updown are done. Unfortunatly
the deduplication code does not call child_updown() (up and down were not
even). Another problem occurs if the newer CHILD_SA has a shorter
lifetime due to rekey fuzz. Then a uneven number of calls to child_updown can be
observed on the responder side.

### DIFF
--- a/src/libcharon/sa/ikev1/tasks/quick_mode.c
+++ b/src/libcharon/sa/ikev1/tasks/quick_mode.c
@@ -1005,13 +1005,22 @@ static bool has_notify_errors(private_quick_mode_t *this, message_t *message)
 /**
  * Check if this is a rekey for an existing CHILD_SA, reuse reqid if so
  */
-static void check_for_rekeyed_child(private_quick_mode_t *this)
+static void check_for_rekeyed_child(private_quick_mode_t *this, bool responder)
 {
 	enumerator_t *enumerator, *policies;
 	traffic_selector_t *local, *remote;
+	traffic_selector_t *my_ts, *other_ts;
 	child_sa_t *child_sa;
 	proposal_t *proposal;
 	char *name;
+
+	if (responder) {
+		my_ts = this->tsr;
+		other_ts = this->tsi;
+	} else {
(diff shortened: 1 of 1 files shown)


lofvir:
Hi @tobiasbrunner,
could you please consider this PR to be merged?
without the change the number of updown calls can become out of sync. This happens in
the unlikely but possible situation when initiating a tunnel twice.

tobiasbrunner:
Thanks, pushed to the _gh-95-qm-dedup_ branch with some minor changes.

> Unfortunatly
> the deduplication code does not call child_updown() (up and down were not
> even).

That's on purpose, `child_rekey()` is called for rekeyed CHILD_SAs instead.

> Another problem occurs if the newer CHILD_SA has a shorter
> lifetime due to rekey fuzz.
> Then a uneven number of calls to child_updown can be observed on the responder side.

How so? And the responder already called `check_for_rekeyed_child()`, so how does this patch change the behavior?

lofvir:
>>Another problem occurs if the newer CHILD_SA has a shorter
>>lifetime due to rekey fuzz.
>>Then a uneven number of calls to child_updown can be observed on the responder side.

>How so? And the responder already called check_for_rekeyed_child(), so how does this patch change ?>the behavior?

After rethinking this, i cannot remember the exact scenario. The important part of this changeset is that the number of child_updown() calls gets symmetric on the initiator side.

If left initiates childs c1 and c2, previously both ended up in state INSTALLED which implies 2 calls to child_updown. If now rekeying comes into play the duplicate INSTALLED SA's are deduplicated and only one INSTALLED SA is left. When tearing down the tunnel only one call to child_updown occurs. The second one got eaten up.

tobiasbrunner:
OK, I removed that part from the commit message and pushed it to master.

PR #96, opened 2018-03-27
Use unused return value of C_GetMechanismList

robinleander:
The pkcs11 plugin does not use the return value of the second `C_GetMechanismList` call, and thus won't detect if it failed.


tobiasbrunner:
Applied to master, thanks!

PR #99, opened 2018-04-17
Add RFC6023 support

llogar:
Childless IKEv2 SA can now be established by issuing (note missing --child parameter)

`$ swanctl --initiate -ike [name]`

Also, a connection 'name' defined in swanctl.conf must not contain any children, 
i.e. children {...} should be empty or absent.

RFC6023 support can be disabled by setting charon.childless_ikev2 = no in strongswan.conf

### DIFF
--- a/conf/options/charon.opt
+++ b/conf/options/charon.opt
@@ -36,6 +36,10 @@ charon.cache_crls = no
 	Certification Authority (CA) to **/etc/ipsec.d/crls** (stroke) or
 	**/etc/swanctl/x509crl** (vici), respectively.
 
+charon.childless_ikev2 = yes
+	Whether we support RFC6023 - Childless Initiation of the Internet Key
+	Exchange Version 2 (IKEv2) Security Association (SA)
+
 charon.cisco_unity = no
 	Send Cisco Unity vendor ID payload (IKEv1 only).
 
--- a/src/libcharon/config/peer_cfg.c
+++ b/src/libcharon/config/peer_cfg.c
@@ -699,12 +699,6 @@ METHOD(peer_cfg_t, equals, bool,
 		);
 }
 
-METHOD(peer_cfg_t, has_children, bool,
-	private_peer_cfg_t *this)
-{
-	return this->child_cfgs->get_count(this->child_cfgs) ? TRUE : FALSE;
-}
-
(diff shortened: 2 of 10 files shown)


tobiasbrunner:
Thanks for these patches. I have not reviewed the changes thoroughly yet, but some of them would definitely be required for any such solution (e.g. condition/extension flags, some code in the tasks).

> Also, a connection 'name' defined in swanctl.conf must not contain any children,
> i.e. children {...} should be empty or absent.

That seems like quite a limitation, what's the point of creating an IKE_SA if there are no children at all that could later be established? Or would you add them later and rely on _charon.reuse_ikesa_?

I did some brainstorming regarding a possible configuration option for childless IKE_SAs a while ago and the following was a possibility I came up with as extension on ike_cfg_t, here as configuration option for a connection in swanctl.conf:

	childless = accept | never | prefer | force

	accept (default): Create CHILD_SA with IKE_AUTH exchange but
		accept childless IKE_SAs as responder
	never: Create CHILD_SA with IKE_AUTH exchange and don't allow
		childless IKE_SAs as responder (i.e. don't send the
		notify in IKE_SA_INIT and don't allow initiators to omit required
		payloads in IKE_AUTH), fail initiation if only the IKE_SA was
		 initiated (e.g. via swanctl --initiate --ike)
	prefer: Try to initiate a childless IKE_SA and create the first
		CHILD_SA, if any, with a separate CREATE_CHILD_SA
		exchange.  Don't fail if the peer does not support it,
		just create the CHILD_SA with the IKE_AUTH exchange.
		As responder this is the same as accept.
		# Probably only required to migrate to a childless initiation
	force: Initiate a childless IKE_SA and create the first
		CHILD_SA, if any, with a separate CREATE_CHILD_SA
		exchange.  Fail if the responder does not support
		childless IKE_SAs. Fail as responder if the initiator does not
		create a childless IKE_SA.

With the _never_ option a global setting to disable support would not be required. This is not completely thought through and maybe is total overkill. The idea was to provide an option for the use case of using a separate DH exchange for the first CHILD_SA (i.e. set it to _force_ on the client). What do you think about this?

tobiasbrunner:
I pushed changes to implement this feature to the _gh99-childless_ branch. Feedback welcome.

PR #100, opened 2018-04-25
Disable roaming (MOBIKE) for IKEv1

cedi:
Hey guys,

I've encountered a problem where charon tries to roam a IKEv1 connection to another WAN Uplink.

Setup:
* I have two WAN Uplinks:
 1. static IP
 2. DHCP assigned IP
*I configure an IKEv1 connection with the config value for `left` with the IP address of the DHCP assigned interface. 
* charon correctly established the connection with the DHCP-assigned IP as source
* I force a DHCP-lease renewal for the DHCP-WAN interface
* charon tries to roam the connection to the second WAN Interface with the static IP

This leads to problems in our setups, since WAN1 and WAN2 are in different vlans.
Also RFC4555 (MOBIKE) is only specified for IKEv2, so i think this is a bug in charons IKEv1 implementation when it tries to roam IKEv1 too.

The roam_job is scheduled from the `kernel_iph_net.c` in case ` static void update_addrs(private_kernel_iph_net_t*, iface_t*, IP_ADAPTER_ADDRESSES*, bool)` is called (eg. when i force a DHCP-lease renewal on the WAN-Interface.

To prevent this, i submit this patch, which skips IKEv1 IKE-SAs in the `roam_job->execute()` method.

Kind Regards,
Cedric Kienzler
Software Engineer, Sophos

For further Questions, feel free to contact me at any time using cedric.kienzler@sophos.com

### DIFF
--- a/src/libcharon/processing/jobs/roam_job.c
+++ b/src/libcharon/processing/jobs/roam_job.c
@@ -69,13 +69,6 @@ METHOD(job_t, execute, job_requeue_t,
 		ike_sa = charon->ike_sa_manager->checkout(charon->ike_sa_manager, id);
 		if (ike_sa)
 		{
-			if (ike_sa->get_version(ike_sa) == IKEV1)
-			{
-				DBG2(DBG_IKE, "RFC4555 is only specified for IKEv2.");
-				charon->ike_sa_manager->checkin(charon->ike_sa_manager, ike_sa);
-				continue;
-			}
-
 			if (ike_sa->roam(ike_sa, this->address) == DESTROY_ME)
 			{
 				charon->ike_sa_manager->checkin_and_destroy(
--- a/src/libcharon/processing/jobs/roam_job.c
+++ b/src/libcharon/processing/jobs/roam_job.c
@@ -69,6 +69,13 @@ METHOD(job_t, execute, job_requeue_t,
 		ike_sa = charon->ike_sa_manager->checkout(charon->ike_sa_manager, id);
 		if (ike_sa)
 		{
+			if (ike_sa->get_version(ike_sa) == IKEV1)
+			{
+				DBG2(DBG_IKE, "RFC4555 is only specified for IKEv2.");
+				charon->ike_sa_manager->checkin(charon->ike_sa_manager, ike_sa);
+				continue;
+			}
+
 			if (ike_sa->roam(ike_sa, this->address) == DESTROY_ME)
 			{
 				charon->ike_sa_manager->checkin_and_destroy(
--- a/src/libcharon/sa/ike_sa.c
+++ b/src/libcharon/sa/ike_sa.c
@@ -2583,9 +2583,15 @@ METHOD(ike_sa_t, roam, status_t,
 		return SUCCESS;
 	}
 
+	if (!this->version == IKEV1)
+	{ 	/* ignore roam events for IKEv1 where we don't have MOBIKE and would
+		 * have to reestablish from scratch */
+		return SUCCESS;
+	}
+
 	/* ignore roam events if MOBIKE is not supported/enabled and the local
 	 * address is statically configured */
-	if (this->version == IKEV2 && !supports_extension(this, EXT_MOBIKE) &&
+	if (!supports_extension(this, EXT_MOBIKE) &&
 		ike_cfg_has_address(this->ike_cfg, this->my_host, TRUE))
 	{
 		DBG2(DBG_IKE, "keeping statically configured path %H - %H",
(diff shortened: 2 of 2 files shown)


cedi:
Please note: i also suspect that there might be another issue with MOBIKE:
We already have set the `mobike` config flag to `no` globally. But it tries to do mobike anyway.

As far as i have seen, the `mobike` parameter is only respected in the `task_manager_v2`, so it would also be an good idea to check the `mobike` config flag in the case of `kernel_iph_net`

tobiasbrunner:
> Also RFC4555 (MOBIKE) is only specified for IKEv2, so i think this is a bug in charons IKEv1 implementation when it tries to roam IKEv1 too.

MOBIKE is not used for IKEv1 (`supports_extension(this, EXT_MOBIKE)` will return `FALSE`), instead the IKE_SA is reauthenticated when the connection changes to a different IP (see the code in `ike_sa_t::roam()`). However, that is probably problematic for IKEv1, where this only affects the IKE_SA, so we might want to consider reestablishing the IKE and all CHILD_SAs instead. There is also a check that makes sure that the local IP is not changed if it is statically configured. But it's specific to IKEv2 for some reason. Anyway, we could also just return `SUCCESS` from `roam()` for IKEv1 SAs (I wouldn't want to do that in the job, which should be kept fairly simple), I pushed a change that does that to the _gh100-ikev1-roam_ branch.

> The roam_job is scheduled from the kernel_iph_net.c in case static void update_addrs(private_kernel_iph_net_t*, iface_t*, IP_ADAPTER_ADDRESSES*, bool) is called (eg. when i force a DHCP-lease renewal on the WAN-Interface.

You are running on Windows?

> We already have set the mobike config flag to no globally. But it tries to do mobike anyway.

There is no global mobike flag. What are you referring to?

cedi:
Hey @tobiasbrunner,

> There is no global mobike flag. What are you referring to?
Sorry for confusion. I'm talking about 

```
conn %default
    mobike = no
```

in `ipsec.conf`


> You are running on Windows?

Nope, Linux (3.14.22)

> Anyway, we could also just return SUCCESS from roam() for IKEv1 SAs (I wouldn't want to do that in the job, which should be kept fairly simple)

Hehe, we had the same Idea. I originally have done that in `roam()`, but my workmate (Afschin) thought it would be better to move that out of `roam()` to the place where `roam()` is called.

> I pushed a change that does that to the gh100-ikev1-roam branch.

I will look into your code, thanks

> There is also a check that makes sure that the local IP is not changed if it is statically configured. But it's specific to IKEv2 for some reason

Wouldn't it be a good thing to implement this check for IKEv1 too? If you have no concerns about this, i would volunteer to provide that patch :smile:

tobiasbrunner:
> Sorry for confusion. I'm talking about
> 
> conn %default
>     mobike = no
> 
> in ipsec.conf

OK, this should definitely disable MOBIKE (support for it is not announce anymore, and therefore should not be negotiated and also not be in the list of supported extensions of the IKE_SA anymore).

>     You are running on Windows?
> 
> Nope, Linux (3.14.22)

So why would you think `kernel_iph_net.c` is involved in any of this? On Linux you'll be using the _kernel-netlink_ plugin as interface to the kernel i.e. `kernel_netlink_net.c` for roaming events. The _kernel-iph_ plugin provides the _kernel_net_t_ implementation on Windows.

>     Anyway, we could also just return SUCCESS from roam() for IKEv1 SAs (I wouldn't want to do that in the job, which should be kept fairly simple)
> 
> Hehe, we had the same Idea. I originally have done that in roam(), but my workmate (Afschin) thought it would be better to move that out of roam() to the place where roam() is called.

I don't think so, because then we'd have the decisions when to roam and when not in different places making this a bit confusing.

>     There is also a check that makes sure that the local IP is not changed if it is statically configured. But it's specific to IKEv2 for some reason
> 
> Wouldn't it be a good thing to implement this check for IKEv1 too? If you have no concerns about this, i would volunteer to provide that patch smile

Not really necessary anymore if we don't do any roaming for IKEv1 (see my patch, which also touches the mentioned check).

cedi:
> So why would you think kernel_iph_net.c is involved in any of this? On Linux you'll be using the kernel-netlink plugin as interface to the kernel i.e. kernel_netlink_net.c for roaming events. The kernel-iph plugin provides the kernel_net_t implementation on Windows.

Sorry, that was my fault. You are right, i just copied the wrong file-name from my shell ;-)

> (see my patch, which also touches the mentioned check)

yea. Didn't had the time to look into your branch till now. Awesome!
I initially thought that check might be a bit more complex / involve a bit more code changes.

How to proceed from here on? Should i close this PR and you merge your fix into mainline?

tobiasbrunner:
> How to proceed from here on? Should i close this PR and you merge your fix into mainline?

Sure, sounds good. I hope nobody has any objections regarding this (didn't really work before anyway, as far as I can tell).

PR #101, opened 2018-06-05
starter: dont lose signals 

Links2004:
The starter will lose update / reload commands when there is a second signal coming in when the previous is still processed.

this can happen more easily with big configurations.

### DIFF
--- a/src/starter/starter.c
+++ b/src/starter/starter.c
@@ -650,6 +650,7 @@ int main (int argc, char **argv)
 		 */
 		if (_action_ & FLAG_ACTION_RELOAD)
 		{
+			_action_ &= ~FLAG_ACTION_RELOAD;
 			if (starter_charon_pid())
 			{
 				for (conn = cfg->conn_first; conn; conn = conn->next)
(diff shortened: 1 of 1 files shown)


src/starter/starter.c
```diff
@@ -855,6 +855,14 @@ int main (int argc, char **argv)
 			}
 		}
 
+ 		/*
+		 * Run again if there's pending action
+		 */
+		if (_action_)
```
tobiasbrunner: I don't think this is required, the `pselect()` call below is skipped if `_action_` is non-zero.

tobiasbrunner:
I guess unsetting the flags earlier makes sense, however, it doesn't prevent races completely. Also note that starter/stroke are deprecated. Use [vici](https://wiki.strongswan.org/projects/strongswan/wiki/Vici)/[swanctl](https://wiki.strongswan.org/projects/strongswan/wiki/swanctl) instead.

src/starter/starter.c
Links2004: true, will change

Links2004:
yes, it is not perfect save against races but better then before.

tobiasbrunner:
Thanks, applied to master.

PR #102, opened 2018-06-05
starter: check if modprobe exists and load all modules

Links2004:
check for existence of modprobe and load all Kernel Modules for strongswan

### DIFF
--- a/src/starter/netkey.c
+++ b/src/starter/netkey.c
@@ -42,13 +42,24 @@ bool starter_netkey_init(void)
 	}
 
 	/* make sure that all required IPsec modules are loaded */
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
I don't think there is a need to do any calls to `modprobe`, these modules are loaded automatically when necessary on modern kernels. I'd actually rather remove the complete function.

tobiasbrunner:
I see. Yes, that's quite old. Besides not being necessary on newer kernels, it doesn't really make sense to always load _all_ of them, if they are eventually not required. If missing modules are a problem, please load them outside of strongSwan.

PR #103, opened 2018-06-05
vici perl: improve message parsing performance

lofvir:
During a test with ~12000 established SAs it was noted that vici
related operations hung.
The operations took over 16 minutes to finish. The time was spent in
the vici message parser, which was assigning the message over and over
again, to get rid of the already parsed portions.

First fixed by cutting the consumed parts off without copying the message.
Runtime for ~12000 SAs is now around 20 seconds.

Further optimization brought the runtime down to roughly 1-2 seconds
by using an fd to read through the message variable.

### DIFF
--- a/src/libcharon/plugins/vici/perl/Vici-Session/lib/Vici/Message.pm
+++ b/src/libcharon/plugins/vici/perl/Vici-Session/lib/Vici/Message.pm
@@ -29,7 +29,9 @@ sub from_data {
     my $data = shift;
     my %hash = ();
 
-    parse($data, \%hash);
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks for the patch.  I guess, technically, there should be some error checking when reading the data, in case the message is too short or otherwise invalid, but we might consider vici (and UNIX sockets) trustworthy enough to ignore that. I pushed your patch and a possible patch that adds some error handling to the `gh103-vici-perl-performance` branch. Let me know if that works for you (or if you think it's unnecessary to add that).

PR #104, opened 2018-06-07
add config options to set NOPMTUDISC and NOECN via xfrm

Links2004:
### xfrm_nopmtudisc = (no | yes)

XFRM_STATE_NOPMTUDISC is used to disable PMTU (Path Maximum Transmission Unit)
discovery, in cases where the discovery is not supported by all Devices on the
Network Path or a fixed MTU should be used.

### xfrm_noecn = (no | yes)

XFRM_STATE_NOECN is used to disable the use of ECN (Explicit Congestion Notification).
This is usefull when the remote endpoint or the underlying network / involved routers
not support this.
https://tools.ietf.org/html/rfc3168
https://en.wikipedia.org/wiki/Explicit_Congestion_Notification

### Note:

both flags are **no** by default, there will no behavior change for existing configs.
this implemntation will only work with linux / netlink.


tobiasbrunner:
Thanks for the patch. There are some problems with it.

First, there are some code style issues (specifically in the _kernel-netlink_ plugin), please see [our wiki](https://wiki.strongswan.org/projects/strongswan/wiki/ProgrammingStyle).

Second, I wouldn't make these options XFRM specific (the flags can actually also be set via the PF_KEYv2 interface on Linux, although, that interface is deprecated, so it's OK not to set them there) and generally control behavior that's defined in RFC 4301, so other kernel interfaces could support them too.
Also, `XFRM_STATE_NOPMTUDISC`, in accordance with that RFC, only has an effect on IPv4 tunnel mode SAs, and it simply controls whether the DF bit is copied to the outer header, or gets cleared (there is no option to force setting the bit, which the RFC mentions), so that should be documented (and maybe the option's name could reflect that too).

Third, an additional and similar flag that might be interesting is `XFRM_SA_XFLAG_DONT_ENCAP_DSCP`, which controls whether DSCP information is copied to the outer header in tunnel mode (like ECN for both families, although ECN is applied in both directions). Maybe we could name the options _copy_df_, _copy_ecn_, and _copy_dscp_ all with a default value of _yes_.

Fourth, I won't apply the changes to starter/stroke, even if you provide patches (no new features are added to that interface, we want people moving to the vici/swanctl backend). So you'd have to maintain those changes in your own fork if you need them.

Links2004:
makes `nocopy_df`, `nocopy_ecn` and `nocopy_dscp` more sense?
since the default value is *no*, otherwise the new flag will change existing configurations,
or is there a way to set default to *yes*?

how does `ipv4_clear_pmtudisc` sound for `XFRM_STATE_NOPMTUDISC`?

tobiasbrunner:
> makes nocopy_df, nocopy_ecn and nocopy_dscp more sense?

Doesn't look that readable to me, maybe _no_copy_df_, but _copy_df_ seems simpler to me.

> since the default value is no, otherwise the new flag will change existing configurations,
> or is there a way to set default to yes?

I guess you'd have to modify `parse_option()` a bit (to indicate whether the `child_cfg_option_t` flag, which will be called `OPT_NO_...`, should be added if the configured value is TRUE or FALSE).

> how does ipv4_clear_pmtudisc sound for XFRM_STATE_NOPMTUDISC?

Maybe _clear_df_ could work, not sure if the _ipv4_ prefix is necessary if it's documented to only affect outbound IPv4 tunnel mode SAs. _clear_ecn_ doesn't seem entirely fitting, though, as it really controls whether data is copied from/to the headers (in both directions). _clear_dscp_, on the other hand, could work too I guess (it only affects the outer IP header).

Links2004:
anything to change on the PR?

Links2004:
ok, now `XFRM_SA_XFLAG_DONT_ENCAP_DSCP` makes sense.

for `copy_dscp` a multi state makes sense.
but for `copy_ecn` I am not so sure, the only possible use case where a in / out out control is needed for ECN is when the network has a Asymmetric Routing and one direction can not handle the ECN flag.
but in this case Its better to disable ECN fully since the tunnel will fail if the route change and all traffic goes via devices that not support ECN.

tobiasbrunner:
> but for copy_ecn I am not so sure, the only possible use case where a in / out out control is needed for ECN is when the network has a Asymmetric Routing and one direction can not handle the ECN flag.
> but in this case Its better to disable ECN fully since the tunnel will fail if the route change and all traffic goes via devices that not support ECN.

OK, thanks for the explanation. I removed that part of the second patch.

PR #105, opened 2018-06-08
Allow charon to change group on files before dropping caps

micah-morton:
Allow charon to start as a non-root user without CAP_CHOWN and still be
able to change the group on files that need to be accessed by charon
after capabilities have been dropped. This requires user charon starts
as to have access to socket/pidfile directory as well as belong to the
group that charon will run as after dropping capabilities.

### DIFF
--- a/src/charon/charon.c
+++ b/src/charon/charon.c
@@ -231,15 +231,24 @@ static bool check_pidfile()
 			DBG1(DBG_LIB, "setting FD_CLOEXEC for '"PID_FILE"' failed: %s",
 				 strerror(errno));
 		}
-		/* Only fchown() the pidfile if we have CAP_CHOWN. Otherwise,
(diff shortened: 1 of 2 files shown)


tobiasbrunner:
Thanks. There were some code style issues, and the code in `stream_service_unix.c` wasn't fully correct (the group was changed again if the first `chown()` call was successful). I pushed a modified version to master.

micah-morton:
Oops. I had it right in the original ChromeOS patch but tried to get too cute with cleaning it up :)

https://chromium-review.googlesource.com/c/chromiumos/overlays/chromiumos-overlay/+/1044968

Thanks

tobiasbrunner:
> Oops. I had it right in the original ChromeOS patch but tried to get too cute with cleaning it up :)

Hehe, yeah, I've been there.

PR #107, opened 2018-07-05
Enable macOS native AES_GCM_ICV16 support

angryzor:
macOS seems to support AES_GCM_ICV16 natively using pfkey.

This change enables AES_GCM if the corresponding definition is detected in the ipsec headers.
With this change it is no longer necessary to use the `libipsec` module to use AES_GCM_ICV16 on macOS.

NOTE:
Because the Apple Configurator section for the native IKEv2 implementation had an option for AES_GCM I tried selecting this pfkey encryption method for AES_GCM_ICV16 and concluded that it worked after some simple testing. I tested this on an IPv4 tunnel, saw that the peer on the other side correctly identified my connection as aes-gcm-256, and was able to ping and ssh various computers on the remote network.

If there is some limitation to this implementation that makes it unfit to be included with strongswan yet, please disregard this PR.

### DIFF
--- a/src/libcharon/plugins/kernel_pfkey/kernel_pfkey_ipsec.c
+++ b/src/libcharon/plugins/kernel_pfkey/kernel_pfkey_ipsec.c
@@ -891,6 +891,9 @@ static kernel_algorithm_t encryption_algs[] = {
 	{ENCR_AES_GCM_ICV12,		SADB_X_EALG_AES_GCM_ICV12	},
 	{ENCR_AES_GCM_ICV16,		SADB_X_EALG_AES_GCM_ICV16	},
 #endif
+#ifdef SADB_X_EALG_AES_GCM
+	{ENCR_AES_GCM_ICV16,		SADB_X_EALG_AES_GCM		},
+#endif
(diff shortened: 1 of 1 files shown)


angryzor:
Maybe it would be better to add a guard on __APPLE__ being defined? Because now the algorithm will be registered twice if SADB_X_EALG_AES_GCM_ICV16 is also defined...

tobiasbrunner:
Thanks. [Looks like](https://github.com/apple/darwin-xnu/blob/master/bsd/net/pfkeyv2.h#L435) macOS also supports ChaCha20/Poly1305. I've pushed your commit (slightly modified) and one that adds the latter to master.

> Maybe it would be better to add a guard on APPLE being defined? Because now the algorithm will be registered twice if SADB_X_EALG_AES_GCM_ICV16 is also defined...

Yes, I noticed that too. While I think it's unlikely that there will be a header defining both, I changed your `#ifdef` to an `#elif defined`. The variant with 16-byte ICV is [the only one that's recommended](https://tools.ietf.org/html/rfc8221#section-5) anyway, so I don't think Apple will add the shorter variants later.

PR #109, opened 2018-07-26
Add Botan plugin to libstrongswan

securitykernel:
Adds a plugin for the [Botan](https://botan.randombit.net) crypto library to libstrongswan. Botan is a C++ library, buth with C wrapper functions ("ffi"). 

The plugin implements the following interfaces:

* crypter
* dh
* ecdh
* ec
* gcm
* hasher
* hmac
* rng
* rsa

The plugin current only supports algorithms that are recommended by the BSI in its technical guidelines, e.g., for hashing SHA-2, but not MD5. We also added SHA-1, as it seems to be a hard requirement for strongswan.

We had to remove tests for generating 768 bit RSA keys, as Botan only allows >=1,024 bit RSA keys to be generated (everything <1,024 bits is considered too weak).

/cc @randombit

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -126,6 +126,7 @@ ARG_DISBL_SET([aes],            [disable AES software implementation plugin.])
 ARG_ENABL_SET([af-alg],         [enable AF_ALG crypto interface to Linux Crypto API.])
 ARG_ENABL_SET([bliss],          [enable BLISS software implementation plugin.])
 ARG_ENABL_SET([blowfish],       [enable Blowfish software implementation plugin.])
+ARG_ENABL_SET([botan],          [enables the Botan crypto plugin.])
 ARG_ENABL_SET([ccm],            [enables the CCM AEAD wrapper crypto plugin.])
 ARG_ENABL_SET([chapoly],        [enables the ChaCha20/Poly1305 AEAD plugin.])
 ARG_DISBL_SET([cmac],           [disable CMAC crypto implementation plugin.])
@@ -1162,6 +1163,12 @@ if test x$gcrypt = xtrue; then
 	)
 fi
 
+if test x$botan = xtrue; then
+	PKG_CHECK_MODULES(botan, [botan-2])
+	AC_SUBST(botan_CFLAGS)
+	AC_SUBST(botan_LIBS)
+fi
+
 if test x$uci = xtrue; then
 	AC_CHECK_LIB([uci],[uci_alloc_context],[LIBS="$LIBS"],[AC_MSG_ERROR([UCI library libuci not found])],[])
 	AC_CHECK_HEADER([uci.h],,[AC_MSG_ERROR([UCI header uci.h not found!])])
@@ -1412,6 +1419,7 @@ ADD_PLUGIN([pem],                  [s charon scepclient pki scripts manager meds
 ADD_PLUGIN([padlock],              [s charon])
 ADD_PLUGIN([openssl],              [s charon scepclient pki scripts manager medsrv attest nm cmd aikgen])
 ADD_PLUGIN([gcrypt],               [s charon scepclient pki scripts manager medsrv attest nm cmd aikgen])
+ADD_PLUGIN([botan],                [s charon scepclient pki scripts manager medsrv attest nm cmd aikgen])
 ADD_PLUGIN([af-alg],               [s charon scepclient pki scripts medsrv attest nm cmd aikgen])
 ADD_PLUGIN([fips-prf],             [s charon nm cmd])
 ADD_PLUGIN([gmp],                  [s charon scepclient pki scripts manager medsrv attest nm cmd aikgen fuzz])
@@ -1581,6 +1589,7 @@ AM_CONDITIONAL(USE_SQLITE, test x$sqlite = xtrue)
 AM_CONDITIONAL(USE_PADLOCK, test x$padlock = xtrue)
 AM_CONDITIONAL(USE_OPENSSL, test x$openssl = xtrue)
 AM_CONDITIONAL(USE_GCRYPT, test x$gcrypt = xtrue)
+AM_CONDITIONAL(USE_BOTAN, test x$botan = xtrue)
 AM_CONDITIONAL(USE_AGENT, test x$agent = xtrue)
 AM_CONDITIONAL(USE_KEYCHAIN, test x$keychain = xtrue)
 AM_CONDITIONAL(USE_PKCS11, test x$pkcs11 = xtrue)
@@ -1860,6 +1869,7 @@ AC_CONFIG_FILES([
 	src/libstrongswan/plugins/padlock/Makefile
 	src/libstrongswan/plugins/openssl/Makefile
 	src/libstrongswan/plugins/gcrypt/Makefile
+	src/libstrongswan/plugins/botan/Makefile
 	src/libstrongswan/plugins/agent/Makefile
 	src/libstrongswan/plugins/keychain/Makefile
 	src/libstrongswan/plugins/pkcs11/Makefile
--- a/src/libstrongswan/Makefile.am
+++ b/src/libstrongswan/Makefile.am
@@ -565,6 +565,13 @@ if MONOLITHIC
 endif
 endif
 
+if USE_BOTAN
+  SUBDIRS += plugins/botan
+if MONOLITHIC
+  libstrongswan_la_LIBADD += plugins/botan/libstrongswan-botan.la
+endif
+endif
+
 if USE_FIPS_PRF
   SUBDIRS += plugins/fips_prf
 if MONOLITHIC
--- a/src/libstrongswan/plugins/botan/Makefile.am
+++ b/src/libstrongswan/plugins/botan/Makefile.am
@@ -1,0 +1,30 @@
+AM_CPPFLAGS = \
+	-I$(top_srcdir)/src/libstrongswan
+
+AM_CFLAGS = \
+	$(PLUGIN_CFLAGS) \
+	$(botan_CFLAGS)
+
+if MONOLITHIC
+noinst_LTLIBRARIES = libstrongswan-botan.la
+else
+plugin_LTLIBRARIES = libstrongswan-botan.la
+endif
+
+libstrongswan_botan_la_SOURCES = \
+	botan_plugin.h botan_plugin.c \
+	botan_rng.h botan_rng.c \
+	botan_hasher.h botan_hasher.c \
+	botan_hmac.h botan_hmac.c \
+	botan_crypter.h botan_crypter.c \
+	botan_rsa_public_key.h botan_rsa_public_key.c \
+	botan_rsa_private_key.h botan_rsa_private_key.c \
+	botan_diffie_hellman.h botan_diffie_hellman.c \
+	botan_ec_diffie_hellman.h botan_ec_diffie_hellman.c \
+	botan_ec_public_key.h botan_ec_public_key.c \
+	botan_ec_private_key.h botan_ec_private_key.c \
+	botan_util.c \
+	botan_gcm.h botan_gcm.c
+
+libstrongswan_botan_la_LDFLAGS = -module -avoid-version
+libstrongswan_botan_la_LIBADD = $(botan_LIBS)
--- a/src/libstrongswan/plugins/botan/botan_crypter.c
+++ b/src/libstrongswan/plugins/botan/botan_crypter.c
@@ -1,0 +1,195 @@
+/*
+ * Copyright (C) 2018 René Korthaus
+ * Copyright (C) 2018 Konstantinos Kolelis
+ * Copyright (C) 2018 Tobias Hommel
+ * Rohde & Schwarz Cybersecurity GmbH
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "botan_crypter.h"
+
+#include <botan/ffi.h>
+
+typedef struct private_botan_crypter_t private_botan_crypter_t;
+
+/**
+ * Private data of botan_crypter_t
+ */
+struct private_botan_crypter_t {
+
+	/**
+	 * Public part of this class
+	 */
+	botan_crypter_t public;
+
+	/**
+	 * The key
+	 */
+	chunk_t	key;
+
+	/**
+	 * The cipher name
+	 */
+	const char* cipher_name;
+};
+
+/**
+ * Do the actual en/decryption
+ */
+static bool crypt(private_botan_crypter_t *this, chunk_t data, chunk_t iv,
+				  chunk_t *dst, uint32_t init_flag)
+{
+	botan_cipher_t cipher;
+	size_t output_written = 0;
+	size_t input_consumed = 0;
+	uint8_t *in, *out;
+	bool success = FALSE;
+
+	in = data.ptr;
+	if (dst)
+	{
+		*dst= chunk_alloc(data.len);
+		out = dst->ptr;
+	}
+	else
+	{
+		out = data.ptr;
+	}
+
+	if (botan_cipher_init(&cipher, this->cipher_name, init_flag))
+	{
+		return FALSE;
+	}
+	
+	if (!botan_cipher_set_key(cipher, this->key.ptr, this->key.len)
+		&& !botan_cipher_start(cipher, iv.ptr, iv.len)
+		&& !botan_cipher_update(cipher, BOTAN_CIPHER_UPDATE_FLAG_FINAL, out,
+							   data.len, &output_written, in, data.len,
+							   &input_consumed)
+		&& (output_written == input_consumed))
+	{
+		success = TRUE;
+	}
+
+	botan_cipher_destroy(cipher);
+	return success;
+}
+
+METHOD(crypter_t, decrypt, bool,
+		private_botan_crypter_t *this, chunk_t data, chunk_t iv, chunk_t *dst)
+{
+	return crypt(this, data, iv, dst, BOTAN_CIPHER_INIT_FLAG_DECRYPT);
+}
+
+
+METHOD(crypter_t, encrypt, bool,
+		private_botan_crypter_t *this, chunk_t data, chunk_t iv, chunk_t *dst)
+{
+	return crypt(this, data, iv, dst, BOTAN_CIPHER_INIT_FLAG_ENCRYPT);
+}
+
+METHOD(crypter_t, get_block_size, size_t,
+		private_botan_crypter_t *this)
+{
+	return AES_BLOCK_SIZE;
+}
+
+METHOD(crypter_t, get_iv_size, size_t,
+		private_botan_crypter_t *this)
+{
+	return AES_BLOCK_SIZE;
+}
+
+METHOD(crypter_t, get_key_size, size_t,
+		private_botan_crypter_t *this)
+{
+	return this->key.len;
+}
+
+METHOD(crypter_t, set_key, bool,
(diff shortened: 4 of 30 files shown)


tobiasbrunner:
Thanks a lot! I'll have a closer look at this after my vacation.

Some quick comments:

> but not MD5

Which is required for RADIUS and obviously EAP-MD5, and also for non-PKCS#8 encoded, encrypted private keys. Even worse, EAP-MSCHAPv2 (e.g. on a client, servers could use EAP-RADIUS, if MD5 was supported) requires MD4 and DES-ECB.

I guess users could always enable our own implementations, if necessary. But if the library supports these algorithms, I don't really see a reason not to support them in the plugin (adding them to e.g. the IKE proposals is a different thing - they are not added to the default proposal, by the way - but that's not something a crypto plugin, which just provides the primitives, has to care about).

If there are scenarios in which you still wanted to avoid loading the algorithms you could probably add a setting to disable them, or simply build Botan without support for them (I guess Botan does not have something like OpenSSL's FIPS-mode in which using such algorithms simply fails).

> We also added SHA-1, as it seems to be a hard requirement for strongswan.

It's a requirement for IKEv2 (e.g. for NAT-D payloads or the Hash and URI functionality) and for X.509 (key IDs), and we also use it for some other applications (e.g. COOKIE payloads or message hashes for retransmission detection).

> We had to remove tests for generating 768 bit RSA keys, as Botan only allows >=1,024 bit RSA keys to be generated (everything <1,024 bits is considered too weak).

I've no problem with that. Actually, I think there is really no point in supporting such weak keys anymore (not even for backwards compatibly). So we could probably remove support for it in all other plugins too.

securitykernel:
Regarding CI: IMHO it would be useful if we add CI jobs with Botan similar to the ones for openssl and gcrypt. The plugin needs a very recent Botan, even the [2.7.0](https://botan.randombit.net/news.html#version-2-7-0-2018-07-02) is not recent enough, because I added [ffi functions for PKCS#1](https://github.com/randombit/botan/pull/1621) to Botan just a few weeks ago. A 2.7.0 with this patch would work out, or a recent master of course.

randombit:
> The plugin current only supports algorithms that are recommended by the BSI in its technical guidelines, e.g., for hashing SHA-2, but not MD5.

IMO for MD5 at least it makes sense to enable it in the plugin if it enabled within the library. That would work in either stock or BSI configurations.

> I guess Botan does not have something like OpenSSL's FIPS-mode in which using such algorithms simply fails

Actually for FFI layer this is exactly how it works; calling say `botan_hash_init` with a unknown algorithm (either due to it being not compiled into the library or simply absent in the version currently being used) it simply returns an error (`BOTAN_FFI_ERROR_NOT_IMPLEMENTED`). So it is possible to check at runtime if a feature is available or not.

securitykernel:
> IMO for MD5 at least it makes sense to enable it in the plugin if it enabled within the library. That would work in either stock or BSI configurations.

Agree. 57bbcfd adds support for MD5 in the hasher. I assume we don't want to support MD5 signatures, do we?

randombit:
> I assume we don't want to support MD5 signatures, do we?

I hope not? But I don't know what IPsec needs here.

tobiasbrunner:
Thanks again, also to your fellow co-workers.

I now reviewed the code and it looks really good. I still found some stuff to change :)

Before providing some details on some of the changes, I have a few questions/points regarding the code:

* Is it necessary to call `botan_mp_clear()` before calling `botan_mp_destroy()` for sensitive data? Or is the data of all objects allocated by Botan cleared before freeing it?
* When generating RSA and ECDSA keys you create "user" type `botan_rng_t` instances. However, in the `rng_t` implementation you mapped `RNG_TRUE`, which we usually use for private keys, to a "system" type `botan_rng_t`. So shouldn't the latter also be used when generating private keys?
* In the`rng_t` implementation a new `botan_rng_t` instance is created for every call to `get_bytes()`. Would it be possible to create this instance in the constructor so it is bound to the lifetime of the `rng_t`? Or is there a specific reason for not doing that?
* I guess, from an API standpoint the `BOTAN_FFI_ERROR_INSUFFICIENT_BUFFER_SPACE` double-calls are fine (it's what we already know from OpenSSL). However, due to how these functions are currently implemented in Botan this is highly inefficient. DH secrets are derived twice, signatures are created twice, both very costly asymmetric crypto operations, and data is encoded/hashed twice.
Are there plans to improve this in Botan? Either by adding new getters for the length of these outputs, similar to `botan_hash_output_length()`, or by changing the implementation, for instance, adding an early check if output must be provided and otherwise just determine and return the length (an even more convenient API would be an allocation by Botan if a pointer to a NULL pointer is provided - i.e. not a NULL pointer, but a pointer to a variable that points to NULL). Otherwise, it might be better to use reasonably sized buffers on the stack, calling these functions once and then copying the buffered data to a properly sized chunk after the actual length is known.

I pushed your patches with some fixups and refactorings to the _gh109-botan-plugin_ branch. Some notes on these changes:

 * I changed the return value of utility functions from `int` to `bool` because that's what we usually use.
 * The DH constructor for MODP_CUSTOM has to be variadic (compatibility with Apple's ARM64 implementation). I removed the redundant constructor for other groups.
 * The fingerprinting and public key encoding has been moved to utility functions as the code is shared by RSA and ECDSA (also doesn't require the _pkcs1_ plugin anymore for this).
 * Likewise was the signature creation moved to a utility function that's shared by RSA and ECDSA.
 * Another utility function maps our hash identifiers to Botan strings (used by the hasher and for PSS signatures).
 * I also simplified the generation of the strings used for EMSA PSS RSA signatures.
 * To the ECDSA private key encoding I added the named curve OID (was actually required to parse such keys again) and also the public key (necessary for older OpenSSL versions). This required keeping track of the OID as there is no way to get that out of Botan (i.e. there is no FFI wrapper for `get_curve_oid()`).
 * The last commit changes how public and private keys are parsed, which allows parsing via `KEY_ANY` (`--type priv/pub` in `pki`), loads unencrypted PKCS#8 encoded keys directly without _pkcs8_ plugin (same for other keys and the _pkcs1_ plugin), and simplifies `private_key_t::get_public_key()`. 

I also added some builds on Travis CI (currently building against Botan's master branch, I guess until 2.8.0 is released). However, in order to run the unit tests with our own memory leak detective two Botan features had to be disabled for that build:
 *  The _threads_ OS feature, which causes frees of "unknown" memory in `pthread_create()` (via `std::async` when generating signatures), i.e. memory not allocated by our `malloc()` override is passed to our `free()` override - strangely this is no issue when we call that function ourselves in our threading unit tests (they run before the Botan-related tests), however, the first test that runs afterwards and creates a thread will also trigger such an error, later tests are again fine (so it seems some memory is replaced when a thread is created and when the C++ standard library does this it somehow bypasses our allocator, but not when destroying the data). I did push a possible workaround for this issue, but I'm not yet sure if that's the right approach (I think I'd first like to understand what exactly happens here).
 * The _locking_allocator_ module, which is implemented by a static singleton (this is reported as leak because it will never be freed). As this is triggered by the first function that uses it to allocate memory, potentially a lot of functions would have to be whitelisted to avoid reports (not so much for the unit-tests, but more the regression tests) and due to C++'s name mangling we can only whitelist FFI functions, which prevents us from actually detecting leaks in our own code caused by calling these functions, so that's not ideal.
 * Still I had to whitelist two ECC related FFI functions because of the static EC group hashtable that is allocated by them (and can't be disabled).
 * I guess we could later further reduce the module list to improve the build time (e.g. pkcs11 is not necessary on Travis).

So far I only tested this with [`pki`](https://wiki.strongswan.org/projects/strongswan/wiki/IpsecPKI) and the unit tests (test vectors), but I think Andreas intends to add regression tests similar to those for [openssl](https://www.strongswan.org/testing/testresults/openssl-ikev2/) and gcrypt.

randombit:
I can help answer a few of these which are more related to the Botan library vs the plugin implementation:

> Is it necessary to call botan_mp_clear() before calling botan_mp_destroy()

Not necessary, destroying the object will cause it to be zeroed. (This should be true for all object types)

> This required keeping track of the OID as there is no way to get that out of Botan (i.e. there is no FFI wrapper for get_curve_oid()).

I can add this for 2.8. What is most useful format for the output? A string representation of the OID? It should be represented as a sequence of `uint32_t` but that is a little awkward to use.

> Problems with BOTAN_FFI_ERROR_INSUFFICIENT_BUFFER_SPACE

You are correct the current design is very inefficient because there is no way to know what the correct size should be. I am running into problems with this in the Rust binding right now. Addressing this exhaustively requires a good deal of support elsewhere in the library but I do plan to address at least some of the cases (eg PK signatures+ciphertext values) in time for 2.8.

> an even more convenient API would be an allocation by Botan if a pointer to a NULL pointer is provided - i.e. not a NULL pointer, but a pointer to a variable that points to NULL

The C API avoids passing ownership of any memory across the calling boundary because this is also used for FFI and most other languages have their own allocator/heap. What you propose would certainly be the most efficient and easy to use situation when using the library from C, but it raises the risk of allocator mismatches and subsequent memory corruption for other langs.

> The locking_allocator module, which is implemented by a static singleton (this is reported as leak because it will never be freed).

Well, it will eventually be freed but only on program shutdown, which is probably tricking your leak detector. You can prevent this by setting env variable `BOTAN_MLOCK_POOL_SIZE` to "0". (Or of course just continue to leave the module disabled).

> memory not allocated by our malloc() override is passed to our free() override

Wow that is very strange. Let me know if there is anything we should change upstream to prevent this.

> Still I had to whitelist two ECC related FFI functions because of the static EC group hashtable that is allocated by them

There is an API to clear this table but it is not exposed to FFI right now.

randombit:
I just created a PR to Botan that adds `botan_pk_op_sign_output_length` to query the signature length. I also have patches for getting PK ciphertext+plaintext lengths, but Botan's support for ECIES/DLIES complicates things so I need further work there.

For DH/ECDH, it seems simpler to me for the calling code to use its knowledge of the key parameters; DH knows the output will be at most `p` bytes, and ECDH is at most 521 bits == 65 bytes.

tobiasbrunner:
Thanks for your replies.

> Is it necessary to call botan_mp_clear() before calling botan_mp_destroy()
> 
> Not necessary, destroying the object will cause it to be zeroed. (This should be true for all object types)

Good to know.

> I can add this for 2.8. What is most useful format for the output? A string representation of the OID? It could be represented as a sequence of uint32_t but that is a little awkward to use.

A string would work for us, but we'd have to parse it again as we need the binary ASN.1 encoding (not that much of an issue, just requires two function calls).

But I realize now that we don't actually need this. It's only required because the plugin tries to be consistent with other plugins and produces the simple `ECPrivateKey` encoding for ECDSA private keys (which requires the OID to be parsed again), but we don't really specify in our API that this must be the case. The encoding is just either binary DER or PEM, however, what exactly is encoded is basically up to the plugins as long as they can parse it again. So I think we can just stick with PKCS#8 encoding and simplify the private key exporting a lot by just calling `botan_privkey_export()`. I'll change this in the branch.

> memory not allocated by our malloc() override is passed to our free() override
> 
> Wow that is very strange. Let me know if there is anything we should change upstream to prevent this.

I don't think it's anything your code does, you simply use `std::async`, which spawns the thread. But yeah, it's quite strange.

> I just created a PR to Botan that adds botan_pk_op_sign_output_length to query the signature length. I also have patches for getting PK ciphertext+plaintext lengths, but Botan's support for ECIES/DLIES complicates things so I need further work there.

Awesome, thanks a lot.

> For DH/ECDH, it seems simpler to me for the calling code to use its knowledge of the key parameters; DH knows the output will be at most p bytes, and ECDH is at most 521 bits == 65 bytes.

Makes sense, I'll change the code accordingly.

securitykernel:
Thanks for your review!

> When generating RSA and ECDSA keys you create "user" type botan_rng_t instances. However, in the rng_t implementation you mapped RNG_TRUE, which we usually use for private keys, to a "system" type botan_rng_t. So shouldn't the latter also be used when generating private keys?

In this case, yes.

> In the rng_t implementation a new botan_rng_t instance is created for every call to get_bytes(). Would it be possible to create this instance in the constructor so it is bound to the lifetime of the rng_t? Or is there a specific reason for not doing that?

Actually this was our first shot for the implementation, but our appliance QA run failed with an error in this area. I cannot remember though what the error messages were. Maybe a race condition? Does strongswan require the rng to be threadsafe? Because neither Botan's `System_RNG` nor `HMAC_DRBG` are threadsafe by default. You have to wrap them in a `Serialized_RNG` object.

randombit:
> However, I noticed that the ffi module currently has a dependency on the x509 module, which we currently don't use

I just pushed a change to allow x509 to be optional for ffi, if not enabled all functions just return not implemented error code.

> and I think it would require several FFI additions to match our API

If you decide you want to start using the X509 support within the plugin and need additional FFI functions just let me know.

> A configure option to disable building the CLI, which we also don't need for the unit tests, would also be great

This is reasonable suggestion. For your purposes though, perhaps you just need to `make libs test`?

> Because neither Botan's System_RNG nor HMAC_DRBG are threadsafe by default.

The System RNG *should* be threadsafe. Or, I guess to be more precise, we assume the underlying action (be it reading from `/dev/urandom` or calling `RtlGenRandom` or `arc4random`) is serialized by the operating system or runtime library.

If this is not true then there is a bigger problem because *all* instances of `Botan::System_RNG` actually refer to a singleton object which queries the system RNG. So even if two different threads allocate different `Botan::System_RNG` objects, they are still referring to the same underlying data structure.

For userspace, I just add an option to FFI so "user-threadsafe" requests a serialized RNG. This might help.

tobiasbrunner:
> When generating RSA and ECDSA keys you create "user" type botan_rng_t instances. However, in the rng_t implementation you mapped RNG_TRUE, which we usually use for private keys, to a "system" type botan_rng_t. So shouldn't the latter also be used when generating private keys?
> 
> In this case, yes.

The question is whether it makes that much of a difference in the end. "system" seems to default to `/dev/urandom`, whereas in our _random_ plugin we use `/dev/random` for `RNG_TRUE` (in practice that probably makes not that much of a difference, though, unless the system entropy is low, I guess). And the DRBG created with "user" is seeded by the "system" RNG too by default (although strictly limited to 256 bits entropy for the initial seed, but that probably also doesn't matter much in practice). But I guess using "system" would be more consistent, so I'll change that.

> Actually this was our first shot for the implementation, but our appliance QA run failed with an error in this area. I cannot remember though what the error messages were. Maybe a race condition? Does strongswan require the rng to be threadsafe? Because neither Botan's System_RNG nor HMAC_DRBG are threadsafe by default. You have to wrap them in a Serialized_RNG object.

Interesting. Could you try to find out what the issue was? Because to me it makes more sense to bind `botan_rng_t` instances to the lifetime of `rng_t` instances (in particular for DRBGs).

I don't think they have to be thread-safe in the sense that multiple threads will operate on a single `rng_t` instance concurrently. But there are at all times multiple instances on which different threads operate (some might be used by multiple threads because they are held in shared objects, but not concurrently). But those are pretty much exclusively "user" instances, which should be independent anyway (except for the initial seeding, but if you create a new instance with each call to  `allocate/get_bytes()` there are probably even more chances for races).

> I just pushed a change to allow x509 to be optional for ffi, if not enabled all functions just return not implemented error code.

Perfect, thanks.

> If you decide you want to start using the X509 support within the plugin and need additional FFI functions just let me know.

Great, thanks. Although it's not our first priority (we'll probably focus on wrapping additional algorithms e.g. curve25519/ed25519 or chacha20poly1305).

> For your purposes though, perhaps you just need to make libs test?

You're absolutely right, didn't realize there are separate make targets for these. This and disabling _x509_ reduced the build time of the library from ~360 to ~230 seconds (full build was ~410 seconds).

> For userspace, I just add an option to FFI so "user-threadsafe" requests a serialized RNG. This might help.

Nice, but as mentioned above, we'll hopefully not need it. However, this is missing a `#if defined(BOTAN_TARGET_OS_HAS_THREADS)` wrapper around it (our build without _threads_ OS feature currently fails).

randombit:
> However, this is missing a #if defined(BOTAN_TARGET_OS_HAS_THREADS)

Sorry about that, fixed now. (Building without thread support is not currently checked by our CI, I should add that.)

randombit:
`botan_pk_op_encrypt_output_length` and `botan_pk_op_decrypt_output_length` are now available in master. They return an upper bound on the output lengths. In some cases it is exact, in others (especially for decrypt) it is just a strict upper bound.

randombit:
While in the context of how botan is used in the plugin it is possible to deduce from the parameters the size of the output of a key agreement, I realized that is not possible in the general case. So I just added to botan master a `botan_pk_op_key_agreement_size` which returns the size of the value that a raw (un-KDFed) DH or ECDH key agreement will produce.

securitykernel:
> Interesting. Could you try to find out what the issue was? Because to me it makes more sense to bind botan_rng_t instances to the lifetime of rng_t instances (in particular for DRBGs).

We are working on that.

tobiasbrunner:
Thanks @randombit.  I've updated the branch to make use of these functions (including DH/ECDH). Now the only open issue I see is where `botan_rng_t` instances are initialized in `rng_t`. Looking forward to your findings @securitykernel.

tobiasbrunner:
> Re RNG: After running the first 3-5 QA scenarios, strongswan crashes. Most of the time, it's in get_bytes(). If we put the botan_rng_get() call in a mutex, everything is fine.

What QA scenarios are these exactly (i.e. can we reproduce this somehow)? The only call in `get_bytes()`, after moving the `botan_rng_t` initialization, is `botan_rng_get()`, so what exactly causes the crash? Is the pointer invalid (segmentation fault)? Something else? Can you determine for what purpose the RNG that causes this was created? You say most of the time it's in `get_bytes()`, are there other crashes? Could you provide backtraces?

securitykernel:
Our QA setup involves n<5 appliances which build up IKE SAs. During the scenarios we execute, we constantly change keys and certificates and build up new IKE SAs. After some time during the QA run, strongswan crashes and restarts. We can not identify the one scenario that causes the crash, this really happens at different times during the QA run.

Here is one backtrace:

```
    -> ??:?
  /usr/lib64/libbotan-2.so.4 @ 0x7ffff4e0f000 (_ZN5Botan9HMAC_DRBG6updateEPKhm+0x20d) [0x7ffff4fe4e0d]
    -> ??:?
  /usr/lib64/libbotan-2.so.4 @ 0x7ffff4e0f000 (_ZN5Botan9HMAC_DRBG20randomize_with_inputEPhmPKhm+0xd1) [0x7ffff4fe5131]
    -> ??:?
  /usr/lib64/libbotan-2.so.4 @ 0x7ffff4e0f000 (_ZN5Botan21RandomNumberGenerator23randomize_with_ts_inputEPhm+0x6c) [0x7ffff4fe5e0c]
    -> ??:?
  /usr/lib64/libbotan-2.so.4 @ 0x7ffff4e0f000 (botan_rng_get+0x2c) [0x7ffff4f09ccc]
    -> ??:?
  /usr/lib64/ipsec/plugins/libstrongswan-botan.so @ 0x7ffff5373000 [0x7ffff5376105]
    -> strongswan/src/libstrongswan/plugins/botan/botan_rng.c:51
  /usr/lib64/ipsec/libcharon.so.0 @ 0x7ffff78b7000 [0x7ffff78f8c91]
    -> strongswan/src/libcharon/sa/ike_sa_manager.c:1007 (discriminator 1)
  /usr/lib64/ipsec/libcharon.so.0 @ 0x7ffff78b7000 [0x7ffff78f95bc]
    -> strongswan/src/libcharon/sa/ike_sa_manager.c:1201
  /usr/lib64/ipsec/libcharon.so.0 @ 0x7ffff78b7000 [0x7ffff78fa067]
    -> strongswan/src/libcharon/sa/ike_sa_manager.c:1464
  /usr/lib64/ipsec/libcharon.so.0 @ 0x7ffff78b7000 [0x7ffff78cd212]
    -> strongswan/src/libcharon/control/controller.c:415
  /usr/lib64/ipsec/libcharon.so.0 @ 0x7ffff78b7000 [0x7ffff78cde4d]
    -> strongswan/src/libcharon/control/controller.c:532
  /usr/lib64/ipsec/plugins/libstrongswan-vici.so @ 0x7ffff2f1e000 [0x7ffff2f2b75b]
    -> strongswan/src/libcharon/plugins/vici/vici_control.c:205
  /usr/lib64/ipsec/plugins/libstrongswan-vici.so @ 0x7ffff2f1e000 (process_request+0xe5) [0x7ffff2f255e5]
    -> strongswan/src/libcharon/plugins/vici/vici_dispatcher.c:291
  /usr/lib64/ipsec/plugins/libstrongswan-vici.so @ 0x7ffff2f1e000 [0x7ffff2f259c1]
    -> strongswan/src/libcharon/plugins/vici/vici_dispatcher.c:347
  /usr/lib64/ipsec/plugins/libstrongswan-vici.so @ 0x7ffff2f1e000 [0x7ffff2f21cfe]
    -> strongswan/src/libcharon/plugins/vici/vici_socket.c:509
  /usr/lib64/ipsec/libstrongswan.so.0 @ 0x7ffff7b5e000 [0x7ffff7b9d611]
    -> strongswan/src/libstrongswan/processing/jobs/callback_job.c:78
  /usr/lib64/ipsec/libstrongswan.so.0 @ 0x7ffff7b5e000 [0x7ffff7b9e2cb]
    -> strongswan/src/libstrongswan/processing/processor.c:235
  /usr/lib64/ipsec/libstrongswan.so.0 @ 0x7ffff7b5e000 [0x7ffff7bb35c6]
    -> strongswan/src/libstrongswan/threading/thread.c:332 (discriminator 3)
  /lib64/libpthread.so.0 @ 0x7ffff718f000 [0x7ffff71962d0]
    -> ??:?
  /lib64/libc.so.6 @ 0x7ffff6be8000 (clone+0x6d) [0x7ffff6cd5f4d]
    -> ??:?
```

This is our `botan_rng.c`: [botan_rng.c.txt](https://github.com/strongswan/strongswan/files/2352822/botan_rng.c.txt)

tobiasbrunner:
I've update the branch again. If Botan is built with threading support, `user-threadsafe` is now used. For the tests with leak-detective (i.e. if threading is disabled) `user` is used instead.

tobiasbrunner:
Merged to master. Thanks again!

PR #113, opened 2018-10-02
swanctl: Fix typos in usage for swanctl rekey/terminate commands

selsky:


tobiasbrunner:
Applied to master, thanks.

PR #114, opened 2018-10-06
dhcp: use "knl" subsystem for logging of RNG failure, not "cfg"

selsky:
Follow example set in src/charon-tkm/src/tkm/tkm_kernel_ipsec.c

This is not a configuration error.


tobiasbrunner:
It also has nothing to do with the IPsec/network kernel interface (for which KNL is used). CFG is used here because the plugin is a configuration backend (providing IP addresses).

PR #115, opened 2018-10-10
task_manager_v1: ensure DPD_ACK is sent in time

lofvir:
If allot of QUICK_MODE tasks is queued and other side
sends a DPD request, there is a good chance for timeouts.

Observerd this in cases where other side is quote slow in responding
QUICK_MODE requests (e.g. Cisco ASA v8.x) and about 100 CHILD_SA's are to be spawn.

Instead of giving DPD in general precedence over QUICK_MODE tasks,
the task has been split up. Now it is possible to execute the DPD_ACK
task next, while the regular initiation of DPD requests remains lower
precedented than QUICK_MODE tasks.


tobiasbrunner:
Looks OK.

> Instead of giving DPD in general precedence over QUICK_MODE tasks,

Is that actually a concern? If Quick Modes are still actively established, no DPD will be queued (because the task manager is "busy"). So why wouldn't we generally give DPD tasks precedence over QM tasks?

lofvir:
That was just a precaution to avoid changing existing behaviour for outgoing DPD requests.
Assuming we have 200 QUICK_MODE tasks queued, a slow peer on the right and a pretty short DPD interval: would charon queue up a DPD task while processing the 200 QUICK_MODE requests (e.g. in between two of those tasks)?

tobiasbrunner:
> would charon queue up a DPD task while processing the 200 QUICK_MODE requests (e.g. in between two of those tasks)?

No, because each of the Quick Mode exchanges will act as DPD exchange, there is a check for active tasks in `ike_sa_t::send_dpd()`: https://github.com/strongswan/strongswan/blob/f18945dfea5d12566aa010513a8fa584910a5e26/src/libcharon/sa/ike_sa.c#L769-L773
That's why I think this could be simplified (i.e. just move the existing if-block before the QM if-block).

lofvir:
Given your feedback i'll update the PR.
Thank you for the clarification.

Tobias Brunner <notifications@github.com> schrieb am Fr., 19. Okt. 2018,
09:43:

PR #116, opened 2018-10-23
OpenSSL plugin invalid bounds check

ambiso:
The openssl plugin contains an invalid length check.

It checks if a given `keyid` would fit into the `keyname` buffer in base64 encoding, however then the `keyid` is serialized in hex, resulting in a possible buffer overflow:

```C
       if (sizeof(keyname) - strlen(keyname) <= keyid.len * 4 / 3 + 1)
       {
                return NULL;
       }
       chunk_to_hex(keyid, keyname + strlen(keyname), FALSE);
```

Additionally I removed the `keyid.len > 40` bounds check as it's done implicitly by checking if the `keyid` would fit in the buffer.

Furthermore I removed the extra semicolon on line 373.

### DIFF
--- a/src/libstrongswan/plugins/openssl/openssl_plugin.c
+++ b/src/libstrongswan/plugins/openssl/openssl_plugin.c
@@ -370,7 +370,7 @@ static private_key_t *openssl_private_key_connect(key_type_t type,
 #ifndef OPENSSL_NO_ENGINE
 	char *engine_id = NULL;
 	char keyname[BUF_LEN];
-	chunk_t keyid = chunk_empty;;
+	chunk_t keyid = chunk_empty;
 	EVP_PKEY *key;
 	ENGINE *engine;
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks for the patches, applied to master.

> resulting in a possible buffer overflow

Fortunately, the length check you removed prevented that (at least since b2266280efc81c84984de167ac2b212851eda608). The buffer has a size of 512 bytes and at most 92 bytes are written to it (40 hex bytes plus the - theoretically negative - slot number and colon). However, at the time this code was added (0b8b66405682ef1bb1217384aa19e87f97b22b34), the buffer was only 64 bytes long so the incorrect Base64 length check was more problematic then (the 40 byte check seemed to be intended to prevent overflowing the 64 byte buffer under the assumption of a Base64 encoding). Since this code operates on configuration data it's not that much of an issue, though.

PR #118, opened 2018-12-05
Extend Strongswan to support IBM TSS

kornelduleba:
This path adds support for the [IBM TSS](https://sourceforge.net/projects/ibmtpm20tss/l) library.
The main advantage of which is that it works on FreeBSD, whereas the currently supported one does not.
Create new tss plugin basing on tpm_tss_tss2_v2.c.
Modify configure.ac to recognize the new plugin.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -312,6 +312,7 @@ ARG_ENABL_SET([perl-cpan],      [enable build of provided perl CPAN module.])
 ARG_ENABL_SET([perl-cpan-install],[enable installation of provided CPAN module.])
 ARG_ENABL_SET([tss-trousers],   [enable the use of the TrouSerS Trusted Software Stack])
 ARG_ENABL_SET([tss-tss2],       [enable the use of the TSS 2.0 Trusted Software Stack])
+ARG_ENABL_SET([tss-ibmtss],    [enable the use of the TSS 2.0 IBM Software Stack])
 
 # compile options
 ARG_ENABL_SET([coverage],       [enable lcov coverage report generation.])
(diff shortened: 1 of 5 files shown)


tobiasbrunner:
I don't like the code duplication. There are already two versions because the [System API (SAPI)](https://trustedcomputinggroup.org/resource/tcg-tss-2-0-system-level-api-sapi-specification/) or even the base TSS/TPM2 specs changed over time (constants, structs). The problem is that the IBM TSS implements neither version, it's similar to the v1 implementation, yet uses different (never standardized by the TCG) functions and has other minor differences.

The v2 implementation uses the current SAPI, as it is implemented by libtss2-sys. So to me it seems that instead of adding another near-copy, the better approach would be to get the IBM TSS developers to actually implement the SAPI/TSS/TPM2 specs properly, so the v2 implementation would only have to be modified slightly to work with either TSS library (that's the whole point of having a standardized interface).

In regards to FreeBSD, it might be possible to get libtss2-sys working with an appropriate TCTI implementation (maybe from IBM's TSS, if they actually implement that interface).

And two notes regarding your implementation, it's missing `quote()`, which is quite essential for several use cases, and please refer to the [developer documentation](https://wiki.strongswan.org/projects/strongswan/wiki/DeveloperDocumentation) on our wiki for code style and license information.

semihalf-Duleba-Kornel:
Sorry for not responding for so long.
Basing on your comment above I deducted that you're quite reluctant about merging this PR.
Would you consider doing so if I addressed all your concerns regarding style, missing PCR related functionality and anything else that would come up later.
As for running libtss2-sys on FreeBSD.
Before deciding to go with IBM TSS I spent some time trying to get it to work, but with no success.
Also the TPM driver in FreeBSD handles concurrent requests, it pretty much works in a similar way to /dev/tpmrm0 on Linux.

PR #119, opened 2018-12-07
add missing calls to ENGINE_finish()

arvedarved:
see https://www.openssl.org/docs/man1.1.0/crypto/ENGINE_finish.html
for details.

Without calls to ENGINE_finish() memory is leaked.


tobiasbrunner:
Thanks for the report. Instead of just adding calls to `ENGINE_finish()`, I think we can just replace the later calls to `ENGINE_free()` with that (according to the docs `ENGINE_init()` acquires a functional reference, which also holds an implicit structural reference, which is released by `ENGINE_finish()`). See _gh119-openssl-engine_ branch.

arvedarved:
looks good to me, thanks. (haven't tested it)

tobiasbrunner:
Thanks, applied to master.

PR #120, opened 2018-12-12
swanctl: Add env var for swanctl conf dir

sunghew:
Currently, `SWANCTLDIR` is set at compile time. This hinders the distribution of any strongswan binary utilizing `swanctl` e.g `charon-systemd`. In contrast, `STRONGSWAN_CONF`, which holds the path to `strongswan.conf`, is set at runtime.

To fix this, this PR adds the environment variable `SWANCTL_DIR` to hold the path to the `swanctl` directory. This variable overrides `SWANCTLDIR`.

### DIFF
--- a/src/swanctl/commands/load_all.c
+++ b/src/swanctl/commands/load_all.c
@@ -31,7 +31,7 @@ static int load_all(vici_conn_t *conn)
 	bool clear = FALSE, noprompt = FALSE;
 	command_format_options_t format = COMMAND_FORMAT_NONE;
 	settings_t *cfg;
-	char *arg, *file = SWANCTL_CONF;
+	char *arg, *file = swanctl_conf();
 	int ret = 0;
 
 	while (TRUE)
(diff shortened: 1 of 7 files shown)


tobiasbrunner:
>  This hinders the distribution of any strongswan binary utilizing `swanctl` e.g `charon-systemd`.

How so exactly?

Also, since the path to `swanctl.conf` is configurable for each command, we could maybe just make these paths relative to that file. I pushed something to that end to the _gh120-swanctl-dir_ branch.

tobiasbrunner:
Thanks, I'll probably merge it soon. Would you mind responding to my question, I'm really interested in what problems you ran into during "distribution" of the binaries (was that just an issue because the binaries were incorrectly configured? or really something unavoidable?).

sunghew:
Ah, sorry. I compiled strongswan on some arbitrary path in my dev environment, and then I copied the binaries to some other machine. I ran `charon-systemd`, but then it would crash since it couldn't find the swanctl config files/directories. I didn't realize that the `-f` flag was nested under some of the swanctl commands, but that would have solved my problem. My issue was avoidable.

tobiasbrunner:
I see, thanks. Pushed the patch to master.

PR #121, opened 2019-01-20
libtpmtss: Read public key exponent instead of assuming its value.

krinfels:
Up to now it was assumed that the RSA public key exponent is equal to 2^16+1. Although this is probably true in the most if not all cases, it is not correct according to the TPM specification.

This patch fixes that by reading the exponent from an appropriate of the structure returned by TPM2_ReadPublic.

The above mentioned standard can be find [here](https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf#page=149).


tobiasbrunner:
Thanks, applied to master. Was this an issue in practice?

krinfels:
> Was this an issue in practice?

No, I've found it bug while reading the source code.

PR #122, opened 2019-01-21
adding basic virtual XFRM interface support

Links2004:


maroviher:
I would like to try your patch. What should I configure? Should I create a virtual network interface manually befor I start a VPN connection? How to add that interface? Should I pass an if_id of that virtual interface to {ipsec,strongswan}.conf?

Links2004:
Hi,
i currently use:

```
export IF_NAME=xfrm42
./ip link add xfrm42 type xfrm dev eth0 if_id 42
./ip link set "${IF_NAME}" up
./ip addr add "192.168.234.2" dev "${IF_NAME}" 
./ip route add "192.168.234.0/24" dev "${IF_NAME}" 

sysctl -w "net.ipv4.ip_forward=1"
```

to create the interface.

Links2004:
via the child parameter *if_id* in a swanctl config

maroviher:
It would be nice if you attach the patch for iproute here in discussion and configuration files for strongswan in your test setup.

maroviher:
> ./ip route add "192.168.234.0/24" dev "${IF_NAME}"
strongswan installs routes automatically in to the table 220, should it be disabled?

maroviher:
XFRMA_IF_ID attribute passed to kernel only in XFRM_MSG_NEWSA but not in XFRM_MSG_NEWPOLICY. The following [condition](https://elixir.bootlin.com/linux/latest/source/net/xfrm/xfrm_user.c#L1620) in kernel is not met:

Missing some changes in [add_policy](https://github.com/strongswan/strongswan/blob/master/src/libcharon/plugins/kernel_netlink/kernel_netlink_ipsec.c#L2824) ?
Or is that the intention?

tobiasbrunner:
See [my update](https://wiki.strongswan.org/issues/2845#note-4) on the strongSwan issue tracker.

tobiasbrunner:
Support for XFRM interfaces was merged with c334cd9f62101a584b058d37863081ec690cd3de.

PR #123, opened 2019-01-21
Support reloading in strongswan.service

jeremyvisser:
strongSwan supports reloading config without disrupting connections via `ipsec reload`. Outside of strongSwan, the most popular service–agnostic way to reload a service is via `systemctl reload <foo>.service`.

By adding an `ExecReload=` line to the systemd unit file, we can trivially support this use case too.

### DIFF
--- a/init/systemd/strongswan.service.in
+++ b/init/systemd/strongswan.service.in
@@ -4,6 +4,7 @@ After=syslog.target network-online.target
 
 [Service]
 ExecStart=@SBINDIR@/@IPSEC_SCRIPT@ start --nofork
+ExecReload=@SBINDIR@/@IPSEC_SCRIPT@ reload
 StandardOutput=syslog
 Restart=on-abnormal
 


tobiasbrunner:
> strongSwan supports reloading config without disrupting connections via `ipsec reload`.

Unfortunately, that's not true. You can read up on why using `reload` as compared to `update` might not be a good idea (and why updating configs via stroke interface might affect existing connections either way) on our issue tracker. And `reload/update` also do not affect the loaded secrets, so other commands might be necessary too (maybe including some that flush caches).

Also, this is the unit file that crudely integrates the legacy config backend (starter/stroke) into systemd. Except for the `start/stop` commands the `ipsec` script is still very much required to control the daemon. And by doing so users can opt for using either of the reloading commands (plus whatever else is require to update the config/credentials).

For proper integration into systemd use the `charon-systemd` daemon with the vici/swanctl config backend and the associated `strongswan-swanctl` systemd unit. This backend also does not suffer from the reloading issue mentioned above.

The bottom line is I'm not going to apply this patch.

jeremyvisser:
Thank you for the detailed explanation. It makes sense, and I learnt something new today.

PR #124, opened 2019-01-29
ikev1: don't trigger updown events for redundant QMs

gollub:
With charon.delete_rekeyed=no redundant QMs might result
in an unenven number of updown calls. E.g. too many down
calls.

To solve this, the idea is update the state for redundant
Child SA to CHILD_REKEYED only if charon.delete_rekeyed is
enabled. This should also maintain the original behavior
as intended by commit a01eb5e4, to avoid another updown
hook trigger.

Additionally a redundant QM check is required for incoming QM
delete requests, to avoid again an unintended updown call.

Fixes #2902.

### DIFF
--- a/src/libcharon/sa/child_sa.c
+++ b/src/libcharon/sa/child_sa.c
@@ -1901,3 +1901,24 @@ child_sa_t * child_sa_create(host_t *me, host_t* other,
 	}
 	return &this->public;
 }
+
+/**
+ * Described in header.
+ */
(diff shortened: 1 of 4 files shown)


tobiasbrunner:
Discussed in [2902](https://wiki.strongswan.org/issues/2902) of the issue tracker, fixed with 451c2e7d582862f20362cb141d6a50fd7e52d7ac.

PR #126, opened 2019-02-07
Replace the CA in load_tester.

krinfels:
The previous one expired in December.
Also increase its size from 1024 to 2048 bits.

It was tested by simply running the plugin.


tobiasbrunner:
Thanks. I updated the CA certificate in the _load-tester-cert_ branch, but kept the key (it's 1024 bits for performance reasons).

PR #130, opened 2019-03-14
dhcp: Add release_on_delete parameter to allow not releasing an IP ad…

selsky:
…dress

Some DHCP servers will always grant us a different address after a DHCPRELEASE, and we may prefer to allow the server to expire leases (giving the client a good chance to receive the same address on next connection) rather than generating a DHCPRELEASE each time the IKE_SA goes down.

### DIFF
--- a/conf/plugins/dhcp.opt
+++ b/conf/plugins/dhcp.opt
@@ -12,6 +12,12 @@ charon.plugins.dhcp.identity_lease = no
 	Derive user-defined MAC address from hash of IKE identity and send client
 	identity DHCP option.
 
+charon.plugins.dhcp.release_on_delete = yes
+	Send DHCPRELEASE message to DHCP server when IKE_SA is deleted. When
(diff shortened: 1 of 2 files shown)


tobiasbrunner:
Is there an actual use case for this (if you want the same IP assigned to the client you could just use static leases) or is it more a workaround for the make-before-break reauthentication issue (IKE_SAs created overlapping by the same peer causing a DHCPRELEASE when the old IKE_SA is terminated while the new SA is still actually in use).

selsky:
Our use case is:
A user closes their laptop in one conference room, goes to another floor and reopens their laptop.  Their IPsec tunnel has closed.  We want their ssh session that was inside the tunnel to resume when they reopen the laptop and the tunnel is re-established.  In order for that to happen, they need to get the same IP address inside the tunnel, so we need to avoid the DHCPRELEASE being sent.

tobiasbrunner:
That's what MOBIKE is for (no need to close the IKE/IPsec SA).

PR #131, opened 2019-03-25
trap-manager: Wait for install to finish before uninstalling

srsmiraato:
There was a race condition between install() and uninstall()
where one thread was in the process of installing a trap
entry, and the child_sa destroyed previously, while the other
thread was uninstalling the same trap entry and ended up
trying to destroy the already destroyed child_sa, resulting
in a segmentation fault in the destroy_entry() function.

The uninstall() function needs to wait until all the threads
are done with the installing before proceeding to uninstall
a trap entry.

### DIFF
--- a/src/libcharon/sa/trap_manager.c
+++ b/src/libcharon/sa/trap_manager.c
@@ -352,6 +352,11 @@ METHOD(trap_manager_t, uninstall, bool,
 	entry_t *entry, *found = NULL;
 
 	this->lock->write_lock(this->lock);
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Please create a separate pull request for the configure option.

srsmiraato:
Removed the configure option commit and created a separate pull request for it.  Thanks.

tobiasbrunner:
Applied to master, thanks!

PR #132, opened 2019-03-26
configure: Add config option for external signal handling

srsmiraato:
By default, charon internally handles the SIGSEGV,
SIGILL, and SIGBUS signals raised by threads
(segv_handler). Make this configurable so that the
signal handling can be done either internally or
externally.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -320,6 +320,7 @@ ARG_ENABL_SET([leak-detective], [enable malloc hooks to find memory leaks.])
 ARG_ENABL_SET([lock-profiler],  [enable lock/mutex profiling code.])
 ARG_ENABL_SET([log-thread-ids], [use thread ID, if available, instead of an incremented value starting from 1, to identify threads.])
 ARG_ENABL_SET([monolithic],     [build monolithic version of libstrongswan that includes all enabled plugins. Similarly, the plugins of charon are assembled in libcharon.])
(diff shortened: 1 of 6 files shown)


tobiasbrunner:
Several issues with this:

1. Could you explain more about why this would be useful or how you use it exactly. 
2. I wonder if it really has to be a configure option or if this couldn't just be a compile option (i.e. just add something like `-DDISABLE_SIGNAL_HANDLER` to the CFLAGS to disable the internal signal handler for fatal signals), because I really don't see many users enabling this.
3. A lot more than just the `sigaction()` calls could be part of ifdefs (including the definition of `segv_handler`).  
4. If at all, it should probably be added to all charon derivatives (e.g. `charon-systemd` or `charon-nm`).

srsmiraato:
Hi Tobias, 

Our system has its own core file handler and we'd like our system to handle the core file of strongswan.  Without this change, one of the threads in the stacktrace would point to the segv_handler() function when a segmentation fault happens.  Unfortunately, we couldn't get any other useful information in the stacktrace.  By removing this internal handling, we were able to get more useful information from the stacktrace during the segmentation fault and allowed us to pinpoint the cause of the problem. 

I will get back to you regarding your next 3 concerns, as I will try them and test it on our system.

Thanks for the feedback.

srsmiraato:
The new commit addresses the next 3 concerns.  I've changed it from configure option to compile option.  Charon derivatives affected are charon-nm, charon-cmd, charon-tkm and charon-systemd.

Thanks

tobiasbrunner:
Looks good, thanks. Applied to master with some minor modifications.

PR #133, opened 2019-04-03
wolfSSL plugin for cryptographic implementations.

SparkiDev:


SparkiDev:
wolfSSL needs to be installed and configured with the following options:
  C_EXTRA_FLAGS="-DWOLFSSL_PUBLIC_MP" --enable-keygen --enable-rsapss

Other algorithms not compiled by default can be enabled with:
  --enable-aesccm
  --enable-aesctr
  --enable-des3
  C_EXTRA_FLAGS="-DWOLFSSL_DES_ECB"
  --enable-ed25519
  --enalbe-camellia

Sean

tobiasbrunner:
Thanks a lot, Sean.

I haven't looked at it in detail, but noticed the GPL headers. Would you consider (re-)submitting the plugin under the MIT X11 license (see [Contributions](https://wiki.strongswan.org/projects/strongswan/wiki/Contributions) on our wiki for details)?

And the two commits could be squashed.

SparkiDev:
Sure, I can change the license for you.
I read the Contributions page and it said it was fine for plugins to be GPL and other cryptographic plugins are also GPL so I stuck with that.
I’ll squash all the commits.
I would have liked to have a Curve25519 test but you don't optionally compile in tests.
So, I’ll leave the test out.

Sean
—
Sean Parkinson
sean@wolfssl.com
wolfSSL Inc
https://github.com/wolfssl
http://twitter.com/wolfSSL

tobiasbrunner:
> Sure, I can change the license for you.

Awesome, thanks.

> I read the Contributions page and it said it was fine for plugins to be GPL and other cryptographic plugins are also GPL so I stuck with that.

Most were written by us, so we have all the rights anyway. But larger contributions e.g. in the _openssl_ plugin were under the MIT X11 license as is, for instance, the complete _botan_ plugin.

> I would have liked to have a Curve25519 test but you don't optionally compile in tests.

There are test vectors for it in the _test-vectors_ plugin (see [here](https://github.com/strongswan/strongswan/blob/f43302dc10020b721ad07aa7aeaf5161e95b70d7/src/libstrongswan/plugins/test_vectors/test_vectors/curve25519.c)).

SparkiDev:
The curve25519 test was based on the test vectors.
I didn’t want to keep turning off and on the configuration to run the test vectors it but it doesn’t matter.

I’ve updated the license to the version on the website.
Let me know if you have issues with installing/configuring wolfSSL.

Thanks,
Sean
—
Sean Parkinson
sean@wolfssl.com
wolfSSL Inc
https://github.com/wolfssl
http://twitter.com/wolfSSL

SparkiDev:
Hi tobiasbrunner,
I've looked through your branch. First, thanks for all the changes you made. I copied from other plugins and was wary of changing their patterns and the code is better.
Thanks for fixing the p and q generation in RSA. The fingerprint code was an issue I came back to many times - thanks for fixing it.
I can't see the issue you had with Ed25519.

RNG_WEAK uses a global with locking as this may be faster than creating a new random number generator and tearing it down again. I don't know the contexts of use.

get_keysize() uses wc_RsaEncryptSize() as this is the public API. I would rather not use mp APIs as they are meant to be internal use only for shared libraries.

wc_curve25519_shared_secret_ex() does only return 0 on success - my mistake. I got it right up the top!

MAX_ENCODED_SIG_SZ is the size of the encrypted signature. MAX_DER_DIGEST_SZ is what I should have used.
wolfSSL doesn't have defines for encoded key sizes. The APIs for encoding don't allow querying of the size either. This is because it is designed for embedded platforms where static buffers will be used. So I used constants that I didn't explain. Please change the constants RSA constants to reflect that they are the ASN.1 encoding lengths.
For ECC signature length I should have used: wc_ecc_sig_size(). ECC public key DER encoding '+ 30' needs to be: MAX_SEQ_SZ + 2 * MAX_ALGO_SZ + MAX_SEQ_SZ + TRAILING_ZERO. (Though this is excessive)
For Ed25519 the '+ 40' should be '+ MAX_SEQ_SZ + MAX_ALGO_SZ + MAX_SEQ_SZ + TRAILING_ZERO'.
Let me know if that doesn't work.

The hardening should be on by default. Seeing this error indicates that options.h is not being included before including settings.h. Where was the error coming up?

You are correct. FIPS mode is either built in and on or not built-in and off. The only change I saw was to the debug output and they make sense.
 
Thanks,
Sean, wolfSSL

tobiasbrunner:
Thanks for the feedback.

> I copied from other plugins and was wary of changing their patterns and the code is better.

Yeah, I noticed that. Some of the code, e.g. in the OpenSSL plugin, is quite old and would look a bit differently today.

> The fingerprint code was an issue I came back to many times - thanks for fixing it.

No problem, the lack of unit tests didn't make that easier.

> I can't see the issue you had with Ed25519.

The problem is that after calling `wc_Ed25519PrivateKeyDecode()` the public key is not set if only the private key was encoded, which is e.g. the case in our Ed25519 unit test. That's why I extracted the code you added after `wc_ed25519_import_private_only()` into a separate function that is now called after both possible paths.

> RNG_WEAK uses a global with locking as this may be faster than creating a new random number generator and tearing it down again. I don't know the contexts of use.

`RNG_WEAK` is used all over the place (for nonces, IVs etc.), way more often than any other type actually (in particular if DH secrets are allocated with plugin-internal RNGs). Some instances are rather short-lived (e.g. to generate padding for a single IKE message), others longer (e.g. during complete IKE exchanges) and one, in particular, is used to allocate IKE SPIs and is kept around for the whole running time of the daemon. The amount of requested random data is usually small, though (32-byte nonces are probably the largest blocks). So I guess it comes down to whether the overhead for the allocation of RNGs outweighs all of these instances having to compete for the same mutex. How expensive is `wc_InitRng()`?

> get_keysize() uses wc_RsaEncryptSize() as this is the public API. I would rather not use mp APIs as they are meant to be internal use only for shared libraries.

I noticed you used it in `verify_emsa_pss_signature()` and it would theoretically be more accurate (e.g. 2047 bit keys have been seen in the wild). I guess in practice it doesn't matter much, so we can also leave it as is.

> MAX_ENCODED_SIG_SZ is the size of the encrypted signature. MAX_DER_DIGEST_SZ is what I should have used.

That was actually my change (took it from the example in the apidoc for `wc_EncodeSignature()`). You originally had `dgst.len + 20` :) But I see that `MAX_DER_DIGEST_SZ` is more accurate here.

> Please change the RSA constants to reflect that they are the ASN.1 encoding lengths.

I did that similarly to the stuff below.

> For ECC signature length I should have used: wc_ecc_sig_size(). ECC public key DER encoding '+ 30' needs to be: MAX_SEQ_SZ + 2 * MAX_ALGO_SZ + MAX_SEQ_SZ + TRAILING_ZERO. (Though this is excessive)

Hehe, yeah I guess, but at least it's also more expressive.  I modified Ed25519 similarly. Also noticed that the buffer for the ECDSA private key was way too large (and the length was not adjusted, added some tests for this too).

> The hardening should be on by default. Seeing this error indicates that options.h is not being included before including settings.h. Where was the error coming up?

When running our configure script.  The header check with `AC_CHECK_HEADER` for `wolfssl/ssl.h` fails when building with `-Werror`. However, as you noted, it does work if `#include <wolfssl/options.h>` is explicitly added to the check. I guess alternatively we could just look for `wolfssl/options.h`?

SparkiDev:
OK, I see the fix for Ed25519.

The cost of wc_initRng varies depending on build. If it is built to use Intel's RD_RAND *and* the chip supports it then the init is cheap and the generate is stateless. If not, then using HashDRBG performs health tests and locking is needed for a shared random.

If the size of the key to bits is important then by all means use the mp API.

I understand the problem with hardening now. A user can either use configure to generate options.h or construct their own user_settings.h. I think that using configure is the most logical for this product. In that case, checking for options.h makes more sense.

Sean

tobiasbrunner:
> The cost of wc_initRng varies depending on build. If it is built to use Intel's RD_RAND _and_ the chip supports it then the init is cheap and the generate is stateless. If not, then using HashDRBG performs health tests and locking is needed for a shared random.

I see. Let's keep it as it is for now, we can always change it later if there are issues.

>  In that case, checking for options.h makes more sense.

OK, changed that.  In case the file doesn't exist, it's always possible to override the check by passing `ac_cv_header_wolfssl_options_h=yes` to the configure script.

PR #134, opened 2019-04-08
ike-sa-manager: Introduce charon.ike_spi_label[_mask]

sladkani:
This allows one to configure a fixed "label" to be set into some bits
(as defined by charon.ike_spi_label_mask) of IKE SPI value generated for
new IKE SAs.

### DIFF
--- a/conf/options/charon.opt
+++ b/conf/options/charon.opt
@@ -160,6 +160,13 @@ charon.ignore_acquire_ts = no
 	config will	be sent. This always happens for IKEv1 connections as the
 	protocol only supports one set of traffic selectors per CHILD_SA.
 
+charon.ike_spi_label = 0x0
+	Fixed SPI "label" bits to be set to SPIs chosen for new IKE SAs.
+
+charon.ike_spi_label_mask = 0x0
+	The bitmask applied to ike_spi_label when set to SPIs chosen for new
+	IKE SAs. Default is 0x0, meaning, no label is assigned to new IKE SA SPIs.
+
 charon.ikesa_limit = 0
 	Maximum number of IKE_SAs that can be established at the same time before
 	new connection attempts are blocked.
--- a/src/libcharon/sa/ike_sa_manager.c
+++ b/src/libcharon/sa/ike_sa_manager.c
@@ -42,6 +42,10 @@
 /* the default number of segments (MUST be a power of 2) */
 #define DEFAULT_SEGMENT_COUNT 1
 
+/* the default SPI label/mask fed to spi_get on new IKE SAs */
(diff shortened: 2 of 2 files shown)


tobiasbrunner:
That's actually what `spi_cb_t` and `set_spi_cb()` is for. It allows a plugin to generate IKE SPIs however it pleases.

ebirger:
Hi Tobias,

Thanks for the quick review!

- I think this functionality would be useful  regardless of the spi allocation mechanism
- we wanted to avoid the overhead of maintaining a full plugin for this specific functionality which is one LoC essentially

tobiasbrunner:
> I think this functionality would be useful regardless of the spi allocation mechanism

What exactly is the use case?

> we wanted to avoid the overhead of maintaining a full plugin for this specific functionality

There is not that much overhead (you can easily create out-of-tree plugins, see e.g. [redirect-plugin](https://github.com/tobiasbrunner/strongswan-redirect-plugin)).

ebirger:
The use case is being able to partition the spi space to support HA deployments.
Similar to the way the ESP spi space can be sharded.
In such case, the spi ‘label’ is an orchestrated property which is unrelated to the allocation scheme.

tobiasbrunner:
>  In such case, the spi ‘label’ is an orchestrated property which is unrelated to the allocation scheme.

I don't fully agree with that, it changes the composition and range of the SPI values and if that doesn't match the mask/label this will not work (IKE SPIs technically don't have to be random, just locally unique, although, in charon it's generally assumed the local SPI is random).

Anyway, to confirm the point of my initial comment, have a look at [tkm_spi_generator.c](https://github.com/strongswan/strongswan/blob/f5ad3cf4919e7689cab8c50d0582d38c70c50bf7/src/charon-tkm/src/tkm/tkm_spi_generator.c), which is basically an implementation of this feature via `spi_cb_t`. It's not really specific to TKM, though, it was just requested by the customer that commissioned _charon-tkm_, and deemed of little use for others, so it was integrated there. I guess if this is useful to other people, we could move that code to the IKE_SA manager, which is what I did in the _gh134-spi-label_ branch.

sladkani:
I've tested gh134-spi-label, looks good.

ebirger:
I see your point with regards to the fact that the SPI range must be synchronized between the generator and the labeling mechanism, so they cannot be completely independent. So I guess it's up to the administrator to make sure the configuration is sane.

Thanks!

tobiasbrunner:
> I've tested gh134-spi-label, looks good.

Thanks for testing. I've pushed the commit to master.

PR #135, opened 2019-04-25
Update proposal_substructure.c

suishixingkong:
fix a wrong in proposal_substructure_create_from_proposal_v2


tobiasbrunner:
Thanks, applied to master. Luckily, the type is not really used when generating payloads and the encoding rules for IKEv1 and IKEv2 are basically the same for the substructure (I guess we would have noticed this mistake a long time ago if that wasn't the case).

PR #140, opened 2019-05-20
fix some wrong

suishixingkong:
fix a wrong in certs_filter. When traversing a certificate, if you specify a public key type but do not specify  id, a certificate with a different public key type is returned.


tobiasbrunner:
Please start writing proper commit messages and use separate branches/pull requests for separate fixes.

PR #141, opened 2019-05-21
fix an error

suishixingkong:
If the key type is specified but the id is null, it is possible to return a certificate that does not match the parameter key type.


tobiasbrunner:
You should really learn more about Git (branching, writing descriptive commit messages). Anyway, I pushed the fixes to master (for simplicity in one commit and with modified message). Thanks.

PR #142, opened 2019-05-23
Update credential_manager.c

suishixingkong:
Fix memory leaks


tobiasbrunner:
That's not a leak. `get_fingerprint()` returns a globally cached `chunk_t` (see `cred_encoding_t` and the public key implementations).

PR #143, opened 2019-05-27
flex: Fix parallel shadow build

orgads:
On some builds, the source file is built before the header is created/copied to the shadow directory.

### DIFF
--- a/src/libstrongswan/settings/settings_lexer.l
+++ b/src/libstrongswan/settings/settings_lexer.l
@@ -16,7 +16,7 @@
 
 #include <utils/parser_helper.h>
 
-#include "settings_parser.h"
+#include <settings/settings_parser.h>
 
 bool settings_parser_open_next_file(parser_helper_t *ctx);
 
--- a/src/starter/parser/lexer.l
+++ b/src/starter/parser/lexer.l
@@ -17,7 +17,7 @@
 #include <utils/parser_helper.h>
 #include <parser/conf_parser.h>
 
-#include "parser.h"
+#include <parser/parser.h>
 
 bool conf_parser_open_next_file(parser_helper_t *ctx);
 


tobiasbrunner:
What is a "shadow" build or the "shadow" directory? And how/why exactly does this patch fix it?

orgads:
Shadow build is a build that is done out-of-source.

```
mkdir build
cd build
../strongswan/configure
make
```

When building from the tar distribution without preserving its timestamps (in our case, we imported the tar into another git repository), there seems to be a timing/dependency issue in the Makefiles. Sometimes flex/bison are running and sometimes they don't. If you build in parallel, then the lexer source is sometimes built before the parser's header is created in the target directory.

`Makefile.am` in libstrongswan has:
```
AM_CPPFLAGS = \
        -I$(top_srcdir)/src/libstrongswan \
```

and there's a similar entry in starter.

but the nested `libstrongswan/settings` or `starter/parser` directories are not in the include path, so if the header is not created yet, `#include "settings_parser.h"` fails.

I'm still not sure this fixes the issue. We'll need some time to see if it still reproduces.

fmartinsons:
I confirmed the same problem (pretty random but definitely due to parallel build), we do the same fix as in the PR.

tobiasbrunner:
OK, closing this.

If this is a problem for you, try building in two steps. Run `make distdir` first to build the generated source/header files (gperf, bison/flex etc.) and then run `make [install]` (and optionally delete the generated distribution directory).

orgads:
Until now, it seems to fix the problem for us.

Can you explain why do you reject it? Even if there's a workaround, why not accept something that works, and doesn't seem to break anything?

tobiasbrunner:
> Can you explain why do you reject it?

You just wrote above that it doesn't fix the issue.

tobiasbrunner:
Wait, that was another guy? What are you people doing?

orgads:
I wrote that I'm *still not sure* it fixes the issue, because it reproduces randomly, so we need some time to test it.

And @fmartinsons wrote that he uses the same fix, and it works for him.

orgads:
Regarding out-of-tree, do you build from the tar file or from git?

It might be caused by timestamps of generated c/h files, for which there are missing dependencies in the Makefile, but tracking that down is harder than the fix I proposed here.

tobiasbrunner:
As I wrote, the Git repository. The point of including these files in the tarball is precisely to avoid having to build them (i.e. to avoid dependencies on gperf, bison/flex etc. for regular users). And why import a tarball into a Git repository if you could just work with Git (each release is tagged)? Can you reproduce the issue if you do a (clean) out-of-tree build from the (clean) Git repository? If so, does the patch help (again, using clean dirs)?

fmartinsons:
Personnaly, we use git repository but I didn't test to clean it (remove the generated file) before doing a make install. We came up with the fix (workaround) suggested by @orgads and live with it for almost a year without any noticed issues.
Note: We use a yocto (https://www.yoctoproject.org/)  build environment to cross compil on different targets (our build machines has 24 or 32 vcpu so 24 or 32 possible parallel make targets but we experienced the problem not so often)

@tobiasbrunner did you think the suggested fix (workaround) will break something for users which doesn't have any issue with pre-generated files ? If so, fine by me, I can cleaning the the pre-generated files before a 'from scratch' build (I think it is even better in a cross compil environment where we try to use the less possible host binaries)

tobiasbrunner:
I'd really like to understand what's going on and why this particular patch fixes it. It basically causes the compiler to search the header not relative to the compilation unit (first), but instead searches it in the with `-I` configured include directories. Thanks to autotools, which adds additional relative and absolute paths via `-I`, the latter searches in both the build and the source directory. The former, though, will only be successful in the build directory where the generated files of the parser/lexer are located, because while the other directories are searched, the file won't be found there without the `settings/` prefix.

So my current guess is that the patch could only possibly make a difference if there are multiple versions of the header file around, i.e. one in the source directory and one in the out-of-tree build directory (the latter maybe created delayed, in which case the file in the source directory is found and used).  That's why I asked if this still worked from a clean repository (i.e. where there is no second header file, which is usually the case if there never was a build directly in the source directory or it isn't an unpacked tarball).

To answer your question, no I don't think it should break anything for regular users. But I also don't see it fix anything unless two versions of the header file are around, which isn't automatically the case and which could theoretically cause problems if the parser ever changes and the header file in the source tree is outdated.

As mentioned above, you can safely pre-generate the files in the build directory via `make distdir` (the header file should theoretically be generated first anyway, as it is listed in `BUILT_SOURCES`, but I guess there could be race conditions, which running this command definitely avoids).

PR #144, opened 2019-06-18
line 832 server was not defined

hbrv-wd:
gateway = identification_create_from_string(server);
in this line server is not defined so I change it to gateway but I do not know it is true or not 
and in line 749 foce_encap was wrong I changed that to force_encap but I am not sure it is true or not
Help me please.


tobiasbrunner:
Thanks for the report. I've fixed these issues in master.

PR #145, opened 2019-06-19
Remove the minimum character limitation for pre-shared keys

tnoell:
This removes the minimum character limitation for pre-shared keys.
Reason: (taken from Augustin Rivero https://bugs.launchpad.net/ubuntu/+source/network-manager-strongswan/+bug/1697536):
This a client configuration not a server configuration. In a server you could enforce the user not to use an insecure password. But is it insecure to configure the client wrongly? It's like not allowing a login prompt to enter a short password. Security must be implemented when setting the password in the server, not when logging in!

A client shouldn't impose restrictions in configuration, otherwise it's not a generic client, it's just a client that works in some cases.

Other people having this problem:
https://ubuntuforums.org/showthread.php?t=2388118
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=896086

### DIFF
--- a/src/frontends/gnome/auth-dialog/main.c
+++ b/src/frontends/gnome/auth-dialog/main.c
@@ -113,7 +113,6 @@ static gboolean get_secrets(const char *type, const char *uuid, const char *name
 	NMAVpnPasswordDialog *dialog;
 	char *prompt, *pw = NULL;
 	const char *new_pw = NULL;
-	guint32 minlen = 0;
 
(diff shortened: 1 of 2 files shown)


Thermi:
Kindly requesting feedback from @tobiasbrunner

tobiasbrunner:
See the [discussion here](https://lists.strongswan.org/pipermail//dev/2019-August/002007.html) (you actually linked back here from there).

PR #146, opened 2019-06-25
libtpmtss: Fix concurrency issue in libtpmtss.

krinfels:
When a private key from TPM is "loaded" strongswan
creates a context structure used for communication
with the TSS.
When multiple tunnels are established at the same time
and using the same private key, it is possible to make
concurrent calls to the TSS with the same context
which results in a multiple threads writing in the same
place in memory and undefined behaviour.
Fix this by protecting calls to the TSS with a mutex
unique for every private key.


strongX509:
The pull request has been included with one major modification (protect a Tss2_Sys_HashSequence with a single lock over several Tss2_Sys calls) in the strongSwan 5.8.1dr1 release.

PR #147, opened 2019-06-28
libipsec can work on configuring auto=route

suishixingkong:
libipsec can work on configuring auto=route

### DIFF
--- a/src/libipsec/Makefile.am
+++ b/src/libipsec/Makefile.am
@@ -14,8 +14,7 @@ ipsec_sa.c ipsec_sa.h \
 ipsec_sa_mgr.c ipsec_sa_mgr.h
 
 libipsec_la_LIBADD = \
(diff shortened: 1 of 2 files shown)


tobiasbrunner:
There are several issues with this:

1. There must not be a dependency on libcharon (look at how expires are treated, actually, I had an old implementation of the relaying of an acquire event in a private branch, which I now pushed to the _libipsec-trap_ branch).
2. Even though duplicate acquires are ignored by the trap manager, ideally, it should be avoided somehow that each further non-matching packet triggers another one (maybe it would even be possible to queue the packets, up to a limit perhaps, so they could be sent once the SA is established).
3. The traffic selectors in the event are not those from the policy, but derived from the packet that matched the policy (i.e. they are very specific, single IPs and including ports and protocol).
4. Trying to check out the SA again right away is unlikely to be successful (at least in real world scenarios).

Anyway, if this works for you, fine, but we can't accept it upstream like this.

PR #148, opened 2019-07-19
Android 9 Support + Fix Crash

raviyer:
Moved to Target version 28.
https://developer.android.com/about/versions/pie/android-9.0-migration
Added FOREGROUND_SERVICE permission to manifest file.

Moved to AndroidX from legacy support libraries.
https://developer.android.com/jetpack/androidx/migrate
Fixed some deprecation warnings associated with this. Only addressed deprection
warnings that do not affect minSdkVersion.

Fixed crash in TextInputLayoutHelper


raviyer:
@tobiasbrunner Can you please review and approve these changes? W/o it does not currently run on Android 9.

tobiasbrunner:
Thanks for the information. I've addressed these issues in the _android-updates_ branch. What's not currently fixed is the deprecation of `ConnectivityManager.CONNECTIVITY_ACTION` (probably requires using network callbacks on newer API levels), I'll look into that at a later date.

PR #150, opened 2019-09-16
Move D-Bus conf file to $(datadir)/dbus-1/system.d

worldofpeace:
Since D-Bus 1.9.18 configuration files installed by third-party should
go in share/dbus-1/system.d. The old location is for sysadmin overrides.


tobiasbrunner:
Thanks. I guess this change isn't compatible with older D-Bus releases. However, since our NM plugin targets newer NM versions (I think >= 1.1), distros shipping that will probably also ship a compatible D-Bus version. So that shouldn't be a problem.

PR #152, opened 2019-10-04
install libvici.h with dev headers

danc86:
This installs libvici.h as `/usr/include/strongswan/vici/libvici.h` so that other programs can build against it.

### DIFF
--- a/src/libcharon/plugins/vici/Makefile.am
+++ b/src/libcharon/plugins/vici/Makefile.am
@@ -47,6 +47,12 @@ libvici_la_SOURCES = \
 libvici_la_LIBADD = $(top_builddir)/src/libstrongswan/libstrongswan.la
 
 
+if USE_DEV_HEADERS
+vici_includedir = ${dev_headers}/vici
+vici_include_HEADERS = libvici.h
+endif
+
+
 TESTS = vici_tests
 
 check_PROGRAMS = $(TESTS)


tobiasbrunner:
As we wrote before (e.g. on the dev mailing list [here](https://lists.strongswan.org/pipermail/dev/2018-October/001939.html), and [here](https://lists.strongswan.org/pipermail/dev/2015-April/001308.html)), libvici is only intended for internal use. The [davici library](https://github.com/strongswan/davici) provides several advantages over it (less restrictive license, independent of any strongSwan libraries, asynchronous API). And as we stated before, `--with-dev-headers` shouldn't be used in general at all, because we don't provide a stable API for our libraries.

danc86:
Okay. You might want to update the wording on the [Vici](https://wiki.strongswan.org/projects/strongswan/wiki/Vici) wiki page, particularly where it says:

> libvici provides a stable low-level C API to exchange messages using the VICI protocol. While it depends on libstrongswan, it uses a stable, coding-style neutral API and should work for many applications.

I took that to mean that third-party applications are encouraged to build against libvici. But it sounds like you are saying you consider libvici an internal implementation detail and you want third parties to build against davici instead.

tobiasbrunner:
Yes, we should probably clarify that. You are obviously still free to use it, just copy or reference the header directly.

PR #153, opened 2019-10-04
vici: Use unique names for CHILD_SAs in the child-updown event too

fxkr:
The unique names were introduced for the list-sas command in commit
04c0219e55d9338b6492548c073189bfd3d5431b. Unfortunately
the child-updown command wasn't updated to match.

The original name is already being returned in the "name" attribute,
so it'll still be available.

Example:

    >>> import vici, json
    >>> s = vici.Session()

    # First, for comparison, the list-sas command:
    >>> print(json.dumps(list(s.list_sas()), sort_keys=True, indent=4, separators=(',', ': ')))
    [
        {
            "vti0": {
                "child-sas": {
                    "vti0-1": {
                        ...

    # A child-updown event before the change:
    >>> for x in s.listen(["child-updown"]): print(json.dumps(x, sort_keys=True, indent=4, separators=(',', ': ')))
    [
        "child-updown",
        {
            "vti0": {
                "child-sas": {
                    "vti0": {   # <-- wrong: inconsistent with list-sas
                        ...

    # A child-updown event after the change:
    >>> s = vici.Session()
    >>> for x in s.listen(["child-updown"]): print(json.dumps(x, sort_keys=True, indent=4, separators=(',', ': ')))
    [
        "child-updown",
        {
            "vti0": {
                "child-sas": {
                    "vti0-1": {  # <-- fixed

### DIFF
--- a/src/libcharon/plugins/vici/vici_query.c
+++ b/src/libcharon/plugins/vici/vici_query.c
@@ -1716,6 +1716,7 @@ METHOD(listener_t, child_updown, bool,
 {
 	vici_builder_t *b;
 	time_t now;
+	char buf[BUF_LEN];
 
 	if (!this->dispatcher->has_event_listeners(this->dispatcher, "child-updown"))
 	{
@@ -1734,7 +1735,11 @@ METHOD(listener_t, child_updown, bool,
 	list_ike(this, b, ike_sa, now);
 	b->begin_section(b, "child-sas");
 
-	b->begin_section(b, child_sa->get_name(child_sa));
+	snprintf(buf, sizeof(buf), "%s-%u",
+	         child_sa->get_name(child_sa),
+	         child_sa->get_unique_id(child_sa));
+
+	b->begin_section(b, buf);
 	list_child(this, b, child_sa, now);
 	b->end_section(b);
 


tobiasbrunner:
> Unfortunately the child-updown command wasn't updated to match.

Because it's not necessary there. The _child-updown_ event is always triggered for a single CHILD_SA so there is no need to use a unique name for the section.

fxkr:
It is actually a problem for API users that want to track all child SA's that StrongSwan has (for health monitoring purposes - this is a real-world use btw). There can be multiple child SA's for an SA, and currently you cannot use the key to distinguish them.

Additionally, to monitor in a race condition free way, you need to:

1. subscribe to child-updown
2. call list_sas once to dump the initial list
3. track the results from both the commands as they come in. So you have to correlate ID's of list-sas and child-updown with each other. Right now they are inconsistent.

tobiasbrunner:
> and currently you cannot use the key to distinguish them.

Which you shouldn't, the section name in _list-sas_ is only "unique" to make sending multiple CHILD_SAs with the same name possible at all (_child-rekey_, for instance, doesn't use the name in the section name at all). It's not intended to be used as key. Use the _name_ and _uniqueid_ attributes (or only the latter) for that if you need it.

fxkr:
It contradicts the [documentation](https://www.strongswan.org/apidoc/md_src_libcharon_plugins_vici_README.html). It literally says "same data".

image

And besides, no interest in fixing a glaring inconsistency? Is there any underlying reason for shooting this PR down that I'm missing? I just don't get it.

fxkr:
(Anyway, as with the other PR, I respect your decision whatever it is in the end. Thanks for maintaining a great open source project for all of us! For the third and larger PR I wanted to do I'll ask on the mailing list if there's any chance of it getting in before I waste too much time on it.)

PR #154, opened 2019-10-04
HACKING: add libgmp3-dev to dependencies list, because the default config requires it

fxkr:
Steps to repro:

    Run an AWS EC2 (free tier is fine) with Amazon's Ubuntu 18.04 AMI.

    git clone ...
    sudo apt-get update
    sudo apt-get install automake autoconf libtool pkg-config gettext perl python flex bison
    ./autogen.sh
    ./configure

Error:

    configure: error: GNU Multi Precision library gmp not found

### DIFF
--- a/HACKING
+++ b/HACKING
@@ -20,6 +20,7 @@ the code, you need the following tools:
     - lex/flex
     - yacc/bison
     - gperf
+    - libgmp3-dev
 
     Optionally:
     - lcov/genhtml


tobiasbrunner:
It's not a dependency that is **required** to build from the repository, the others are.

fxkr:
Right now, if you follow the exact steps in HACKING:

```
For interested developers, we have a public repository. To check out and compile
the code, you need the following tools:

    - Git
    - a recent GNU C compiler (>= 3.x)
    - automake
    - autoconf
    - libtool
    - pkg-config
    - gettext
    - perl
    - python
    - lex/flex
    - yacc/bison
    - gperf

    Optionally:
    - lcov/genhtml
    - Doxygen

To check out the master branch, use:

    git clone git://git.strongswan.org/strongswan.git

or using HTTP:

    git clone https://git.strongswan.org/strongswan.git

After a successful check out, give the autotools a try:

    cd strongswan/
    ./autogen.sh

Then you're in, start the build as usual:

    ./configure [options]
    make
    make install
```

the build will fail with the error I gave.

So the documentation as it is right now is incorrect.

Are you sure you do not want a HACKING file that will actually give you a working build?

Would it be an acceptable compromise to add `- libgmp3-dev (needed if using default ./configure options, but optional in general)`?

tobiasbrunner:
> Are you sure you do not want a HACKING file that will actually give you a working build?

I've no problem with that. HACKING is not intended for regular users but for devs.

fxkr:
I thought it's to ramp up *new* developers, who can surely benefit from a complete/correct example. Why even have docs if they're wrong? I'm disappointed - I really cannot understand why you are rejecting this -, but I respect your decision. Thanks for HACKING on StrongSwan anyway.

tobiasbrunner:
The whole point of the configure script is doing these dependency checks, so I really don't see a problem if there is an error that depends on the _variable_ configure options (also see [this page](https://wiki.strongswan.org/projects/strongswan/wiki/InstallationDocumentation#Building-strongSwan) on the wiki). However, for most of the dependencies listed in HACKING there isn't a hard check in the script because they are not required when building from a tarball. For instance, _autotools_ and _gettext_ are needed to build the script itself, and the script won't fail if _perl_, _python_, _flex/bison_ and _gperf_ are not found because the generated files are part of the tarball, yet a build from the repository will later fail without them.

fxkr:
Thanks for the explanation.


The next dev will also figure out that if they only have the listed dependencies, the options aren't optional.

To be fair, `HACKING` does mention the `[options]`, and `./configure` does tell what's missing, so it's no big deal.

PR #155, opened 2019-10-14
kernel_pfkey: Pass ESN flag to kernel if ESN is enabled

semihalf-duda-patryk:
Current behaviour is that ESN can be negotiated even when kernel doesn't
support ESN. Pfkey plugin have ESN flag provided but this information is
not propagated to kernel.

This patch adds passing ESN flag to kernel when ESN was negotiated and
appropriate flag is present in kernel headers.

Signed-off-by: Patryk Duda <pdk@semihalf.com>

### DIFF
--- a/src/libcharon/plugins/kernel_pfkey/kernel_pfkey_ipsec.c
+++ b/src/libcharon/plugins/kernel_pfkey/kernel_pfkey_ipsec.c
@@ -1766,6 +1766,14 @@ METHOD(kernel_ipsec_t, add_sa, status_t,
 	}
 	PFKEY_EXT_ADD(msg, sa);
 
+#ifdef SADB_X_SAFLAGS_ESN
+	if (id->proto != IPPROTO_COMP && data->esn)
+	{
+		DBG2(DBG_KNL, "  using extended sequence numbers (ESN)");
+		sa->sadb_sa_flags |= SADB_X_SAFLAGS_ESN;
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
What platform supports this? FreeBSD and macOS don't as far as I can tell, and the Linux kernel neither (the PF_KEY interface is deprecated there, though). I saw it mentioned in OpenBSD's version of `pfkeyv2.h`, is that where you use it?

semihalf-duda-patryk:
I'm preparing patches for FreeBSD kernel which adds support for ESN. Strongswan doesn't support OpenBSD, so it was tested against FreeBSD only.

tobiasbrunner:
I see, great. I pushed a modified version of this to the _gh155-pfkey-esn_ branch. Let me know if that works for you.

semihalf-duda-patryk:
I have tested both scenarios:

Strongswan compiled when SADB_X_SAFLAGS_ESN is not present in pfkeyv2.h and running with ESN enabed in configuration. Strongswan correctly reported that ESN is not supported by kernel.

Second scenario was compiling it with SADB_X_SAFLAGS_ESN present in pfkeyv2.h and running with ESN enabled in configuration. Tunnel was established correctly, so for me it works perfectly.

tobiasbrunner:
OK, great. Thanks for testing. Applied to master.

PR #156, opened 2019-10-18
Fix for issue with blinding not defined

dgarske:


tobiasbrunner:
Thanks, applied to master.

PR #157, opened 2019-10-23
Support the deprecated sha-1WithRSASignature 1.3.14.3.2.29. Needed to…

ratcashdev:
… avoid "unable to parse signature algorithm" when parsing an X.509 AZURE generated VPN gateway certificate.

### DIFF
--- a/src/libstrongswan/credentials/keys/public_key.c
+++ b/src/libstrongswan/credentials/keys/public_key.c
@@ -126,6 +126,7 @@ signature_scheme_t signature_scheme_from_oid(int oid)
 		case OID_MD5:
 			return SIGN_RSA_EMSA_PKCS1_MD5;
 		case OID_SHA1_WITH_RSA:
+		case OID_SHA1_WITH_RSA_OIW:
 		case OID_SHA1:
 			return SIGN_RSA_EMSA_PKCS1_SHA1;
 		case OID_SHA224_WITH_RSA:


tobiasbrunner:
Could you please provide a bit more information on this? Who/what tool created the certificate with this signature algorithm identifier? Was there an option to use a different identifier? Was there an option to use SHA-256 (SHA-1 shouldn't be used anymore)?

ratcashdev:
Certificates are generated in windows, using:
```
"C:\Program Files (x86)\Windows Kits\10\bin\10.0.17763.0\x86\makecert" -sky exchange -r -n "CN=Azure-ZONE-DMZ-VPN" -pe -a sha1 -len 2048 -ss My "Azure-ZONE-DMZ-VPN.cer"
```

tobiasbrunner:
Does the tool support `-a sha256`? If yes, use it. If no, you might want to consider using a proper tool to create certificates.

> The person using these arguments stated that some sources recommended using sha1 for LINUX...

I don't see why that would be the case (in particular if the tool encodes the signature with a legacy OID that nobody else uses).

ratcashdev:
Yes, I am also trying to push through using openssl for certificate generation.
Regardless of that, what do you think of this PR?
It took me quite some time to track down the quite meaningless "unable to parse signature algorithm" message.
If your objective is to drive away ppl from using these signatures, please use a more friendly message, like "The signature is using a deprecated algorithm (OID HERE) that should not be used and is not supported by this tool".
Alternatively, simple merge this PR...

tobiasbrunner:
> If your objective is to drive away ppl from using these signatures

Absolutely. But adding an extra error message for this rare issue (you are the first one to report this and I'm sure Microsoft's tool never did this correctly) is not worth the effort, especially because nobody nowadays uses SHA-1 to issue certificates.

ratcashdev:
@tobiasbrunner it's not about the signature, but the OID. OpenSSL still does accept it, so I am wondering why strongswan shouldn't.

tobiasbrunner:
> it's not about the signature, but the OID.

Yes, the OID of the signature algortihm.

> OpenSSL still does accept it, so I am wondering why strongswan shouldn't.

So? I don't see why we should add support for an obsolete identifier ([since at least 21 years](https://tools.ietf.org/html/rfc2437)) for an obsolete algorithm (moving away from SHA-1 has been recommended for about 10 years now, and web browsers, for instance, stopped supporting SHA-1 for certificate signatures altogether nearly three years ago).

PR #160, opened 2019-11-06
testing: fix sha description in esp-alg-null

zhangkaiheb:
change the sha1 to sha256 according the ipsec.conf file.


tobiasbrunner:
Thanks, applied to master (fixed the first one also for the IKEv1 version of the scenario).

PR #162, opened 2020-01-11
Support running tests through setuptools

Duffyx:
Could after this PR it be possible to update the PyPi package of this module to also include the test folder and Python files in that folder? I saw that it was not included in the sources fetched from PyPi.
This PR helps in porting this module to FreeBSD, as FreeBSD does not use PyPi for modules but the ports tree.


Duffyx:
Also, the module seems to work fine with higher Python version, so I also added them.

tobiasbrunner:
`setup.py` is created by the Makefile (it sets the version number based on the configure script). Unfortunately, your proposed changes [are deprecated](https://github.com/pytest-dev/pytest-runner#deprecation-notice). So other than the version updates I don't see us applying these changes upstream.

Duffyx:
Oh yeah I see now. To be honest changing the supported versions and making the PyPi package **include the test dir and its files** in its src ball would already help a lot. :-). Closing this PR as it doesn't provide any value except the version changes. I don't think a new PR makes sense as you are already informed about the supported python versions.

tobiasbrunner:
FYI, I pushed some changes to the _vici-python_ branch. Let me know what you think.

Duffyx:
Looking good! Looking forward to the updated package in PyPi & thanks for the rapid response.

tobiasbrunner:
Merged and [released](https://pypi.org/project/vici/5.8.2/).

PR #166, opened 2020-03-25
charon-nm: Allow fixed values for charon-nm.port and charon-nm.port_nat_t

rkunze:
This change allows to optionally set `charon-nm.port` and/or `charon-nm.port_nat_t` to a fixed value, while retaining the default of using a random port.

My use case for this is a roadwarrior VPN connection over IPv6 (with MOBIKE enabled), where the client sits behind a customer grade router with a packet filter firewall (Fritz!Box in my case, but should not really matter). 

In this scenario, initiating the VPN works because the firewall detects the outgoing packets to ports 500/4500 on the server and allows answer packets. Likewise, ESP works because the firewall detects an active connection initiated by the client side. But when the server tries to rekey the child SA, the firewall has already timed out the initial UDP connection, and the CREATE_CHILD_SA request fails.

The simple solution here is to configure the firewall to allow incoming packets to the client on the UDP port(s) used by Strongswan, but the random ports used by charon-nm make this impractical.

### DIFF
--- a/src/charon-nm/charon-nm.c
+++ b/src/charon-nm/charon-nm.c
@@ -196,9 +196,11 @@ int main(int argc, char *argv[])
 							   "charon-nm.syslog.daemon.default", 1));
 	charon->load_loggers(charon);
 
-	/* use random ports to avoid conflicts with regular charon */
-	lib->settings->set_int(lib->settings, "charon-nm.port", 0);
-	lib->settings->set_int(lib->settings, "charon-nm.port_nat_t", 0);
+	/* use random ports by default to avoid conflicts with regular charon */
+	lib->settings->set_int(lib->settings, "charon-nm.port",
+		lib->settings->get_int(lib->settings, "charon-nm.port", 0));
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks. I pushed a simpler change to the _gh166-nm-ports_ branch.

> But when the server tries to rekey the child SA, the firewall has already timed out the initial UDP connection, and the CREATE_CHILD_SA request fails.

I guess an alternative would be to force sending NAT keep-alives even if no NAT is detected. That might prevent the firewall from blocking inbound traffic. I suppose you could also send regular DPDs from the server to achieve a similar effect (the NM client currently does not send any DPDs).

rkunze:
> Thanks. I pushed a simpler change to the gh166-nm-ports branch.

Great, thank you!

> I guess an alternative would be to force sending NAT keep-alives even if no NAT is detected. 

Probably, yes. But setting a static port was easier to figure out for me (as a complete Strongswan newbie, I've only ever used it until now).

PR #167, opened 2020-04-10
src/Makefile.am: fix build with netlink but without libcharon

ffontaine:
xfrmi has been moved to a separate directory since version 5.8.0 and https://github.com/strongswan/strongswan/commit/d74ddd78937506436ba1ec8ebfbbd8a2cdaf0cc0

However, build of xfrmi will fail without libcharon:
`make[4]: *** No rule to make target '../../src/libcharon/plugins/kernel_netlink/libstrongswan-kernel-netlink.la', needed by 'xfrmi'.  Stop.`

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -145,5 +145,7 @@ if USE_TPM
 endif
 
 if USE_KERNEL_NETLINK
+if USE_LIBCHARON
   SUBDIRS += xfrmi
+endif
 endif


tobiasbrunner:
We can probably just move the `if USE_KERNEL_NETLINK` block into the existing `if USE_LIBCHARON` block (see _gh167-xfrmi-libcharon_ branch).

May I ask, what exactly do you build if you build without _libcharon_ (the core component of the strongSwan project)?

ffontaine:
I'm trying to bump the strongswan that we have on buildroot: https://git.buildroot.net/buildroot/tree/package/strongswan and I stumbled on this issue. We're still on 5.6.3 which is getting old. We have backported some CVE patches but having an up to date package is always better.

Since the very beginning, we have an option to allow our users to enable/disable charon as allowed by your configure script. However, I don't know if many of our users enable strongswan without charon.

I fully agree that your proposal in gh167-xfrmi-libcharon branch is better than mine.

tobiasbrunner:
Thanks for the context.

> Since the very beginning, we have an option to allow our users to enable/disable charon as allowed by your configure script. However, I don't know if many of our users enable strongswan without charon.

_charon_ is only one user of _libcharon_ (basically a small wrapper around it). Others are e.g. _charon-systemd_, _charon-cmd_ or _charon-nm_, which is why you can build without _charon_ daemon just fine and e.g. use _charon-systemd_ instead. But it generally makes not much sense to build strongSwan without any user of _libcharon_. The fix still makes sense, of course.

PR #168, opened 2020-04-28
pkcs11-plugin: Add support for tokens which don't support hashing with signature

mskalski:
Some PKCS#11 libraries do not support combined hashing and signature creation.

For now only RSA signature is performed (hashing is done using external hasher i.e. openssl or gnutls), as virtually all PKCS#11 libraries support this kind of operation.

Added option in `strongswan.conf`: `pkcs11.use_sign_hasher` for enabling old behaviour - using signature mechanism with hashing done using PKCS#11 library.

Also added mechanisms with SHA-224.

### DIFF
--- a/conf/plugins/pkcs11.opt
+++ b/conf/plugins/pkcs11.opt
@@ -30,3 +30,6 @@ charon.plugins.pkcs11.use_pubkey = no
 
 charon.plugins.pkcs11.use_rng = no
 	Whether the PKCS#11 modules should be used as RNG.
+	
+charon.plugins.pkcs11.use_sign_hasher = no
+	Whether the PKCS#11 modules should use signature mechanisms with hashing.
--- a/src/libstrongswan/plugins/pkcs11/pkcs11.h
+++ b/src/libstrongswan/plugins/pkcs11/pkcs11.h
@@ -512,6 +512,8 @@ typedef unsigned long ck_mechanism_type_t;
 #define CKM_SHA256_RSA_PKCS_PSS		(0x43)
 #define CKM_SHA384_RSA_PKCS_PSS		(0x44)
 #define CKM_SHA512_RSA_PKCS_PSS		(0x45)
+#define CKM_SHA224_RSA_PKCS		(0x46)
+#define CKM_SHA224_RSA_PKCS_PSS		(0x47)
 #define CKM_RC2_KEY_GEN			(0x100)
 #define CKM_RC2_ECB			(0x101)
 #define	CKM_RC2_CBC			(0x102)
--- a/src/libstrongswan/plugins/pkcs11/pkcs11_private_key.c
+++ b/src/libstrongswan/plugins/pkcs11/pkcs11_private_key.c
@@ -117,6 +117,9 @@ METHOD(private_key_t, get_keysize, int,
 	return this->pubkey->get_keysize(this->pubkey);
 }
 
+
+#define HASH_NOT_USED (hash_algorithm_t)(-1)
+
 /**
  * See header.
  */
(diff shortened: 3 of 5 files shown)


tobiasbrunner:
Thanks.

> Some PKCS#11 libraries do not support combined hashing and signature creation.

You mean the tokens don't support these mechanisms and neither does the middleware/libraries for them provide them in software so PKCS#11 clients don't have to bother? Could you provide any details on what kind of tokens/libraries you are referring to? (I'm just wondering because this never has been an issue in all the years the _pkcs11_ plugin existed.)

> Added option in `strongswan.conf`: `pkcs11.use_sign_hasher` for enabling old behaviour - using signature mechanism with hashing done using PKCS#11 library.

The default shouldn't change. However, I don't think a config option is necessary for this. Instead we should query the token's supported mechanisms and fall back to software hashing automatically. I pushed such a change to the _gh168-pkcs11-sig-hash_ branch.

> Also added mechanisms with SHA-224.

I don't think that's necessary, I doubt anybody has been missing it (using at least SHA-256 should be the default nowadays for signatures).

mskalski:
> Thanks.
> 
> > Some PKCS#11 libraries do not support combined hashing and signature creation.
> 
> You mean the tokens don't support these mechanisms and neither does the middleware/libraries for them provide them in software so PKCS#11 clients don't have to bother? Could you provide any details on what kind of tokens/libraries you are referring to? (I'm just wondering because this never has been an issue in all the years the _pkcs11_ plugin existed.)

Yes, I mean PKCS#11 middleware, as pkcs11-plugin does not know about underlying hardware.
I have some smartcards (particularly named ENCARD) we use internally in my company, and PKCS#11 library does not support hashing at all. As you can see this patch was made over 3 years ago for use pkcs11 plugin with these cards. I think I might had send this patch few years ago to strongswan-users list, but it was missed.

In fact in was not so uncommon for other PKCS#11 libraries I worked with to have no support for hashing, as they give mainly support for digital signature performed by card. On the other hand I did not find any PKCS#11 library which does hashing on card, so in my opinion there is no point to use mechanisms with hashing, especially when strongswan have great hasher plugins.
 

> > Added option in `strongswan.conf`: `pkcs11.use_sign_hasher` for enabling old behaviour - using signature mechanism with hashing done using PKCS#11 library.
> 
> The default shouldn't change. However, I don't think a config option is necessary for this. Instead we should query the token's supported mechanisms and fall back to software hashing automatically. I pushed such a change to the _gh168-pkcs11-sig-hash_ branch.

Yes it could be done like that, the only drawback I can see is slowing signature a little bit, as it makes extra calls into PKCS#11 library and implementations vary. Sometimes querying supported mechanisms makes access to underlying hardware and smartcards are generally slow.

But your `gh168-pkcs11-sig-hash` branch is something I had in my mind when tried to make pkcs11 plugin to work with our smartcards. My only concern is that creating every time mechanism_enumerator for checking if every next algorithm on table is supported is suboptimal from the point of view calls into PKCS#11 library. But it is compact end elegant :-)


> 
> > Also added mechanisms with SHA-224.
> 
> I don't think that's necessary, I doubt anybody has been missing it (using at least SHA-256 should be the default nowadays for signatures).

Yes, you're right, it was for completness, but implementation is almost at no cost at all and SHA-224 is available in strongswan hasher plugins.

mskalski:
I'll check if it your `gh168-pkcs11-sig-hash` branch is working with our cards, I don't expect it doesn't.

tobiasbrunner:
>  On the other hand I did not find any PKCS#11 library which does hashing on card, so in my opinion there is no point to use mechanisms with hashing, especially when strongswan have great hasher plugins.

There definitely are cards that support hashing on-card (at least according to the OpenSC sources). But I've little practical experience with this. And if the card or middleware supports hashing it allows removing some of the plugins. Also, because `CKM_RSA_PKCS` is likely also supported if mechanisms with hashing are, checking in this order seemed more logical to me (in particular because we'd have to check if a hasher can be instantiated if we first selected `CKM_RSA_PKCS`  and then check mechanisms with hashing as fallback, which would complicate the code quite a bit).

> Sometimes querying supported mechanisms makes access to underlying hardware and smartcards are generally slow.

I've only looked at OpenSC, which doesn't do that (the list of mechanisms is created once during initialization of the card).

> My only concern is that creating every time mechanism_enumerator for checking if every next algorithm on table is supported is suboptimal from the point of view calls into PKCS#11 library.

At most two such calls are required per signature, and since creating the signature via PKCS#11 is relatively slow anyway the overhead might not be that significant.

mskalski:
> > On the other hand I did not find any PKCS#11 library which does hashing on card, so in my opinion there is no point to use mechanisms with hashing, especially when strongswan have great hasher plugins.
> 
> There definitely are cards that support hashing on-card (at least according to the OpenSC sources). But I've little practical experience with this. And if the card or middleware supports hashing it allows removing some of the plugins. Also, because `CKM_RSA_PKCS` is likely also supported if mechanisms with hashing are, checking in this order seemed more logical to me (in particular because we'd have to check if a hasher can be instantiated if we first selected `CKM_RSA_PKCS` and then check mechanisms with hashing as fallback, which would complicate the code quite a bit).

Yes, I know most of today cards support hashing, but in my observations no PKCS#11 library use hashing on card. 

> 
> > Sometimes querying supported mechanisms makes access to underlying hardware and smartcards are generally slow.
> 
> I've only looked at OpenSC, which doesn't do that (the list of mechanisms is created once during initialization of the card).

Problem is not with mechanism caching or not, but usually every call into PKCS#11 library with arguments  of type CK_SLOT_ID or CK_SESSION_HANDLE should involve checking if token (card) is still present and was not replaced in the meantime. So this is the case where calling PKCS#11 functions can be slow, especially for hashing large data input. But it depends on many factors i.e. smartcard reader speed, and of course PKCS#11 implementation details.

> > My only concern is that creating every time mechanism_enumerator for checking if every next algorithm on table is supported is suboptimal from the point of view calls into PKCS#11 library.
> 
> At most two such calls are required per signature, and since creating the signature via PKCS#11 is relatively slow anyway the overhead might not be that significant.

OK, seems legit.

mskalski:
As your branch `gh168-pkcs11-sig-hash` (0a67ab5e0b401d34de8d68675a3df4701e933222) works well with our smartcards, I think this PR can be closed.

Thanks again :-)

tobiasbrunner:
Thanks for testing. I pushed the change to master.

PR #170, opened 2020-05-10
Makefile.am: store appdata to /usr/share/metainfo

ConiKost:
The path '/usr/share/appdata' is deprecated and
should be changed to '/usr/share/metainfo'.

See section: 2.1.2. Filesystem locations
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

Signed-off-by: Conrad Kostecki <conrad@kostecki.com>


tobiasbrunner:
Thanks. I've pushed a commit to that effect to master (73b60338dc8a95273fbdf387915a5572183fe478).

PR #171, opened 2020-05-19
ike keying retry while redirected: Re-initiate with either the redirected peer or with the configured remote

sladkani:
Currently, if an initiator has been redirected, and it fails to reach the redirected peer during IKE_CONNECTING (and keying_tries is set), then re-initiation begins directly to the IP of the redirected peer (ignoring the original remote peer configured).

There might be situations when an admin would like to enforce the re-initiation to the the originally configured remote (which is the entity responsible arbitration using redirection) - for example, a different or more appropriate redirect answer might be returned.

Add a flag which controls the behavior of keying_tries: whether to re-initiate with the redirected peer (default), or with the configured remote.

swanctl.conf:
  connections.<conn>.keyingtry_redirected
ipsec.conf Conn sestion:
  keyingtry_redirected

---

This really seems an anomaly -  as in other situations (e.g. DPD failure *after* being established), re-initiation is performed with the originally configured remote.

If so, we can consider eliminating this new 'conn' flag and just fix the IKE_CONNECTING retry code as suggested.


tobiasbrunner:
> This really seems an anomaly - as in other situations (e.g. DPD failure _after_ being established), re-initiation is performed with the originally configured remote.

Although, reestablishing (which creates a new IKE_SA object from scratch) can't be fully compared to restarting the current IKE_SA, I agree. We should retry connecting to the original host. I pushed a fix to the _gh171-redirect-reset_ branch (theoretically, `remote_host` could initially have been set to a value, which we can't restore when resetting, but that's generally only the case with mobile IPv6, so not sure if that's really an issue). Let me know if that works for you.

sladkani:
> I pushed a fix to the _gh171-redirect-reset_ branch 

I saw you augmented `reset()` to perform the clearing of `this->remote_host` in the redirected case.

Alas there is one case which I'm concerned with (not tested though):

 				case INVALID_KE_PAYLOAD:
				{
				        ... 
					DBG1(DBG_IKE, "peer didn't accept DH group %N, "
						 "it requested %N", diffie_hellman_group_names,
						 bad_group, diffie_hellman_group_names, this->dh_group);

					if (this->old_sa == NULL)
					{	/* reset the IKE_SA if we are not rekeying */
						this->ike_sa->reset(this->ike_sa, FALSE);
					}

In case (after redirect) responder returned the initiator an "invalide ke", the initiator should probably continue with the redirect peer (and not start with the original remote).

I'm not sure I'm reading this properly though. What does `this->old_sa == NULL` mean?
If I'm correct, please consider passing a flag to `reset()`.

sladkani:
> In case (after redirect) responder returned the initiator an "invalide ke", the initiator should probably continue with the redirect peer (and not start with the original remote).
> 
> If I'm correct, please consider passing a flag to `reset()`.

This may also be the case for processing a COOKIE notification.  Restart with *new* redirected peer.

tobiasbrunner:
> This may also be the case for processing a COOKIE notification. Restart with _new_ redirected peer.

Good catch, thanks. I think we can use the `new_spi` flag as that is not set when retrying due to notifies. I updated the commit.

sladkani:
> > This may also be the case for processing a COOKIE notification. Restart with _new_ redirected peer.
> 
> Good catch, thanks. I think we can use the `new_spi` flag as that is not set when retrying due to notifies. I updated the commit.

Thanks Tobias.

Tested. Confirmed this works as expected in the "redirected-to a peer which is not responding, let's re-initiate with the original peer" case.

tobiasbrunner:
Thanks for testing. Pushed to master.

PR #172, opened 2020-05-20
Remove a certificate installed by a vici authority on unload

raviyer:
vici.load-authority installed certificates were not removed when
the authority was unloaded. The underlying mem_cred has no support
to remove certs. This change adds this support to mem_cred, vici_cred
and removes and destroys the certificate if required when the authority
is unloaded

### DIFF
--- a/src/libcharon/plugins/vici/vici_authority.c
+++ b/src/libcharon/plugins/vici/vici_authority.c
@@ -500,6 +500,17 @@ CALLBACK(unload_authority, vici_message_t*,
 	{
 		if (streq(authority->name, authority_name))
 		{
+			if (authority->cert)
+			{
+				certificate_t *c =
+					this->cred->remove_cert(this->cred,
+								authority->cert);
+				if (c != authority->cert)
+				{
+					DESTROY_IF(c);
+				}
+
+			}
 			this->authorities->remove_at(this->authorities, enumerator);
 			authority_destroy(authority);
 			found = TRUE;
--- a/src/libcharon/plugins/vici/vici_cred.c
+++ b/src/libcharon/plugins/vici/vici_cred.c
@@ -589,6 +589,12 @@ METHOD(vici_cred_t, add_cert, certificate_t*,
 	return this->creds->add_cert_ref(this->creds, TRUE, cert);
 }
 
+METHOD(vici_cred_t, remove_cert, certificate_t*,
+	private_vici_cred_t *this, certificate_t *cert)
+{
+	return this->creds->remove_cert(this->creds, cert);
+}
+
 METHOD(vici_cred_t, destroy, void,
 	private_vici_cred_t *this)
 {
@@ -618,6 +624,7 @@ vici_cred_t *vici_cred_create(vici_dispatcher_t *dispatcher)
 				.cache_cert = (void*)_cache_cert,
 			},
 			.add_cert = _add_cert,
+			.remove_cert = _remove_cert,
 			.destroy = _destroy,
 		},
 		.dispatcher = dispatcher,
--- a/src/libcharon/plugins/vici/vici_cred.h
+++ b/src/libcharon/plugins/vici/vici_cred.h
@@ -49,6 +49,13 @@ struct vici_cred_t {
 	certificate_t* (*add_cert)(vici_cred_t *this, certificate_t *cert);
 
 	/**
+	 * Remove a certificate from the certificate store
+	 *
+	 * @param cert	certificate to be removed from the store
+	 * @return		The removed certificate
+	 */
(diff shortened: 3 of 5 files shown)


tobiasbrunner:
This doesn't seem ideal. If for some reason two authority sections reference the same CA certificate, removing one section would also remove the CA certificate of the other from the credential set. While I suppose that's not very common, it will also happen if an authority section is renamed and the new one is loaded before the old one is removed (that's how e.g. `swanctl` operates). And the same certificate might also be loaded directly via `load-cert()` command e.g. via `swanctl` if it is stored in `x509ca`, so removing the authority section, which might onlybe used to add URIs to a CA and not to load the certificate from somewhere else, would then also remove that CA certificate completely from the credential set.

The latter actually reveals a potential issue with the existing implementation, where the certificates associated with authority sections are stored in `vici_cred_t`. If CA certificates in authority sections are not stored in the same location as those loaded via `load-cert()`, the `clear-creds()` command is problematic as it will also clear those CA certificates loaded exclusively with authority sections, which are then not added again to the credential set. Only if they are re-added via `load-cert()`, e.g. from swanctl's `x509ca` directory, or the authority sections are reloaded, would the certificates still be available afterwards. I think the credential set is primarily shared to save memory and to avoid enumerating the same CA certificate multiple times (duplicate certificates are currently not filtered, so e.g. duplicate certificate requests might be sent - see e.g. [issue 2560](https://wiki.strongswan.org/issues/2560) on our issue tracker).

What exactly is your use case here? Do you load CA certificates only via `load-authority()` or also via `load-cert()`?

Anyway, I pushed a fix for the existing `clear-creds()` conflict to the _gh172-vici-authority_ branch. If CA certificates are only loaded via authority section, this also "unloads" the CA certificate of an unloaded authority section. It also includes the patch to avoid duplicate certificate requests I mentioned in the issue linked above.

If it would be preferable to avoid the potential memory overhead and enumerating duplicate CA certificates from the two vici credential sets altogether, a slightly different approach would be to store all CA certificates in `vici_authority_t` with reference counting and tracking if they were loaded with an authority section. Then we could keep only one reference per CA certificate, no matter from where it was loaded. Since there usually are not that many authority sections, this might be overkill, though.

tobiasbrunner:
> a slightly different approach would be to store all CA certificates in `vici_authority_t` with reference counting and tracking if they were loaded with an authority section.

I pushed such a change to the _gh172-vici-ca-certs_ branch (includes the changes of the other branch except the one regarding certificate requests as generally no duplicate CA certificates will be enumerated here).

raviyer:
Thanks for the detailed reply.
> This doesn't seem ideal. If for some reason two authority sections reference the same CA certificate, removing one section would also remove the CA certificate of the other from the credential set. While I suppose that's not very common, it will also happen if an authority section is renamed and the new one is loaded before the old one is removed (that's how e.g. `swanctl` operates). And the same certificate might also be loaded directly via `load-cert()` command e.g. via `swanctl` if it is stored in `x509ca`, so removing the authority section, which might onlybe used to add URIs to a CA and not to load the certificate from somewhere else, would then also remove that CA certificate completely from the credential set.
Agreed I will look into these scenarios.


> 
> What exactly is your use case here? Do you load CA certificates only via `load-authority()` or also via `load-cert()`?
We only load via load-authority (programmatically using vici APIs and unload-authority when we want to not trust the cert any longer.

> 
> Anyway, I pushed a fix for the existing `clear-creds()` conflict to the _gh172-vici-authority_ branch. If CA certificates are only loaded via authority section, this also "unloads" the CA certificate of an unloaded authority section. It also includes the patch to avoid duplicate certificate requests I mentioned in the issue linked above.
Thanks will look at it.

> 
> If it would be preferable to avoid the potential memory overhead and enumerating duplicate CA certificates from the two vici credential sets altogether, a slightly different approach would be to store all CA certificates in `vici_authority_t` with reference counting and tracking if they were loaded with an authority section. Then we could keep only one reference per CA certificate, no matter from where it was loaded. Since there usually are not that many authority sections, this might be overkill, though.
Sounds good I may look into doing that.

raviyer:
> > a slightly different approach would be to store all CA certificates in `vici_authority_t` with reference counting and tracking if they were loaded with an authority section.
> 
> I pushed such a change to the _gh172-vici-ca-certs_ branch (includes the changes of the other branch except the one regarding certificate requests as generally no duplicate CA certificates will be enumerated here).

Thanks again. I will see the change and also study to see how ref-counting can be employed

raviyer:
Another thought is that since there is no unload-cert and clear-creds also removes all authorities, how about simply fail a load-cert or load-authority if the certificate already exists?

PR #173, opened 2020-05-24
charon-nm: allow configurable remote traffic selectors

heldchen:
this change allows to customize the previously hard-coded remote traffic selectors. it corresponds to ipsec.conf's `rightsubnet` and charon-cmd's `--remote-ts` parameters, but keeps the comfort of being able to toggle the connection through gnome's gui. 

this does not actually write the newly added "remote-ts" configuration node into the NetworkManager configuration file, but will use an existing value. exposing the config setting in the gui could be done in an amended or followup PR if this is a desired change.

use case: remote firewall appliance wrongly accepts the `0.0.0.0/0` but does not actually route external traffic, leaving the user with a partially working internet connection.

### DIFF
--- a/src/charon-nm/nm/nm_service.c
+++ b/src/charon-nm/nm/nm_service.c
@@ -839,10 +839,24 @@ static gboolean connect_(NMVpnServicePlugin *plugin, NMConnection *connection,
 	}
 	ts = traffic_selector_create_dynamic(0, 0, 65535);
 	child_cfg->add_traffic_selector(child_cfg, TRUE, ts);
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks, looks good. I've pushed it with some error handling and whitespace/commit message changes to the _gh173-nm-remote-ts_ branch. Let me know if that works for you.

heldchen:
thanks tobias for your code improvements, appreciated! works well for me - looking forward to see this in a future release version :-)

tobiasbrunner:
Applied to master.

PR #176, opened 2020-07-02
VPP support 5.8.2

choppsv1:
Add support for VPP (https://wiki.fd.io/view/VPP)

Tested with VPP versions 19.08 (LTS), 20.01 and 20.05 (latest release)


tobiasbrunner:
Not sure when I'll have time to review this. Would you consider submitting those plugins under the MIT X11 license (see [Contributions](https://wiki.strongswan.org/projects/strongswan/wiki/Contributions))?

choppsv1:
I didn't originate the changes (but significantly modified them to finish/update them to newer VPP versions). From code forensics (presence of "netlink" and "winsock" in a couple comments in the initial changes/commits) I gathered the original author copied the kernel netlink files and socket winsock files as templates. Given that fact, I copied the copyright/licenses from the corresponding files from those plugins, adding his company and then mine to the top.

Given the original copied files were already GPLd I'm not sure it makes sense to add an MIT license on top of that, as they are already encumbered with the GPL, right?  It's all API goo anyway, just trying to help out by contributing the changes back. There has been interest expressed on the VPP lists for these at least.

Thanks,
Chris.

tobiasbrunner:
> I didn't originate the changes (but significantly modified them to finish/update them to newer VPP versions).

Yeah, I saw that there were other authors. That there were no headers/copyrights for those changes may make this problematic anyway (but IANAL). Did you contact them before this PR? It might be a good idea to contact them and ask how they (or their employer) think about this.

> Given the original copied files were already GPLd I'm not sure it makes sense to add an MIT license on top of that, as they are already encumbered with the GPL, right?

No, we (or rather our university) have the rights to most of the codebase (as described on the page I linked above), which allows  us to also license it commercially (which is a considerable income stream). So it's always preferable for us if external contributions are made under the MIT X11 license, so we can offer that code to corporate customers. If this plugin was GPL only, such a customer could only use it and the whole codebase, including any other plugins they might have developed, under the GPL.

choppsv1:
Yes, I've sent mail (as far as I could) to the authors. The primary no longer works for PANTHEON.tech, according to their linkedin page. I also sent an email to Kyle Mestery's email as indicated on github, who appears to still work at cisco.

The only reason I felt comfortable with taking this route is that it was derived from GPL'd files so there was no new license being granted and only carrying forward the one required by derivative works. I would not feel comfortable with doing anything else w/o the original authors consent. So I guess this will need to stall here until we hear from them or their corporate representatives.

Thanks,
Chris.

PR #177, opened 2020-07-03
Create charon plugin that adds and removes proxy ARP entries

sddj:
The FARP plugin uses linux specific packet filters and so is not compatible with macos. This plugin adds and removes proxy ARP entries to the host routing as VPN connections go up and down.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -278,6 +278,7 @@ ARG_ENABL_SET([forecast],       [enable forecast plugin forwarding broadcast/mul
 ARG_ENABL_SET([duplicheck],     [advanced duplicate checking plugin using liveness checks.])
 ARG_ENABL_SET([error-notify],   [enable error notification plugin.])
 ARG_ENABL_SET([farp],           [enable ARP faking plugin that responds to ARP requests to peers virtual IP])
+ARG_ENABL_SET([proxyarp],       [enable add a proxy ARP entry for the peers virtual IP])
 ARG_ENABL_SET([ha],             [enable high availability cluster plugin.])
 ARG_ENABL_SET([led],            [enable plugin to control LEDs on IKEv2 activity using the Linux kernel LED subsystem.])
 ARG_ENABL_SET([load-tester],    [enable load testing plugin for IKEv2 daemon.])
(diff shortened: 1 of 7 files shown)


tobiasbrunner:
Thanks for your efforts, but what exactly is the advantage of this plugin over simply calling `arp -s <vip> <ether addr or auto> pub` from a [vici](https://wiki.strongswan.org/projects/strongswan/wiki/Vici) or [updown](https://wiki.strongswan.org/projects/strongswan/wiki/updown) script? Seems like a lot of code to replicate that command. (Note that due to its dynamic nature, the _farp_ plugin has a different scope, it can e.g. proxy ARP for multiple IPs, ranges or even complete remote subnets.) Also, what's the point of the Linux specific code when we can simply use the _farp_ plugin there?

sddj:
Thanks for the review.  You make really good points.  I only added the Linux specific code to get past the CI checks. It was a good exercise for me to learn more about how to build a plugin and get it to work in StrongSwan.  Just thought it might be useful to try to contribute.  Should I close the pull request?

tobiasbrunner:
>  I only added the Linux specific code to get past the CI checks.

Ah, I see. You could have disabled it for those builds and enabled it for the FreeBSD/macOS builds (like we do with e.g. the _kernel-pfroute_ plugin, see [scripts/test.sh](https://github.com/strongswan/strongswan/blob/master/scripts/test.sh)).

>  Just thought it might be useful to try to contribute. Should I close the pull request?

Let me assure you that that's really appreciated.  But I've currently no plans to adopt this plugin. A solution using `arp(8)` and a script could perhaps be documented on our wiki. For instance, on [forwarding and split-tunneling](https://wiki.strongswan.org/projects/strongswan/wiki/ForwardingAndSplitTunneling) or perhaps just on the page for the [farp plugin](https://wiki.strongswan.org/projects/strongswan/wiki/Farpplugin) that's already linked there, or simply as a comment to [issue 3498](https://wiki.strongswan.org/issues/3498) (similar to the snippets in [issue 1008](https://wiki.strongswan.org/issues/1008) for proxy NDP), which could then be linked from the farp plugin page until we manage to port the plugin to BSD-based systems (a note about its incompatibility is missing there anyway).

sddj:
Thanks again for the review.

PR #179, opened 2020-07-15
SN and OID

socketpair:
https://tools.ietf.org/html/rfc4519#section-2.31


src/libstrongswan/utils/identification.c
```diff
@@ -67,8 +67,7 @@ static const x501rdn_t x501rdns[] = {
 	{"UID", 				OID_PILOT_USERID,			ASN1_PRINTABLESTRING},
 	{"DC", 					OID_PILOT_DOMAIN_COMPONENT, ASN1_PRINTABLESTRING},
 	{"CN",					OID_COMMON_NAME,			ASN1_PRINTABLESTRING},
-	{"S", 					OID_SURNAME,				ASN1_PRINTABLESTRING},
-	{"SN", 					OID_SERIAL_NUMBER,			ASN1_PRINTABLESTRING},
+	{"SN", 					OID_SURNAME,				ASN1_PRINTABLESTRING},
```
socketpair, line 70: https://github.com/strongswan/strongswan/commit/a8c09d8cc0044fbff6c89c2e8966427ee7ed1ef0#diff-019ea69dc5521c8a94784efcbc8478c7R72

The bug was introduced 14 years ago !

tobiasbrunner:
I did some digging. This mapping has its origins in the X.509 patch for FreeS/WAN on which parts of strongSwan are based, which was started back in 2000. It was aligned with how OpenSSL did this in earlier versions. However, there it was changed already in March of 2002 (see https://github.com/openssl/openssl/commit/ffbe98b7630d604263cfb1118c67ca2617a8e222) to make it compatible with RFC 2256 (the predecessor of RFC 4519). Unfortunately, that change apparently went unnoticed by any of us or our users so it never made it into our code base.

We'll probably include this change in one of our next releases, I've pushed it to the _gh179-sn-rdn_ branch for now (changed the commit message a bit and added a fix for our testing environment).

PR #180, opened 2020-07-24
Send CISCO FLexVPN vendor ID

Thermi:
Send CISCO FLexVPN vendor ID if send_cisco_flexvpn_vendor_id is set to "yes".
Option description:
Send the CISCO FlexVPN vendor ID. It is required in order to make CISCO brand devices allow negotiating a local_ts (from strongSwan's point of view) that is not the assigned "virtual" IP address, if a "virtual" IP address is requested by strongSwan. Sending the CISCO FlexVPN vendor ID makes the CISCO peer allow this peer to negotiate a TS of 0.0.0.0/0 == 0.0.0.0/0 instead (unknow if it works for IPv6 too). This has been tested with a "tunnel mode ipsec ipv4" CISCO template but should also work for GRE encapsulation.

### DIFF
--- a/conf/options/charon.opt
+++ b/conf/options/charon.opt
@@ -392,6 +392,15 @@ charon.routing_table_prio
 charon.rsa_pss = no
 	Whether to use RSA with PSS padding instead of PKCS#1 padding by default.
 
+charon.send_cisco_flexvpn_vendor_id = no
+	Send the CISCO FlexVPN vendor ID. It is required in order to make CISCO
+	brand devicesallow negotiating a local_ts (from strongSwan's point of view)
+	that is not the assigned "virtual" IP address, if a "virtual" IP address is
+	requested by strongSwan. Sending the CISCO FlexVPN vendor ID makes the CISCO
+	peer allow this peer to negotiate a TS of 0.0.0.0/0 == 0.0.0.0/0 instead
+	(unknow if it works for IPv6 too). This has been tested with a "tunnel mode
+	ipsec ipv4" CISCO template but should also work for GRE encapsulation.
+
 charon.send_delay = 0
 	Delay in ms for sending packets, to simulate larger RTT.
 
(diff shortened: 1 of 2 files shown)


Thermi:
Check failures are unrelated to the PR.

tobiasbrunner:
Thanks, Noel. I've pushed a modified version of your patch to the _cisco-flexvpn-vendor-id_ branch (renamed the option to match that for Cisco Unity and modified description and commit message).

Do you know of any documentation of this feature/vendor ID from Cisco? While I found some documentation of FlexVPN, this particular vendor ID is never mentioned.

Thermi:
Thank you for pushing it to that branch.

Other than [that](https://quickview.cloudapps.cisco.com/quickview/bug/CSCvv05031), which I just found, I have no documentation.
The behaviour was discovered by observing behaviour and checking against the sent vendor IDs.

tobiasbrunner:
> Other than [that](https://quickview.cloudapps.cisco.com/quickview/bug/CSCvv05031), which I just found, I have no documentation.

Thanks. Interesting, the workaround (only visible after logging in) actually refers to strongSwan and sending the VID payload to fake support for FlexVPN.

johnfzc:
That bug was was opened in response to a support ticket I created with Cisco about the issue, you can use this [public link](https://quickview.cloudapps.cisco.com/quickview/bug/CSCvv05031) for a more cursory description of the issue without a Cisco support account. It's a request for an improvement that would allow a Cisco side configuration to permit clients to use the 0.0.0.0/0 tunnel selector instead of forcing the remote side selector to use of the assigned IP when the VID is not received.

There is no other public documentation of this behaviour that I am aware of; there was another bug but I can't locate it. I believe the support team identified the cause of the issue from the IOS source code. I'll update the issue if I can obtain any other public information about it.

tobiasbrunner:
Thanks for the update.

PR #181, opened 2020-09-23
Remove obsolete setting regarding the Standard Output

Mingli-Yu:
The Standard output type "syslog" is obsolete, causing a warning since systemd
version 246 [1].

Please consider using "journal" or "journal+console"

[1] https://github.com/systemd/systemd/blob/master/NEWS#L202

Signed-off-by: Mingli Yu <mingli.yu@windriver.com>

### DIFF
--- a/init/systemd-starter/strongswan-starter.service.in
+++ b/init/systemd-starter/strongswan-starter.service.in
@@ -4,7 +4,6 @@ After=syslog.target network-online.target
 
 [Service]
 ExecStart=@SBINDIR@/@IPSEC_SCRIPT@ start --nofork
-StandardOutput=syslog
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
If your system uses `systemd`, please consider using [charon-systemd](https://wiki.strongswan.org/projects/strongswan/wiki/charon-systemd), which logs directly to the journal, and the corresponding `strongswan.service` unit, not the legacy starter/charon combination.

I guess by removing the setting this will most likely default to `journal` (via `DefaultStandardOutput`), which apparently was the fallback now. So setting it explicitly does not seem necessary. I pushed the commit with a modified message to the _gh181-systemd-stdout_ branch.

PR #182, opened 2020-09-29
code improvement for charon-tkm.c file to avoid duplicate code block

shrestha-tripathi:
#### Reference Issues/PRs
None

#### What does this improvement fix? 
A better way to handle both SIGINT and SIGTERM signals in charon-tkm.c file, without code block duplication. This will also prevent unnecessary entry into the switch block incase the sigprocmask call fails for some reason.

### DIFF
--- a/src/charon-cmd/charon-cmd.c
+++ b/src/charon-cmd/charon-cmd.c
@@ -134,14 +134,10 @@ static int run()
 				break;
 			}
 			case SIGINT:
-			{
-				DBG1(DBG_DMN, "signal of type SIGINT received. Shutting down");
-				charon->bus->alert(charon->bus, ALERT_SHUTDOWN_SIGNAL, sig);
-				return 0;
-			}
 			case SIGTERM:
 			{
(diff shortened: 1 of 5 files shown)


tobiasbrunner:
The code is basically the same as in all the other _libcharon_ wrappers and allows easy extension e.g. to also handle `SIGHUP` or to change the behavior for a specific signal. But I guess the two cases could be combined, but that should be done in all the other wrappers too, see the _gh182-shutdown-signals_ branch.

shrestha-tripathi:
Made the changes in all libcharon wrappers.

tobiasbrunner:
I don't get it? You just copied the changes from my branch.

shrestha-tripathi:
I just wanted to contribute to the project, and you mentioned about extension for other signals like SIGHUP . So I made the changes in all Charon wrappers

tobiasbrunner:
> I just wanted to contribute to the project,

That's appreciated, but you now just (badly) copied my changes in said branch. Anyway, I've pushed my commit to master.

shrestha-tripathi:
Lol ok no problem ... I thought that’s what you wanted me to do.

PR #183, opened 2020-10-16
Update python vici - command_wrappers.py

dbotwinick:
added missing return statements in python vici library. some were actually optional, but some were required to effectively use the library. adding return statements to all functions should have no negative repercussions.


tobiasbrunner:
> but some were required

Which ones?

dbotwinick:
Actually I take it back that they're required... a bit embarrassing, but the original issues that led me to decide it was needed were actually corrected way back in c5113c810505818bad10e49204cba4ac60541488. I reviewed the current state of the code again and I think it should make sense as is right now. Probably still wouldn't _hurt_ to return from those items... but I don't think there is much to be gained from doing it. Closing the pull request.

tobiasbrunner:
Ah, I see. No problem, thanks for looking into it.

PR #184, opened 2020-10-30
Allow defining the EAP identity when using SQL plugin.

juhovh:
Currently there doesn't seem to be any way to do `EAP-MS-CHAPv2` authentication using the SQL configuration plugin as a backend. The missing piece is the lack of ability to define `AUTH_RULE_EAP_IDENTITY` and the suggested fix is to add a `eap_id` field to the `peer_configs` table. This will break backwards compatibility with existing databases, but the addition of one single column is not very invasive and should be easy to perform.

A side effect that was identified during the testing was the complications coming from using user defined identities. More specifically, since shared secrets are used for the CHAPv2 password authentication, and the shared secrets are linked to both the server identity and the user identity, it was possible to simply set the server FQDN as the username (for example vpn.example.org) and log in with any password configured for that server. This is a security hole, so an additional check was added (the `m.id != o.id` part) to make sure that the server and user identities are in fact distinct. I'm not 100% sure if this is the best fix, since it seems to me like the CHAPv2 user identities might be better off having a distinct identity type, but that kind of a change would be even a bit more invasive so I left it out.


juhovh:
Since I see I have some errors from decreasing the code coverage, I just want to let you know that I had some trouble running the tests on my local environment and didn't write tests for the functionality. If I can just get a green light that this PR is doing the right thing, I can definitely invest some time in writing tests for it as well. I just needed this change in my personal hobby project and thought to push it here for review.

tobiasbrunner:
> Currently there doesn't seem to be any way to do `EAP-MS-CHAPv2` authentication using the SQL configuration plugin as a backend. The missing piece is the lack of ability to define `AUTH_RULE_EAP_IDENTITY` and the suggested fix is to add a `eap_id` field to the `peer_configs` table.

That seems only to be a problem if the client requires an EAP-Identity exchange (I guess some like Windows do), i.e. you'd configure `ID_ANY` as EAP-Identity. Because configuring an actual identity (in particular in multiple connections) does not have the effect one might think (the _sql_ plugin is no different from other config backends in this regard).

Also, what's the advantage of configuring such static roadwarrior connections in an SQL database anyway? (You could still configure credentials in the database while configuring the connection via swanctl/vici.)

> More specifically, since shared secrets are used for the CHAPv2 password authentication, and the shared secrets are linked to both the server identity and the user identity, it was possible to simply set the server FQDN as the username (for example vpn.example.org) and log in with any password configured for that server.

Good catch. Could you please create a separate PR for this? I also wonder if it should be possible to associate the secrets only with the user identities (like it is with other backends). That is, the query, if both identities are available, could maybe use LEFT JOINs to the `shared_secret_identity` table (while ensuring that at least one is matched and maybe ordered so that multiple matches are preferred).

> Since I see I have some errors from decreasing the code coverage, I just want to let you know that I had some trouble running the tests on my local environment and didn't write tests for the functionality.

That's OK as there are no unit tests for the _sql_ plugin, which are separate from the regression tests that don't influence the coverage.

juhovh:
Thank you for your time to have a look at this and give feedback, it's much appreciated since I'm mainly working with my personal use case here, and there's so many different ways to configure strongSwan.

> That seems only to be a problem if the client requires an EAP-Identity exchange (I guess some like Windows do), i.e. you'd configure `ID_ANY` as EAP-Identity. Because configuring an actual identity (in particular in multiple connections) does not have the effect one might think (the _sql_ plugin is no different from other config backends in this regard).

I must say I haven't experimented with configuring an actual identity, since configuring `ID_ANY` was my main use case as you guessed, what would be the effect of that if not what one might think? I guess if we're certain configuring an actual identity doesn't make sense, having a way to configure the `AUTH_RULE_EAP_IDENTITY` as a boolean could work as well, but wouldn't align well with the other backends.

> Also, what's the advantage of configuring such static roadwarrior connections in an SQL database anyway? (You could still configure credentials in the database while configuring the connection via swanctl/vici.)

This is a valid question, I'm still very much in an experimentation stage myself. Basically I'm building a sort of REST API / Web UI that would allow me to configure my VPN remotely without having to access the server filesystem. I also want to do server public key authentication using a commonly trusted CA and handle certificate renewals automatically. Now, naturally I could do all this by modifying the filesystem and send reload requests to the daemon, but using the SQL backend for all persistent data as a single source of truth and if necessary VICI only for signaling to the daemon (to e.g. terminate a SA if necessary) would seem tempting. Also the ability to spin up instances from a static disk image and get all their configuration from the SQL db seems desirable.

Can you maybe shed a bit more light on what was the original motivation for the SQL backend? The documentation on it is a little bit thin, but would be interesting to hear more.

> > More specifically, since shared secrets are used for the CHAPv2 password authentication, and the shared secrets are linked to both the server identity and the user identity, it was possible to simply set the server FQDN as the username (for example vpn.example.org) and log in with any password configured for that server.
> 
> Good catch. Could you please create a separate PR for this? I also wonder if it should be possible to associate the secrets only with the user identities (like it is with other backends). That is, the query, if both identities are available, could maybe use LEFT JOINs to the `shared_secret_identity` table (while ensuring that at least one is matched and maybe ordered so that multiple matches are preferred).

Yes, it was a bit confusing for me in the beginning that for SQL the secrets are associated to both the local and remote identities. It does come with an interesting upside though, which is that one can have multiple user identities in the same db, some of which connected to server identity A and some of which connected to B. That way it could work as a sort of "user database" for a specific VPN service.

So, just to confirm that I understand this right, what you're suggesting is that I would open up a separate PR where I would keep the current change (making sure that the identities are distinct), but in addition allow shared secrets that are only connected to either the local or remote identity, not necessarily both? Basically this would mean that once a shared secret is connected to any identity, then that identity is allowed to connect (at least in the EAP-Identity case with user defined identities).

> > Since I see I have some errors from decreasing the code coverage, I just want to let you know that I had some trouble running the tests on my local environment and didn't write tests for the functionality.
> 
> That's OK as there are no unit tests for the _sql_ plugin, which are separate from the regression tests that don't influence the coverage.

👍 All clear.

tobiasbrunner:
> what would be the effect of that if not what one might think?

See [issue 1057](https://wiki.strongswan.org/issues/1057)  and related.

> but wouldn't align well with the other backends

I agree, in particular if actual identities were to be supported in the future.

> Basically I'm building a sort of REST API / Web UI that would allow me to configure my VPN remotely without having to access the server filesystem.

Did you have a look at [strongMan](https://github.com/strongswan/strongMan)? Or the [VICI](https://wiki.strongswan.org/projects/strongswan/wiki/Vici) interface?

> I also want to do server public key authentication using a commonly trusted CA and handle certificate renewals automatically.

As you mentioned, you can easily do that via swanctl (or directly via VICI). In particular, if you use Let's Encrypt via `certbot` as you can hook into the update process and simply reload the config via swanctl to use the new certificate/key.

> Can you maybe shed a bit more light on what was the original motivation for the SQL backend?

I think it was primarily created as a prototype to demonstrate the modular configuration/attribute/credential backends (added together in March 2008, there were others like [smp](https://wiki.strongswan.org/projects/strongswan/wiki/SMP)/[manger](https://wiki.strongswan.org/projects/strongswan/wiki/Manager) and of course the stroke and vici plugins). It has only seen minor development since then as there never was much interest in it. 

> It does come with an interesting upside though, which is that one can have multiple user identities in the same db, some of which connected to server identity A and some of which connected to B. That way it could work as a sort of "user database" for a specific VPN service.

Yes, that's true. If that wasn't the case, i.e. if matching the client identity was enough, and the database was shared by multiple "services" on the same host, a user of service A could authenticate at service B if there was a user with the same name (I guess that could be avoided by assigning service specific usernames).

But note that this it not how other config backends behave. For instance, with vici, you can assign an arbitrary number of identities with a secret and just one of them has to match (no matter if it's the local or the remote identity). Secrets that match both are preferred. But, configuring the local (i.e. server) identity there would obviously be a bad idea because all of these secrets would be tried during client authentication and any user could authenticate with any other username (mainly a problem if there are e.g. different groups of users that have access to different remote subnets or something similar).

> So, just to confirm that I understand this right, what you're suggesting is that I would open up a separate PR where I would keep the current change (making sure that the identities are distinct), but in addition allow shared secrets that are only connected to either the local or remote identity, not necessarily both? Basically this would mean that once a shared secret is connected to any identity, then that identity is allowed to connect (at least in the EAP-Identity case with user defined identities).

I guess your fix is enough for now, I'd just prefer it in a separate commit and PR. The second part was a suggestion to align the _sql_ plugin with the other credential sets, where a single match is enough. But considering that matching both identities could be an advantage here, we might not want to change that.

PR #185, opened 2020-11-02
controller: Fix terminate_ike's returned status in case no 'callback' is specified

sladkani:
When 'callback == NULL', terminate_ike_execute() is invoked without the
listener waiting on the ike state change.

Now, if 'force' is false, then ike_sa->delete() just queues an
IKE_DELETE task, and returns SUCCESS - indicating successful task
manager initiation.

However, terminate_ike_execute() ignores this success and sets
'listener->status = FAILED'.

This is incorrect, as this will be the overall return code of
terminate_ike(), although no failure did occur. This eventually leads
vici's "terminate" to return "Command failed: terminating SA failed",
as seen in this example:

    In [9]: list(session.terminate({'ike-id': 2960, 'timeout': -1}))
    ---------------------------------------------------------------------------
    CommandException                          Traceback (most recent call last)
    <ipython-input-9-5f95b5cea88f> in <module>()
    ----> 1 list(session.terminate({'ike-id': 2960, 'timeout': -1}))

    vici/session.pyc in streamed_request(self, command, event_stream_type, message)
        136                 raise CommandException(
        137                     "Command failed: {errmsg}".format(
    --> 138                         errmsg=command_response["errmsg"]
        139                     )
        140                 )

    CommandException: Command failed: terminating SA failed

Fix: In case no 'callback' is specified (i.e. caller intends NOT to
wait for the result of the deletion), and ike_sa->delete() returns
SUCCESS, also set 'listener->status = SUCCESS' to indicate the
IKE_DELETE was queued succesfully.

Signed-off-by: Shmulik Ladkani <shmulik@metanetworks.com>

### DIFF
--- a/src/libcharon/control/controller.c
+++ b/src/libcharon/control/controller.c
@@ -553,6 +553,7 @@ METHOD(job_t, terminate_ike_execute, job_requeue_t,
 {
 	interface_listener_t *listener = &job->listener;
 	uint32_t unique_id = listener->id;
+	status_t delete_status;
 	ike_sa_t *ike_sa;
 
 	ike_sa = charon->ike_sa_manager->checkout_by_id(charon->ike_sa_manager,
@@ -569,9 +570,18 @@ METHOD(job_t, terminate_ike_execute, job_requeue_t,
 	listener->ike_sa = ike_sa;
 	listener->lock->unlock(listener->lock);
 
-	if (ike_sa->delete(ike_sa, listener->options.force) != DESTROY_ME)
+	delete_status = ike_sa->delete(ike_sa, listener->options.force);
+	if (delete_status != DESTROY_ME)
 	{	/* delete queued */
-		listener->status = FAILED;
+		if (!listener->logger.callback && delete_status == SUCCESS)
+		{
+			/* queued succesfully and no callback set, report as success */
+			listener->status = SUCCESS;
+		}
+		else
+		{
+			listener->status = FAILED;
+		}
 		charon->ike_sa_manager->checkin(charon->ike_sa_manager, ike_sa);
 	}
 	else


tobiasbrunner:
I wonder if this is actually on purpose to allow differentiating between actual termination (forced or due to the SA's state, or after waiting for the response) and just initiating an exchange (in which case the IKE_SA has most likely not yet been terminated when the call returns, so e.g. counting it in `terminated` is not entirely correct). Note that termination with a timeout will also result in an error if the SA has not yet been destroyed when the timeout expires, even with `force` set (`OUT_OF_RES` is returned in either case, which _vici_ treats as failure, as documented in the README).

sladkani:
> I wonder if this is actually on purpose to allow differentiating between actual termination (forced or due to the SA's state, or after waiting for the response) and just initiating an exchange (in which case the IKE_SA has most likely not yet been terminated when the call returns, so e.g. counting it in `terminated` is not entirely correct). 

Not sure I follow your question here. 

Currently, vici, passing a NULL callback (e.g. when given a negative timeout), reports "terminating SA failed", ALTHOUGH the terminate is perfectly initiated (and in all normal cases, succeeds when due) - see this snip from vici_control.c:

```
	while (enumerator->enumerate(enumerator, &del))
	{
		if (child || child_id)
		{
		    ...
		}
		else
		{
			if (charon->controller->terminate_ike(charon->controller, *del, force,
											log_cb, &log, timeout) == SUCCESS)
			{
				done++;
			}
		}
	}
	
	...
	
	else if (done < array_count(ids))
	{
		if (array_count(ids) == 1)
		{
			errmsg = "terminating SA failed";
		}
		else
		{
			errmsg = "not all matching SAs could be terminated";
		}
	}
```

Since the return code in the non-callback case is FAILED, how can one differentiate between the scenarios?

BTW: similar "error" reporting is also done in charon-xpc/xpc_channels.c `stop_connection` (which also passes NULL callback and tests the return code).

My claim is that if one issues `controller->terminate_ike()` with a NULL callback, he wishes NOT to wait for the terminate result, and as such, we aren't allowed to return a FAILED indication for his "fire and forget" request.

Whats your idea for differentiating this cases, and how do you think the call-sites need to be amended?

tobiasbrunner:
> Since the return code in the non-callback case is FAILED, how can one differentiate between the scenarios?

What do you mean?

> My claim is that if one issues `controller->terminate_ike()` with a NULL callback, he wishes NOT to wait for the terminate result, and as such, we aren't allowed to return a FAILED indication for his "fire and forget" request.

It might still be of interest if the SA was actually immediately destroyed (and is gone afterwards), or only a delete was queued (and the SA and its CHILD_SAs might still exist for a while). Not sure if there is an actual use case for this, though. But it was always like that, hence my questioning whether it might have been on purpose.

One could also look at this like an instant timeout, which, by definition, fails if termination did not succeed before, which is only the case here if the SA can be destroyed immediately.

sladkani:
> > My claim is that if one issues `controller->terminate_ike()` with a NULL callback, he wishes NOT to wait for the terminate result, and as such, we aren't allowed to return a FAILED indication for his "fire and forget" request.
> 
> It might still be of interest if the SA was actually immediately destroyed (and is gone afterwards), or only a delete was queued (and the SA and its CHILD_SAs might still exist for a while). 

Note that if the caller wishes an immediate destruction, he would set force = TRUE
(if a NULL callback is specified, and force is TRUE, then SA gets immediately destroyed, `ike_sa->delete()` returns DESTROY_ME and status is set to SUCCESS).

If the caller sets force = FALSE, then the delete is just queued.

> Not sure if there is an actual use case for this, though. But it was always like that, hence my questioning whether it might have been on purpose.

OTOH if we consider the FAILED report for the case where the "delete is just queued" was **on purpose**, how can the user know the resulting "terminating SA failed" is not a real error but signals a "successful queue"?

tobiasbrunner:
> Note that if the caller wishes an immediate destruction, he would set force = TRUE

Yes, I was arguing that the result might still be of interest to the caller for some reason. And it also allows to accurately set the counter in the response if `terminated` should denote the number of SAs gone after the call returns and not also include those with just a queued delete (for which a different counter could be used, I guess).

> If the caller sets force = FALSE, then the delete is just queued.

Unless, the SA is in a state that doesn't allow queuing the delete and it gets destroyed immediately, which is what can be detected if there is a different status. Whether using FAILED is a good idea, is a different question (NEED_MORE would fit better, but for some reason that's already used for when the callback returns FALSE, other existing status values don't really fit).

> how can the user know the resulting "terminating SA failed" is not a real error but signals a "successful queue"?

Termination can only _really_ fail if the SA is not found (which could be distinguished via the number of `matches` in the response, at least if the number of expected SAs is known - however, in the Python bindings that's currently not actually an option as the response is not available in the exception, or even when returning successfully, as _streamed_request()_ does not return the response). The other "failure" cases (timeout, queued exchange) could be handled by vici appropriately (e.g. with separate counters). But if there is no real use case for this, it seems like over-engineering.

To be honest, I really don't have anything against changing this. I just wondered about the current code and wanted to discuss whether there might be a reason/use case for a different return value. The change could even be simplified because both paths will set the status to `SUCCESS` if no callback is set (see the commit in the _gh185-vici-terminate_ branch).

sladkani:
Thanks @tobiasbrunner  for your comments and assistance.

PR #186, opened 2020-11-30
openssl: Use ECDH_compute_key() for 'x-coordinate only' setting

mahantesh-nvidia:
* ECDH_compute_key() was not used because it only gives
  x-coordinate of the result. However, the default setting,
  as per the errata mentioned, is to use x-coordinate only.
  Use ECDH_compute_key() for this setting as it additionally
  allows hw offload of the computation using dynamic engine
  feature in OpenSSL.
  EC_POINT_mul() doesn't allow hw offload.

Signed-off-by: Mahantesh Salimath <mahantesh@nvidia.com>

### DIFF
--- a/src/libstrongswan/plugins/openssl/openssl_ec_diffie_hellman.c
+++ b/src/libstrongswan/plugins/openssl/openssl_ec_diffie_hellman.c
@@ -179,33 +179,47 @@ static bool compute_shared_key(private_openssl_ec_diffie_hellman_t *this,
 	EC_POINT *secret = NULL;
 	bool x_coordinate_only, ret = FALSE;
 
-	priv_key = EC_KEY_get0_private_key(this->key);
-	if (!priv_key)
-	{
-		goto error;
-	}
-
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks. Note that `ECDH_compute_key()` is deprecated and will apparently be removed with OpenSSL 3.0. So this whole thing will have to be refactored to use the EVP interface. For now, I pushed the commit, with some changes (like error handling), to the _gh186-openssl-ecp_ branch and will try to work on refactoring this to the EVP interface. I also don't think the _ecp_x_coordinate_only_ option is still necessary, so I'll remove that too.

tobiasbrunner:
I pushed additional commits to the branch to refactor this (mostly) to the EVP interface. Let me know if this still works with HW offloading.

mahantesh-nvidia:
Thanks @tobiasbrunner , I tested with your additional commits and it still works with HW offloading.

tobiasbrunner:
OK, thanks for testing. I'll close the PR once I merge the branch.

PR #187, opened 2020-12-02
libtls: Add 'tls.send_certreq_authorities' configuration knob

sladkani:
Formerly, when tls server builds the Certificate Request message, all
loaded X509_CA certificates are written into the certificate_authorities
list.

Alas, certain EAP-TLS clients fail to process this message if the
certificate_authorities list is too long, returning fatal TLS alert
'illegal parameter'.

Allow configuring whether the tls server fills the
certificate_authorities list with its known CAs or not; Providing an
empty list is a valid configuration according to TLS spec:

   certificate_authorities
      A list of the distinguished names [X501] of acceptable
      certificate_authorities, represented in DER-encoded format.  These
      distinguished names may specify a desired distinguished name for a
      root CA or for a subordinate CA; thus, this message can be used to
      describe known roots as well as a desired authorization space.  If
      the certificate_authorities list is empty, then the client MAY
      send any certificate of the appropriate ClientCertificateType,
      unless there is some external arrangement to the contrary.

The default is TRUE which preserves former behavior.

### DIFF
--- a/conf/options/charon.opt
+++ b/conf/options/charon.opt
@@ -476,6 +476,10 @@ charon.tls.mac
 charon.tls.suites
 	List of TLS cipher suites.
 
+charon.tls.send_certreq_authorities
+	In server's Certificate Request message, controls whether
+	certificate_authorities list should be populated.
+
 charon.user
 	Name of the user the daemon changes to after startup.
 
--- a/src/libtls/tls_server.c
+++ b/src/libtls/tls_server.c
@@ -153,6 +153,11 @@ struct private_tls_server_t {
 	 * Did we receive the curves from the client?
 	 */
 	bool curves_received;
+
+	/**
+	 * Should we send a certificate_authorities in the CertificateRequest
+	 */
+	bool send_certreq_authorities;
 };
 
 /**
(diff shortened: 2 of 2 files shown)


tobiasbrunner:
> Alas, certain EAP-TLS clients fail to process this message if the
> certificate_authorities list is too long, returning fatal TLS alert
> 'illegal parameter'.

Seems strange. How long was the list?

Anyway, this might be something to consider, but some of the affected code might change when adding TLS 1.3 support (_tls13_ branch, plus current work for server-side support over at ryru/strongswan). So I'll review it once I merge that code.

sladkani:
> > Alas, certain EAP-TLS clients fail to process this message if the
> > certificate_authorities list is too long, returning fatal TLS alert
> > 'illegal parameter'.
> 
> Seems strange. How long was the list?

Didn't pinpoint to the exact threshold, but strongSwan had ~270 CAs loaded.
Client was windows (both 7 and 10), cert based ikev2 in eap-tls mode.

> Anyway, this might be something to consider, but some of the affected code might change when adding TLS 1.3 support (_tls13_ branch, plus current work for server-side support over at ryru/strongswan). So I'll review it once I merge that code.

Thanks. Do note that in certain server configurations or deployments, specifying the CA "hints" in the certificate_authorities list is pointless as the server in such deployments do not need to "describe known roots/desired authorization space".

tobiasbrunner:
> but strongSwan had ~270 CAs loaded.

On a server? That seems very risky as without further restrictions (e.g. on identities) it would allow anybody with a client or server certificate from any of these (presumably commercial) CAs to connect to that server.

sladkani:
> > but strongSwan had ~270 CAs loaded.
> 
> On a server? That seems very risky as without further restrictions (e.g. on identities) it would allow anybody with a client or server certificate from any of these (presumably commercial) CAs to connect to that server.

Yes, numerous restrictions apply. The CAs are from a private chain of trust. Identities are also being restricted. And there are other forms of restrictions.

tobiasbrunner:
A modified version of your patch is now in the _tls13_ branch.

PR #188, opened 2020-12-15
ipsec: Adding full hw-offload configuration support

ayungman:
Full offload does both crypto and encapsulation in HW, reducing SW overhead.
In this mode HW fully offloads the ESP data-path including: crypto, transport mode encapsulation, replay protection, ESP sequence number generation, ESN, error reporting.

This commit introduces a new configuration mode: **hw_offload = full**

Until now the configuration available to user for HW offload were:
- hw_offload = no
- hw_offload = yes
- hw_offload = auto
 
With this commit users will be able to configure full-offload using:
- hw_offload = full

Changes:
new flag HW_OFFLOAD_FULL added
modified config_hw_offload() to support the new flag
new value in swanctl.conf connections:children:hw_offload=full

### DIFF
--- a/src/include/linux/xfrm.h
+++ b/src/include/linux/xfrm.h
@@ -503,6 +503,7 @@ struct xfrm_user_offload {
 };
 #define XFRM_OFFLOAD_IPV6	1
 #define XFRM_OFFLOAD_INBOUND	2
+#define XFRM_OFFLOAD_FULL	4
 
 #ifndef __KERNEL__
 /* backwards compatibility for userspace */
--- a/src/libcharon/plugins/kernel_netlink/kernel_netlink_ipsec.c
+++ b/src/libcharon/plugins/kernel_netlink/kernel_netlink_ipsec.c
@@ -1448,13 +1448,15 @@ static bool netlink_detect_offload(const char *ifname)
 #endif
 
 /**
- * There are 3 HW offload configuration values:
+ * There are 4 HW offload configuration values:
  * 1. HW_OFFLOAD_NO   : Do not configure HW offload.
  * 2. HW_OFFLOAD_YES  : Configure HW offload.
  *                      Fail SA addition if offload is not supported.
  * 3. HW_OFFLOAD_AUTO : Configure HW offload if supported by the kernel
  *                      and device.
  *                      Do not fail SA addition otherwise.
+ * 4. HW_OFFLOAD_FULL : Configure HW FULL offload if supported by the kernel
+ *                      Fail SA addition if offload is not supported.
  */
 static bool config_hw_offload(kernel_ipsec_sa_id_t *id,
 							  kernel_ipsec_add_sa_t *data, struct nlmsghdr *hdr,
@@ -1471,7 +1473,8 @@ static bool config_hw_offload(kernel_ipsec_sa_id_t *id,
 		return TRUE;
 	}
 
-	hw_offload_yes = (data->hw_offload == HW_OFFLOAD_YES);
+	hw_offload_yes = ( (data->hw_offload == HW_OFFLOAD_YES) ||
+			   (data->hw_offload == HW_OFFLOAD_FULL) );
 
 	if (!charon->kernel->get_interface(charon->kernel, local, &ifname))
 	{
@@ -1499,6 +1502,12 @@ static bool config_hw_offload(kernel_ipsec_sa_id_t *id,
 		offload->flags |= XFRM_OFFLOAD_IPV6;
 	}
 	offload->flags |= data->inbound ? XFRM_OFFLOAD_INBOUND : 0;
+
+	/* activate full HW offload */
+	if ( data->hw_offload == HW_OFFLOAD_FULL )
+	{
+		offload->flags |= XFRM_OFFLOAD_FULL;
+	}
 
 	ret = TRUE;
 
--- a/src/libcharon/plugins/vici/vici_config.c
+++ b/src/libcharon/plugins/vici/vici_config.c
@@ -1031,6 +1031,7 @@ CALLBACK(parse_hw_offload, bool,
 		{ "no",		HW_OFFLOAD_NO	},
 		{ "yes",	HW_OFFLOAD_YES	},
 		{ "auto",	HW_OFFLOAD_AUTO	},
+		{ "full",	HW_OFFLOAD_FULL	},
 	};
 	int d;
 
--- a/src/libstrongswan/ipsec/ipsec_types.c
+++ b/src/libstrongswan/ipsec/ipsec_types.c
@@ -37,10 +37,11 @@ ENUM(ipcomp_transform_names, IPCOMP_NONE, IPCOMP_LZJH,
 	"IPCOMP_LZJH"
 );
 
-ENUM(hw_offload_names, HW_OFFLOAD_NO, HW_OFFLOAD_AUTO,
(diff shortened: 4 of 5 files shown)


tobiasbrunner:
Why does this require a separate option? Is there a reason not to use it (i.e. configure `yes`) if the hardware supports full offloading? Is it possible to always add the XFRM flag if offloading is enabled to use the most efficient offloading available (or would that result in a kernel error if full offloading is not supported by the hardware)? Similarly, how does this interact with `auto` (i.e. can we detect the hardware's abilities for full/partial offload and enable the most efficient form automatically)?

ayungman:
One reason not to use the other values is backward compatibility.
Other is kernel raise error if full offloading is not supported by the hardware.
Yes is already known and used as crypto only. Might confuse and fail connection.
Auto is documented and used as try crypto but do not fail if no available, so again the user is not expected to different behavior.

tobiasbrunner:
> One reason not to use the other values is backward compatibility.

What do you mean?

> Other is kernel raise error if full offloading is not supported by the hardware.

Hence my question whether we can detect what kind of offloading the hardware supports and enable the most efficient.

> Yes is already known and used as crypto only. Might confuse and fail connection.

What's known? Confuse what and fail how?

> Auto is documented and used as try crypto but do not fail if no available, so again the user is not expected to different behavior.

As user, I'd expect `auto` to configure the "best" offloading available (or gracefully fall back to the software implementation if none is available). Why should `auto` or even `yes` be limited to offloading crypto only?

A possible reason for new values (`full` and maybe `crypto`, but again, why would you want to offload crypto only if the hardware supports full offloading) I can see is to request a specific mode explicitly and force a failure if the hardware does not support that specific mode. But if you don't care whether it's full offloading or crypto only (which is dependent on the hardware and kernel in use, so forcing a mode that's not supported seems not really useful), that's already covered by `yes`.

ayungman:
What your saying "_As user, I'd expect_" sound logical, but it luck the option to force Crypto only for kernel that does not support it. if we set the flag XFRM_OFFLOAD_FULL on **Yes** then kernel might fail the SA if it does not support full and will not try to do Crypto only. So need to tell it what to do. I do agree auto should do it, not sure kernel has that fall logic. 
maybe the best value were explicit:  no, crypto, full.

tobiasbrunner:
> but it luck the option to force Crypto only for kernel that does not support it

Sorry, I don't understand.

> if we set the flag XFRM_OFFLOAD_FULL on **Yes** then kernel might fail the SA if it does not support full and will not try to do Crypto only

Adding the flag shouldn't matter if the kernel does not know it (i.e. if it's an older kernel that has no support for full offloading yet) as binary flags are usually simply checked with `&` so additional flags don't matter.

The problem is if the kernel knows it but the hardware doesn't support full offloading. So again my question regarding querying the level of support from the hardware in `netlink_find_offload_feature()/netlink_detect_offload()`. How do the device features differ if the hardware supports full offloading as compared to crypto only? If there is no difference, that would be disappointing and should probably be changed on the kernel/driver level.

BorisPis:
There may be reasons for the user to control whether one offload is used and not the other. For example, if the administrator trusts the crypto offload implementation only for some SAs, but not for others. So, we want to expose fine-grained control of these features rather than sticking to auto alone. The question is how to expose fine-grained control?

Let's say that auto should always use the best option, i.e. full offload (new) falls back to cyrpto offload (existing) which falls back to software.  Then how do we accommodate users who desire crypto offload for one and full offload for another SA, where HW can use full offload for both?

tobiasbrunner:
> For example, if the administrator trusts the crypto offload implementation only for some SAs, but not for others.

It's already possible to disable offloading per SA. But is there a use case to use full offloading for one SA and limit offloading to crypto only for another (if the interfaces that they use support full offloading and/or might even be the same)? That is, if you don't trust the full offload feature, would you trust the crypto offload feature of the same device? (Maybe because the full offload feature implementation is incomplete and crypto offload with software framing provides missing features?)

But could that maybe also be solved with a global strongswan.conf option that enables full offloading (or disables it, depending on what we can detect and the default should be)? Or is it likely that there are several interfaces used by different SAs and on some you want to use full offload while on others you only want to offload crypto (even if full offload is supported, because otherwise  we don't have a problem provided we can we can detect the supported offload level for each interface)? Considering that, why is this even an option per SA and not on the driver/interface level?

> Then how do we accommodate users who desire crypto offload for one and full offload for another SA, where HW can use full offload for both?

As long as _auto_ and _yes_ (without fallback to software) uses the best available option (i.e. we can detect what that is for each interface), I've no problem with adding additional settings that limit and require offloading at a specific level (e.g. `full` and `crypto`) if there is a use case for that.

BorisPis:
> It's already  #possible to disable offloading per SA.

We don't want to disable it, but rather choose crypto or full offload explicitly.

> But is there a use case to use full offloading for one SA and limit offloading to crypto only for another (if the interfaces that they use support full offloading and/or might even be the same)? That is, if you don't trust the full offload feature, would you trust the crypto offload feature of the same device? (Maybe because the full offload feature implementation is incomplete and crypto offload with software framing provides missing features?)

Exactly as you suggested, sometimes full offload will be too restrictive on either (1) reply-protection, (2) framing, (3) selectors, (4) sequence number generation, (5) soft/hard limits, (6) packet fragmentation causing loss, etc. Therefore, providing flexibility to the user is important. 

To be more specific, a major use-case we envision for full offload involves virtualization: suppose some VM communicates with another over the network. The hypervisor tunnels VM packets within an overlay (e.g., VXLAN, GRE, etc.). We want to protect this tunnel. The NIC already can offload overlays, and with full offload it can also offload IPsec. The performance gains of this approach are an order of magnitude better compared to existing software-based solutions, especially if the CPU is not very powerful (e.g., ARM-based smartNIC). Additionally, overlays often employ the IP don't fragment bit, and this helps avoid case (6) above.

Now, suppose the user wants to offload only VM encapsulated traffic with full-offload, while other traffic, such as SSH session from the Internet to the hypervisor should use crypto offload, as Intenet traffic is more likely to be fragmented. How should we approach this using strongswan?
Per-SA configuration works fine as the user knows what is needed. But, global configuration is tricky, as it may be hard to determine automatically, what is an overlay and what is directed to the internet, right?

> But could that maybe also be solved with a global strongswan.conf option that enables full offloading (or disables it, depending on what we can detect and the default should be)? Or is it likely that there are several interfaces used by different SAs and on some you want to use full offload while on others you only want to offload crypto (even if full offload is supported, because otherwise we don't have a problem provided we can we can detect the supported offload level for each interface)? Considering that, why is this even an option per SA and not on the driver/interface level?
> 

For simplicity, we can provide an easy automatic configuration for: try full-offload, otherwise fallback to crypto, and otherwise fallback to software. However, it may struggle to capture the use-case described above. This can be managed in kernel XFRM.

> > Then how do we accommodate users who desire crypto offload for one and full offload for another SA, where HW can use full offload for both?
> 
> As long as _auto_ and _yes_ (without fallback to software) uses the best available option (i.e. we can detect what that is for each interface), I've no problem with adding additional settings that limit and require offloading at a specific level (e.g. `full` and `crypto`) if there is a use case for that.

So we will expand auto to capture the (full or crypto) case and add an option to choose whether full or crypto is disabled for some SA.

tobiasbrunner:
Was eventually added via #1462.

PR #190, opened 2021-01-18
TLS 1.3 downgrade protection patch

ryru:
This patch add downgrade protection as specified in RFC 8446 section 4.1.3.


src/libtls/tls_peer.c
```diff
@@ -286,6 +287,28 @@ static status_t process_server_hello(private_tls_peer_t *this,
 	}
 	extensions->destroy(extensions);
 
+	/* downgrade protection (see RFC 8446, section 4.1.3) */
+	memcpy(server_random_end, &this->server_random[24], 8);
```
tobiasbrunner, line 291: I'd use chunks for this. This one could be
```suggestion
	server_random_end = chunk_create(&this->server_random[24], 8);
```

src/libtls/tls.h
```diff
@@ -213,6 +213,10 @@ extern enum_name_t *tls_extension_names;
  */
 extern chunk_t tls_hello_retry_request_magic;
 
+/* Magic values for downgrade protection (see RFC 8446, section 4.1.3) */
+extern char tls_downgrade_protection_tls11[];
+extern char tls_downgrade_protection_tls12[];
```
tobiasbrunner, line 218: Maybe use chunks instead? (see other comments)

src/libtls/tls_peer.c
```diff
@@ -286,6 +287,28 @@ static status_t process_server_hello(private_tls_peer_t *this,
...
+	if (this->tls->get_version_max(this->tls) == TLS_1_3 &&
+		version < TLS_1_3)
+	{
+		if (memcmp(server_random_end, tls_downgrade_protection_tls11, 8) == 0 ||
+			memcmp(server_random_end, tls_downgrade_protection_tls12, 8) == 0)
```
tobiasbrunner, line 296: There is a actually a `memeq()` helper to avoid the `== 0`, but with chunks it's even easier as you don't have to add the length again.
```suggestion
		if (chunk_equals(server_random_end, tls_downgrade_protection_tls11) ||
			chunk_equals(server_random_end, tls_downgrade_protection_tls12))
```

src/libtls/tls_peer.c
```diff
@@ -286,6 +287,28 @@ static status_t process_server_hello(private_tls_peer_t *this,
...
+		{
+			this->alert->add(this->alert, TLS_FATAL, TLS_ILLEGAL_PARAMETER);
+			return NEED_MORE;
+		}
+	}
+	if (this->tls->get_version_max(this->tls) == TLS_1_2 &&
```
tobiasbrunner, line 302: Maybe use `else if` here.

src/libtls/tls_server.c
```diff
@@ -1159,6 +1159,23 @@ static status_t send_server_hello(private_tls_server_t *this,
 	}
 	else
 	{
+		if (this->tls->get_version_max(this->tls) == TLS_1_3 &&
+			version < TLS_1_3)
```
tobiasbrunner, line 1163: This can never be true because `version` is set to `this->tls->get_version_max()` just above. So this patching of `server_random` should probably happen in `process_client_hello()` when we still can determine what the original max. version was (e.g. store it in a local variable before the `if (versions.len)` check and add this fixup after determining/setting the version).

src/libtls/tls_server.c
```diff
@@ -1159,6 +1159,23 @@ static status_t send_server_hello(private_tls_server_t *this,
...
+			char *downgrade_protection;
+
+			downgrade_protection = tls_downgrade_protection_tls11;
+			if (version == TLS_1_2)
+			{
+				downgrade_protection = tls_downgrade_protection_tls12;
+			}
+			memcpy(&this->server_random[24], downgrade_protection, 8);
+		}
+		if (this->tls->get_version_max(this->tls) == TLS_1_2 &&
```
tobiasbrunner, line 1174: Maybe use `else if`.

tobiasbrunner (changes requested):
Thanks, I've added some notes.

src/libstrongswan/utils/chunk.h
```diff
@@ -129,7 +129,7 @@ bool chunk_from_fd(int fd, chunk_t *chunk);
 chunk_t *chunk_map(char *path, bool wr);
 
 /**
- * munmap() a chunk previously mapped with chunk_map()
+ * unmap() a chunk previously mapped with chunk_map()
```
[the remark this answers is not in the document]
ryru: :flushed:

ryru:
I addaped your requests and tested downgrade functionality:

- Case 1: set `original_max_version` to `TLS_1_3` and start a server with `./scripts/tls_test --listen localhost --port 8443 --key caKey.der --cert cert.der --debug 2 --max-version 1.2`. Client ends connection with a illegal parameter fatal alert as expected.

- Case 2: set `original_max_version` to `TLS_1_2` and start a server with `./scripts/tls_test --listen localhost --port 8443 --key caKey.der --cert cert.der --debug 2 --max-version 1.1`. Client ends connection with a illegal parameter fatal alert as expected.

src/libtls/tls_server.c
```diff
@@ -482,6 +484,27 @@ static status_t process_client_hello(private_tls_server_t *this,
...
 		}
 	}
+
+	/* downgrade protection (see RFC 8446, section 4.1.3) */
+	if (original_max_version == TLS_1_3 &&
+		version < TLS_1_3)
+	{
+		char *downgrade_protection;
+
+		downgrade_protection = tls_downgrade_protection_tls11.ptr;
```
ryru, line 494: To me it seems to be the easiest way to work with a pointer to char since the server-random value is stored as char anyway.

src/libtls/tls_peer.c
```diff
@@ -286,6 +286,28 @@ static status_t process_server_hello(private_tls_peer_t *this,
...
 
+	/* downgrade protection (see RFC 8446, section 4.1.3) */
+	server_random_end = chunk_create(&this->server_random[24], 8);
+	if (this->tls->get_version_max(this->tls) == TLS_1_3 &&
+		version < TLS_1_3)
+	{
+		if (chunk_equals(server_random_end, tls_downgrade_protection_tls11) ||
+			chunk_equals(server_random_end, tls_downgrade_protection_tls12))
+		{
+			this->alert->add(this->alert, TLS_FATAL, TLS_ILLEGAL_PARAMETER);
```
ryru, line 297: A debug message could be helpful here:
`DBG2(DBG_TLS, "invalid server random possible downgrade attack");`

tobiasbrunner (changes requested):

tobiasbrunner:
Great, thanks a lot. I've pushed the commit with some minor changes (mainly renamed the variables so they match the other code and the name of the method) to the _tls13_ branch (interesting that GitHub added this as an event to this PR).

PR #192, opened 2021-01-22
Use strongswan RNG interface in Botan plugin

securitykernel:
Uses a new API introduced in https://github.com/randombit/botan/pull/2600 to register the internal RNG crypto API as a random number generator for use in all Botan API calls that require a random number generator. With this in place, users using the Botan plugin can make it use an different random number generator via another plugin instead of being forced to use Botan's RNG.

This API functions is likely introduced in one of the upcoming versions of Botan, so we need to decide what to do until then, e.g., provide some kind of fallback to the old behaviour.


tobiasbrunner:
Looks interesting. Wouldn't a destructor (that receives the context) make sense for `botan_rng_init_custom()` (the original draft in the issue had that)? That would allow simplifying the code quite a bit as `botan_get_strongswan_rng()` could handle the creation of the `rng_t` object and `botan_rng_destroy()` could automatically destroy that instance again.

> This API functions is likely introduced in one of the upcoming versions of Botan, so we need to decide what to do until then, e.g., provide some kind of fallback to the old behaviour.

Yes, a fallback is definitely necessary (would be a lot easier with the above as everything could be handled in `botan_get_strongswan_rng()` or whatever we call it). If a define like `BOTAN_HAS_CUSTOM_RNG` was available, that could easily be done (otherwise, we could perhaps add a configure check for it).

I also wonder if a config option would make sense to force the plugin to use the internal RNG even if other plugins providing RNGs are loaded.

securitykernel:
@tobiasbrunner Good idea, added back the destructor and refactor the plugin. Removed quite a bit of duplicated code. Now the only thing missing is the fallback for older Botan versions. Unfortunately, there's no `BOTAN_HAS_CUSTOM_RNG` definition from Botan. For the C API, you can only query the API version at runtime using `botan_ffi_api_version()`. Its probably the best way to add a configure check to strongswan.

securitykernel:
I may have inadvertently broke ed25519 signatures in the Botan plugin, but looking at the diff now, I can not spot the error. If you have a hint ...

```
 Passed all 2 'prf_plus' test cases
  Running suite 'ed25519':
    Running case 'ed25519_sign': -----
      Failure in 'test_ed25519_sign': pubkey != NULL (suites/test_ed25519.c:326, i = 0)
      Failure in 'test_ed25519_sign': pubkey != NULL (suites/test_ed25519.c:326, i = 1)
      Failure in 'test_ed25519_sign': pubkey != NULL (suites/test_ed25519.c:326, i = 2)
      Failure in 'test_ed25519_sign': pubkey != NULL (suites/test_ed25519.c:326, i = 3)
      Failure in 'test_ed25519_sign': pubkey != NULL (suites/test_ed25519.c:326, i = 4)
    Running case 'ed25519_gen': -
      Failure in 'test_ed25519_gen': pubkey->get_encoding(pubkey, PUBKEY_PEM, &encoding) (suites/test_ed25519.c:395, i = 0)
    Running case 'ed25519_fail': -
      Failure in 'test_ed25519_fail': pubkey != NULL (suites/test_ed25519.c:529, i = 0)
    Running case 'ed25519_speed': +
  Passed 1/4 'ed25519' test cases
```

tobiasbrunner (changes requested):
Thanks. That looks a lot cleaner already.

>  Its probably the best way to add a configure check to strongswan.

Agreed, you could e.g. add the following to the `test x$botan = xtrue` block in `configure.ac`:

	saved_LIBS=$LIBS
	LIBS="$botan_LIBS"
	AC_CHECK_FUNCS(botan_rng_init_custom)
	LIBS=$saved_LIBS

So you could then check if the function is available via `#ifdef HAVE_BOTAN_RNG_INIT_CUSTOM`.

Note that the module check will later have to be adapted for Botan 3, that is, a check for `botan-3` via `PKG_CHECK_MODULES` will be necessary. Something like:

```
	PKG_CHECK_MODULES(botan, [botan-3], [],
		[PKG_CHECK_MODULES(botan, [botan-2])])
```



> I may have inadvertently broke ed25519 signatures in the Botan plugin, but looking at the diff now, I can not spot the error. If you have a hint ...

Yes, the plugin features in `f_pubkey` are not actually added to `f`. But as mentioned in the comment, these changes might not be necessary.

src/libstrongswan/plugins/botan/botan_util.c
```diff
@@ -336,11 +323,28 @@ int strongswan_get_random(uint8_t *out, size_t out_len, rng_t *rng)
...
+	{
+		rng->destroy(rng);
+	}
+}
+
+bool botan_get_rng(botan_rng_t *botan_rng)
 {
-	if(botan_rng_init_custom(botan_rng, "strongswan", rng, strongswan_get_random, NULL))
+	// TODO add a fallback for Botan versions that don't ship this API function
+	rng_t* rng = lib->crypto->create_rng(lib->crypto, RNG_STRONG);
```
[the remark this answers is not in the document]
securitykernel: Changes applied.

src/libstrongswan/plugins/botan/botan_plugin.c
```diff
@@ -351,8 +381,34 @@ METHOD(plugin_t, get_features, int,
...
+		plugin_features_add(f, f_hmac, countof(f_hmac), &count);
+		plugin_features_add(f, f_privkey, countof(f_privkey), &count);
+		plugin_features_add(f, f_rsa, countof(f_rsa), &count);
+		plugin_features_add(f, f_ecdsa, countof(f_ecdsa), &count);
+		plugin_features_add(f, f_ed25519, countof(f_ed25519), &count);
+		if (lib->settings->get_bool(lib->settings,
+								"%s.plugins.botan.use_rng", TRUE, lib->ns))
+		{
+			plugin_features_add(f, f_rng, countof(f_rng), &count);
+		}
```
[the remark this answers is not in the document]
securitykernel: Reverted the changes to `botan_plugin.c` and moved the check to `botan_get_rng()`.

src/libstrongswan/plugins/botan/botan_util.c
```diff
@@ -313,3 +313,78 @@ bool botan_dh_key_derivation(botan_privkey_t key, chunk_t pub, chunk_t *secret)
 	botan_pk_op_key_agreement_destroy(ka);
 	return TRUE;
 }
+
+int strongswan_get_random(rng_t *rng, uint8_t *out, size_t out_len)
```
tobiasbrunner, line 317: These should be static and the signature doesn't match what `botan_rng_init_custom()` expects. Both things can be fixed by using our `CALLBACK()` macro.

src/libstrongswan/plugins/botan/botan_x25519.c
```diff
@@ -155,7 +155,7 @@ diffie_hellman_t *botan_x25519_create(diffie_hellman_group_t group)
 		},
 	);
 
-	if (botan_rng_init(&rng, "user"))
+	if (!botan_get_rng(&rng, RNG_TRUE))
```
tobiasbrunner, line 158: Note that `"user"` is usually mapped to `RNG_STRONG`.

src/libstrongswan/plugins/botan/botan_rsa_private_key.c
```diff
@@ -397,7 +397,7 @@ static bool calculate_pq(botan_mp_t *n, botan_mp_t *e, botan_mp_t *d,
 {
 	botan_mp_t k = NULL, one = NULL, r = NULL, zero = NULL, two = NULL;
 	botan_mp_t n1 = NULL, x = NULL, y = NULL, g = NULL, rem = NULL;
-	botan_rng_t rng = NULL;
+	botan_rng_t rng;
```
tobiasbrunner, line 400: This is actually necessary because we use `goto` in this function and if there is an error before the RNG is initialized, we would crash when trying to destroy this uninitialized RNG.

tobiasbrunner (approved):
Looks good. I've added some comments but they are just to clarify some things, as I've already pushed the commits, squashed together and with some refactorings/fixups, to the _gh192-botan-rng_ branch.

I'll wait with merging the branch until randombit/botan#2600 is closed.

src/libstrongswan/plugins/botan/botan_x25519.c
securitykernel: Missed that one.

src/libstrongswan/plugins/botan/botan_util.c
securitykernel: Didn't know this callback, thx.

src/libstrongswan/plugins/botan/botan_rsa_private_key.c
securitykernel: But Botan will error out with a `nullptr` `botan_rng_t` being passed, see https://github.com/randombit/botan/blob/release-2/src/lib/ffi/ffi_rng.cpp#L25. So to me it looks like the whole function is never actually executed. That is actually a thing I wanted to point out already when pushing this change earlier this week.

tobiasbrunner: No, it errors out if you pass it a NULL pointer (i.e. if you'd call `botan_rng_init(NULL, "system")`), not if the passed pointer points to NULL.

securitykernel: You're right.

PR #195, opened 2021-02-02
TLS 1.3 configuration file enhancements

ryru:
This PR addresses two configuration file enhancements and additions regarding TLS 1.3 DH group and signature scheme configuration.

- DH group configuration by the config specifier `ke_group`
- Approach to filter signature schemes in configuration files


src/libtls/tls_crypto.c
```diff
@@ -1484,6 +1486,56 @@ static struct {
...
+					sig == TLS_SIG_RSA_PSS_PSS_SHA384 ||
+					sig == TLS_SIG_RSA_PSS_PSS_SHA512 ||
+					sig == TLS_SIG_RSA_PKCS1_SHA256 ||
+					sig == TLS_SIG_RSA_PKCS1_SHA384 ||
+					sig == TLS_SIG_RSA_PKCS1_SHA512)
+				{
+					enumerator->destroy(enumerator);
+					return TRUE;
+				}
+			}
```
tobiasbrunner, line 1532:
```suggestion
			if (enum_from_name(tls_signature_scheme_names, token, &sig) &&
				sig == signature)
			{
				enumerator->destroy(enumerator);
				return TRUE;
			}
```

src/libtls/tls_crypto.c
```diff
@@ -2496,6 +2549,44 @@ int tls_crypto_get_supported_groups(diffie_hellman_group_t **out)
 	return count;
 }
 
+/**
+ * See header.
+ */
+int tls_crypto_get_supported_signatures(tls_version_t version, bool sign,
```
tobiasbrunner, line 2555: Not sure if the `sign` argument is really necessary. The function is only used for tests where we need both verifying and signing, so the PSS_PSS schemes should probably always get filtered. What's missing is a filter via `lib->plugins->has_feature()` (see `get_signature_algorithms()`), i.e. it's currently the potentially supported schemes and not the actually supported ones.

tobiasbrunner (changes requested):

tobiasbrunner:
Thanks, looks good. Applied the two commits to the _tls13_ branch with minor changes (one was only registering the credential set once).

I think we go with the TLS strings, as using the existing pubkey constraints would complicate the parsing, would require several additional changes (e.g. a way to convert a signature scheme to a parsable config string), and is kinda awkward (at least currently, a hash algorithm is always necessary, even for EdDSA). In the future, we maybe could add support for the constraints we use for IKE also for EAP-TLS (e.g. when configured via _remote.auth_ in the config), which is currently only supported for certificate chain validation. But that wouldn't directly involve this global option (filtering would additionally be done based on an existing `auth_config_t`), so supporting the same syntax would mainly be for consistency (and could still be added if necessary).

By the way, I noticed another issue related to ECC. While we use/accept TLS 1.2 suites with `KEY_ECDSA` even if we only have an EdDSA key available, the same is not the case if we only have X25519 available (i.e. no ECP groups). All suites that list a ECP group will be filtered even though X25519 should work with them. RFC 8422 also does not tie DH groups/curves to specific cipher suites. While it might make sense as client to select a (default) group/curve that's consistent with the strength of the other algorithms in the suite, it seems a bit too rigid when filtering/matching them.

src/libtls/tls_crypto.c
```diff
@@ -1195,6 +1200,11 @@ static void build_cipher_suite_list(private_tls_crypto_t *this)
 	{
 		filter_key_suites(this, suites, &count, KEY_ECDSA);
 	}
+	if (!this->eddsa)
+	{
+		filter_key_suites(this, suites, &count, KEY_ED25519);
+		filter_key_suites(this, suites, &count, KEY_ED448);
```
tobiasbrunner, line 1206: This won't work. There are no suites listing these key types. And if we only have an EdDSA key, the suites that could be used (those listing `KEY_ECDSA`) are already filtered at this point. I guess we could change the if statement right above to `if (!this->ecdsa && !this->eddsa)` but that doesn't really have any advantage over the current code that uses the single `ecdsa` flag. So not sure if there actually is any point in separating this further (maybe we could just rename the flag to `ecc`?).

tobiasbrunner:
I've added some commits to address the ECDH/cipher suite issue.

PR #197, opened 2021-02-15
FreeBSD 12.2

fkorotkov:
Seems it's fixing the installation. 12.2 is the new CURRENT and the packages are frequently updated under the hood so that might have been the issue.


tobiasbrunner:
Great! Thanks a lot for looking into this.

PR #215, opened 2021-02-22
Add IKE and ESP support for full-sized HMAC_SHA2 algorithms

mskalski:
Add support for IKE (proposals) and ESP (for Linux `netlink`) for following proprietary algorithms:

- `AUTH_HMAC_SHA2_256_256`
- `AUTH_HMAC_SHA2_384_384`
- `AUTH_HMAC_SHA2_512_512`

Also enables saving ESP keys for `AUTH_HMAC_SHA2_256_256` algorithm (other are too long to be supported in Wireshark).
Fixed typo in `af_alg` strongswan plugin.


tobiasbrunner:
These use identifiers from the private range (they are only defined as internal identifiers for use in TLS or PKCS#12). So I wonder what point there is to add support for them as they would only work between strongSwan instances (potentially only of the same version as the identifier values might change). The truncation also does have potential security benefits (see [RFC 4868](https://tools.ietf.org/html/rfc4868), which defines the truncated variants, for details) and the overhead without it is quite significant for ESP. If you need a 256-bit ICV, just use regular, truncated SHA-512, which generally is also faster to calculate than SHA-256 on 64-bit CPUs.

Thanks for the fix in the af-alg plugin, though. I've applied that to master.

mskalski:
Yes, I'm aware of private range of hmac-sha2-full identifiers. But these might be useful for negotiating ESP algorithms which are forced to be full-sized for some devices.

I think it could be convenient for plugins / devices using these algorithms (especially when passing to kernel for ESP/AH) to have them already defined, as code changes in this PR are rather simple.

Please note, that some other algorithms from private range (i.e `AUTH_CAMELLIA_XCBC_96`, `AUTH_HMAC_SHA1_128` `ENCR_TWOFISH_CBC` or `ENCR_SERPENT_CBC`) already have defined proposal keywords and are supported by e.g. linux `kernel_netlink` plugin.

Thanks for looking at it.

tobiasbrunner:
> But these might be useful for negotiating ESP algorithms which are forced to be full-sized for some devices.

What do you mean?

> Please note, that some other algorithms from private range (i.e `AUTH_CAMELLIA_XCBC_96`, `AUTH_HMAC_SHA1_128` `ENCR_TWOFISH_CBC` or `ENCR_SERPENT_CBC`) already have defined proposal keywords and are supported by e.g. linux `kernel_netlink` plugin.

Yes, but these are for legacy reasons (from a time where AES or the modern AEAD algorithms were not yet well established) and mostly interoperable with other FreeS/WAN forks.

mskalski:
> > But these might be useful for negotiating ESP algorithms which are forced to be full-sized for some devices.
> 
> What do you mean?

Well, I just have to cooperate with some devices (among other) using sha-256_256 ESP algorithm. Also for legacy compability reasons.

tobiasbrunner:
> Well, I just have to cooperate with some devices (among other) using sha-256_256 ESP algorithm. Also for legacy compability reasons.

What devices are this that use non-standardized algorithms for ESP? And for what purpose (there really doesn't seem to be any benefit to use the full length)? And are these the only algorithms they support? And why legacy compatibility? Which legacy IPsec implementations use these algorithms that were never defined (or is this something like the 96-bit SHA-256 Linux bug)? And also, for those use cases our private identifiers won't be useful anyway, right?

mskalski:
> > Well, I just have to cooperate with some devices (among other) using sha-256_256 ESP algorithm. Also for legacy compability reasons.
> 
> What devices are this that use non-standardized algorithms for ESP? And for what purpose (there really doesn't seem to be any benefit to use the full length)? And are these the only algorithms they support? And why legacy compatibility? Which legacy IPsec implementations use these algorithms that were never defined (or is this something like the 96-bit SHA-256 Linux bug)? And also, for those use cases our private identifiers won't be useful anyway, right?

Sorry I wasn't clear enough. These are some classified IKEv1 devices which are still in use now, but I want to support them apart of new ones implementing IKEv2 correctly. So this is (for me) the meaning of  `legacy` :-).

Yes - the algorithm `sha256_256` is the only one auth algorithm the device supports. I've read about potential security benefits and rationales for choosing specific auth lengths, but auth `sha256_256` just was done before SHA256 was standardised.

And you're right, private identifiers are not useful at all (as are not useful at all for IKEv1 because translation made inside IKEv1 encoding), but possibility of adding full-sized ESP algorithm would be really appreciated.

tobiasbrunner:
I feel really sorry for you that you have to work with such crap. I guess I can live with the changes to `keymat.c` and the _kernel-netlink_ plugin (mainly because there is currently no easy way to extend these things from a plugin). But to avoid interoperability issues, I don't want people configuring them, so no new keywords (if you need them, can define them yourself from a plugin via `lib->proposal->register_token()`). Finally, the changes in the _save-keys_ plugin seem of not much use (considering the plugin is rarely built and Wireshark, rightly so, doesn't support the algorithms). I've pushed the modified commits to the _gh215-sha2-no-trunc_ branch. Let me know if that's OK for you.

mskalski:
LGTM, And I agree with proposals, as using `register_token` is what I've already done.

I would rather keep also changes in `save_keys` plugin - as you said this plugin is mainly for development, so its inclusion does not decrease any security **more** :-). But it would help me a lot if this was in upstream code. If not - it is also OK.

Thank you very much for your help.

tobiasbrunner:
I don't like the commented out entries there (I guess Wireshark would just refuse to load the file if we'd write `ANY 384 bit ...`?). But I agree that adding the SHA-256 entry doesn't do any harm, even if Wireshark will just ignore it. I also found a bug, so there would be changes to the plugin anyway. I pushed these changes to the branch.

mskalski:
The branch `gh215-sha2-no-trunc` looks perfect :-), Thank you again

tobiasbrunner:
Thanks, merged to master.

PR #250, opened 2021-03-11
Update openssl_ec_diffie_hellman.c

wasy168:
Use openssl normal functions instead of deprecated functions

### DIFF
--- a/src/libstrongswan/plugins/openssl/openssl_ec_diffie_hellman.c
+++ b/src/libstrongswan/plugins/openssl/openssl_ec_diffie_hellman.c
@@ -100,7 +100,7 @@ static bool chunk2ecp(const EC_GROUP *group, chunk_t chunk, EVP_PKEY *key)
 	}
 
 	point = EC_POINT_new(group);
-	if (!point || !EC_POINT_set_affine_coordinates_GFp(group, point, x, y, ctx))
+	if (!point || !EC_POINT_set_affine_coordinates(group, point, x, y, ctx))
 	{
 		goto error;
 	}
@@ -164,7 +164,7 @@ static bool ecp2chunk(const EC_GROUP *group, EVP_PKEY *key, chunk_t *chunk)
 
 	ec_key = EVP_PKEY_get1_EC_KEY(key);
 	point = EC_KEY_get0_public_key(ec_key);
-	if (!point || !EC_POINT_get_affine_coordinates_GFp(group, point, x, y, ctx))
+	if (!point || !EC_POINT_get_affine_coordinates(group, point, x, y, ctx))
 	{
 		goto error;
 	}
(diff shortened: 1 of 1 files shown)


strongX509:
If I look at the `ec.h` openssl header file

https://github.com/openssl/openssl/blob/master/include/openssl/ec.h#L609

it seems the other way around:
```
int EC_POINT_set_affine_coordinates()
```
is the regular function and 
```
OSSL_DEPRECATEDIN_3_0 int EC_POINT_set_affine_coordinates_GFp()
```
will be deprecated with Openssl 3.0.

wasy168:
Yes, in version 5.9.2 are the functions xxx_GFp using in openssl_ec_diffie_hellman.c, will be better if the regular functions:

EC_POINT_set_affine_coordinates and EC_POINT_get_affine_coordinates can be used.


________________________________
Von: Andreas Steffen ***@***.***>
Gesendet: Mittwoch, 17. März 2021 12:04
An: strongswan/strongswan ***@***.***>
Cc: wasy168 ***@***.***>; Author ***@***.***>
Betreff: Re: [strongswan/strongswan] Update openssl_ec_diffie_hellman.c (#250)


If I look at the ec.h openssl header file

https://github.com/openssl/openssl/blob/master/include/openssl/ec.h#L609

it seems the other way around:

int EC_POINT_set_affine_coordinates()


is the regular function and

OSSL_DEPRECATEDIN_3_0 int EC_POINT_set_affine_coordinates_GFp()


will be deprecated with Openssl 3.0.

strongX509:
I propose the following fix:
```
diff --git a/src/libstrongswan/plugins/openssl/openssl_ec_diffie_hellman.c b/src/libstrongswan/plugins/openssl/openssl_ec_diffie_hellman.c
index 53691205d..68a517035 100644
--- a/src/libstrongswan/plugins/openssl/openssl_ec_diffie_hellman.c
+++ b/src/libstrongswan/plugins/openssl/openssl_ec_diffie_hellman.c
@@ -506,7 +506,11 @@ EC_GROUP *ec_group_new_brainpool(bp_curve *curve)
                goto failed;
        }
        G = EC_POINT_new(group);
+#ifdef OPENSSL_VERSION_NUMBER >= 0x1010100fL
+       if (!G || !EC_POINT_set_affine_coordinates(group, G, x, y, ctx))
+#else
        if (!G || !EC_POINT_set_affine_coordinates_GFp(group, G, x, y, ctx))
+#endif
        {
                goto failed;
        }
```
since the first two occurences are restricted to Openssl versions < 1.1.0 anyway and Boringssl is still on the old format:
```
#if OPENSSL_VERSION_NUMBER < 0x1010000fL || defined(OPENSSL_IS_BORINGSSL)
```

strongX509:
Committed fix https://github.com/strongswan/strongswan/commit/bd323ae6c8325812b95a0b61598a35940b06bfb1 to master

PR #327, opened 2021-04-15
github: Make it work with latest Ubuntu image (20.04)

Thermi:


tobiasbrunner:
Thanks. I've adopted the os-release idea into the linked commit.

PR #328, opened 2021-04-15
test.sh: Only run lgtm if LGTM_TOKEN is set and test is set to lgtm

Thermi:


tobiasbrunner:
Thanks, applied to master with some modifications.

Thermi:
Hmmmh. I dislike this because it means that currently no forked strongSwan repo can have completely passed test suites. That's because lgtm currently does not support forked repos.

tobiasbrunner:
> That's because lgtm currently does not support forked repos.

What do you mean? What about [this](https://lgtm.com/projects/gl/contauro-ag/strongswan/)?

> It also means it can never completeley pass any checks in a PR because the checks are taken from the forked repo

The SonarCloud and LGTM tests currently don't run for these anyway only for pushs (the LGTM test that is actually run for PRs is directly done by LGTM, not via Action).

Thermi:
That's not a fork, but a copy complete reuploaded copy of the repo on *Gitlab*. It's no tracked as a fork by Github.

Okay, so at least PRs won't error out. It's still not good.

tobiasbrunner:
I see, didn't know about the fork issue. I guess we can make it not error out.

PR #340, opened 2021-04-21
testing: Use kvm64 instead of core2duo

Thermi:
With that change you can use nested virtualization to run the test suite in a virtualized Ubuntu 20.04 (which then runs the debian based VMs that the test suite starts).

Without the changes, at the very least the "monitor" CPU feature isn't satisfied in the VM, which causes the core2duo specified VMs to not start. With kvm64 that issue does not occur.

### DIFF
--- a/testing/config/kvm/moon.xml
+++ b/testing/config/kvm/moon.xml
@@ -11,7 +11,7 @@
     <boot dev='hd'/>
   </os>
   <cpu>
-    <model fallback='allow'>core2duo</model>
+    <model fallback='allow'>kvm64</model>
     <feature policy='optional' name='aes'/>
     <feature policy='optional' name='pclmuldq'/>
   </cpu>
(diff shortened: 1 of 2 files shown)


tobiasbrunner:
Not sure if that's better as it's also deprecated according to the [QEMU docs](https://qemu-project.gitlab.io/qemu/system/qemu-cpu-models.html#other-non-recommended-x86-cpus).

Martin added this when he developed the ChaCha20/Poly1305 kernel patches (I guess to compare against hardware accelerated AES-GCM). So the goal was to use hardware acceleration for AES-GCM on these two hosts without failing on CPUs that don't provide the required extensions. I think instead of explicitly setting a specific model, we could just set the `mode` attribute to `host-passthrough` (without requesting any specific features), which should give us the best performance/feature set on modern hardware. 

I pushed such a change to the _340-kvm-cpu_ branch (also changed it for all guests). Could you try if that also works for the nested virtualization use case?

Thermi:
Works.

tobiasbrunner:
Thanks for testing.

PR #341, opened 2021-04-22
testing.conf: default all vars, so they can be overridden on command …

Thermi:
…line


tobiasbrunner:
A potential issue with this is that one of the variables might already be defined in the shell for some reason (in particular one of the shorter or more generic ones). Couldn't you use the `testing.conf.local` approach?

Thermi:
I see. I didn't because I didn't know if I had to change more stuff on the fly to get around problems in the test suites. :P

PR #343, opened 2021-04-22
ccm: destroy iv_gen on crypter creation failure

mtomaschewski:
The initialization creates a `crypter` and `iv_gen`, but checks only `crypter` creation failure and does not destroy `iv_gen` then.

### DIFF
--- a/src/libstrongswan/plugins/ccm/ccm_aead.c
+++ b/src/libstrongswan/plugins/ccm/ccm_aead.c
@@ -415,8 +415,10 @@ ccm_aead_t *ccm_aead_create(encryption_algorithm_t algo,
 		.icv_size = icv_size,
 	);
 
-	if (!this->crypter)
+	if (!this->crypter || !this->iv_gen)
 	{
+		DESTROY_IF(this->crypter);
+		DESTROY_IF(this->iv_gen);
 		free(this);
 		return NULL;
 	}


src/libstrongswan/plugins/ccm/ccm_aead.c
```diff
@@ -415,8 +415,12 @@ ccm_aead_t *ccm_aead_create(encryption_algorithm_t algo,
 		.icv_size = icv_size,
 	);
 
-	if (!this->crypter)
+	if (!this->crypter || !this->iv_gen)
 	{
+		if (this->crypter)
+			this->crypter->destroy(this->crypter);
+		if (this->iv_gen)
+			this->iv_gen->destroy(this->iv_gen);
```
tobiasbrunner, line 423: The `iv_gen` creation can't fail and please use the `DESTROY_IF` macro to destroy `iv_gen` if no crypter can be created.

mtomaschewski: It was just something that caught my eye... I've changed to use `DESTROY_IF` and updated the pull.
Even `iv_gen` creation can't fail, I'd keep the `| !this->iv_gen)` check there and call `DESTROY_IF` for both, so someone else does  not stumble over it again.

tobiasbrunner: > Even `iv_gen` creation can't fail, I'd keep the `| !this->iv_gen)` check there and call `DESTROY_IF` for both, so someone else does not stumble over it again.

The problem is that such a check could indicate to e.g. static source code analyzers that the call could fail, which would require either adding such a useless check everywhere it is used or removing it again :)

PR #344, opened 2021-04-22
testing: mkdir -p $SHAREDDIR

Thermi:
Otherwise testing fails because the dir does not exist when trying to mount it.


tobiasbrunner:
Thanks, applied to master.

PR #347, opened 2021-04-23
library: Initialize lib->leak_detective to NULL if the build is witho…

Thermi:
…ut leak detective.

Fixes at the very least a crash in the test cases when run with valgrind.

### DIFF
--- a/src/libstrongswan/library.c
+++ b/src/libstrongswan/library.c
@@ -345,6 +345,8 @@ bool library_init(char *settings, const char *namespace)
 		lib->leak_detective->set_report_cb(lib->leak_detective,
 										   report_leaks, sum_leaks, this);
 	}
+#else
+	lib->leak_detective = NULL;
 #endif /* LEAK_DETECTIVE */
 
 	pfh = printf_hook_create();


tobiasbrunner:
Hm, the complete struct is initialized to 0, so I don't really see why this would be necessary.

Thermi:
Without the change, at least one test case crashes because there's garbage in the field, so comparisons with NULL fail and other code that dereferences the invalid pointer access unmapped memory. With the change that doesn't happen anymore. I'll check the details later again.

Am May 4, 2021 2:08:30 PM UTC schrieb Tobias Brunner ***@***.***>:
>Hm, the complete struct is initialized to 0, so I don't really see why
>this would be necessary.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/strongswan/strongswan/pull/347#issuecomment-831970953

-- 
Sent from mobile

tobiasbrunner:
Yeah, would be interesting to see where/how this happens. Because we use this struct initialization (via INIT macro and without) all over the place.

PR #352, opened 2021-04-25
Fix compatibility issues with BoringSSL

deiger:


tormentous:
a few more to look...
https://github.com/strongswan/strongswan/discussions/349#discussion-3335757

deiger:
> We don't use the upstream version of BoringSSL (and currently have no plans to support it).

Which versions are you using? These changes apply also to the recent Android versions, see [here](https://android.googlesource.com/platform/external/boringssl/+/refs/heads/android11-mainline-release/src/include/openssl/evp.h#928) and [here](https://android.googlesource.com/platform/external/boringssl/+/refs/heads/android11-mainline-release/src/include/openssl/crypto.h#119) for Android 11.

tobiasbrunner:
The _ndk-static_ branch from [here](https://git.strongswan.org/?p=android-ndk-boringssl.git;a=summary).

deiger:
Wow, that's 5 years old. I'm really surprised that these were the only changes needed to in order to compile against head.
Given that there were [several CVEs](https://vulmon.com/searchpage?q=boringssl) over the years, would it make sense to update it?

tobiasbrunner:
I've currently no plans to do so. AFAIK they changed the build system, so no idea if/how this could still be integrated in our build process and Google doesn't recommend using it anyway. So I guess it be better to switch to OpenSSL at some point in the future, but that has it's own issues.

deiger:
I'm not sure what updating the build process would entail, as we have an internal build system, but the code changes would be very minimal.

tobiasbrunner:
Why do you use upstream BoringSSL (which, again, Google does not generally recommend) and not OpenSSL if you have your own build system etc.?

deiger:
I work for Google. For internal use it is strongly recommended...

tobiasbrunner:
I see :) So lets apply these fixes then. The second isn't a problem. But the first is because our fork doesn't provide it and while I could move the macro to that fork, this would break the build with older strongSwan versions as I didn't put an `#ifndef` around it. I could maybe move the macro to our BoringSSL fork to fix our build (the problem with that is that it breaks the build for older strongSwan versions because.

Just note that the _openssl_ plugin might not use newer features provided by current versions of BoringSSL as those are generally enabled via OPENSSL_VERSION. Only in some cases is there also a `OPENSSL_IS_BORINGSSL` but that's obviously only done for stuff our old fork already supports.

deiger:
Thank you!
That makes perfect sense. I'll try to go over the `OPENSSL_VERSION` checks in the code, and see if there are further features that can be enabled by the recent BoringSSL.

davidben:
BoringSSL maintainer here. (@deiger, please reach out to us when patching 3p libraries; we can help you with best practices.) There shouldn't be any need for `OPENSSL_IS_BORINGSSL_STRONGSWAN`. If projects need to support a range of BoringSSL revisions, the right way to do this is to check `BORINGSSL_API_VERSION`. So if you're worried about `EVP_PKEY_base_id`, use:

```
#if defined(OPENSSL_IS_BORINGSSL) && BORINGSSL_API_VERSION < 10
#define EVP_PKEY_base_id(p) EVP_PKEY_type(p->type)
#endif
```

In addition, this PR is missing at least one other necessary change: this ifdef should be dropped and instead call `DH_set_length`. Just as in OpenSSL, reaching into BoringSSL private structures isn't supported and will fail to compile in later revisions. If you want to handle older versions, use `defined(OPENSSL_IS_BORINGSSL) && BORINGSSL_API_VERSION < 11` for that one.
https://github.com/strongswan/strongswan/blob/master/src/libstrongswan/plugins/openssl/openssl_diffie_hellman.c#L171

That said, note we don't support arbitrary old revisions of BoringSSL. The expectation is that all `BORINGSSL_API_VERSION` checks in downstream logic is temporary.

Regarding `OPENSSL_VERSION_NUMBER`, please leave all that alone. We intentionally adjust `OPENSSL_VERSION_NUMBER` to match the OpenSSL APIs we are roughly targeting. Second-guessing that adds extra work for you *and* us at the same time.

CodeFetch:
@deiger Why is Google trying to enforce boringssl usage recently? Your development guidelines are not open-source friendly. You often introduce breaking changes to API/ABI compatibility and don't push your changes to other open-source projects upstream. In the best case you send an email with a lot of spaghetti code and don't reply to changes requested from maintainers.

Citing from https://boringssl.googlesource.com/boringssl/:
`Although BoringSSL is an open source project, it is not intended for general use, as OpenSSL is. We don't recommend that third parties depend upon it. Doing so is likely to be frustrating because there are no guarantees of API or ABI stability.`

Indeed nearly every encounter with Google developers I have witnessed so far ended in long discussions with no results. So maybe Google should rethink their policies and contribute to established open-source projects or otherwise stop bugging other projects with their unreliable hard-forks to save us all from this frustration. Open-source developers are not your slaves.

davidben:
I think there may be some confusion here. We use BoringSSL internally, but we usually *don't* want open source developers to carry BoringSSL ifdefs. We prefer to manage the compatibility issues ourselves rather than burden the external project. Indeed the request is to *remove* unnecessary compatibility bits so that it runs the same code as in OpenSSL. Supporting multiple versions of OpenSSL is enough of ifdef soup as it is, and we certainly don't want to make it worse!

Sometimes the project does want to support BoringSSL, in which case we figure out what works best for everyone. Here, it sounds like strongswan wants to support some particular old revision, so the suggestion is to use `BORINGSSL_API_VERSION` to detect that old revision.

tobiasbrunner:
@davidben Thanks for your comment. Really appreciate the insight. Our version of BoringSSL is _really_ old, though, so it doesn't actually define `BORINGSSL_API_VERSION` yet. But I guess we could just use that fact in an ifdef too if necessary.

I've pushed a commit to the _352-boringssl_ branch that removes many `OPENSSL_IS_BORINGSSL` checks (where there already are `OPENSSL_VERSION` checks) and adds `BORINGSSL_API_VERSION` checks to two of them (as suggested by you). @deiger could you maybe test if this works with the BoringSSL release you were using?

iamsais:
@tobiasbrunner @davidben I'm trying to update BoringSSL (latest) for Strongswan 5.9.2 (Android), in thought of providing FIPS Compliance.

After reading your comment https://wiki.strongswan.org/issues/2979#note-1 (not recommended update Boring SSL), I have stopped updating BoringSSL. The current modified ndk-boringssl is old, what is your recommendation to update OpenSsl and removing weaker ciphers (with FIPS Compliant)?

tobiasbrunner:
> The current modified ndk-boringssl is old, what is your recommendation to update OpenSsl and removing weaker ciphers (with FIPS Compliant)?

Just use OpenSSL (if FIPS compliance is your goal you need to follow strict procedures anyway).

tobiasbrunner:
That doesn't sound right. Did you recompile against the newer version using matching header files? (The headers definitely have to match and you also can't use a binary version of the plugin built against a different version of OpenSSL.)

tobiasbrunner:
> Which version of OpenSSL is compatible with the OpenSSL plugin.

Basically all of them.

> Can we use the libcrypto.so as PreShared Library?

Not in an Android app as that will most likely conflict with the BoringSSL library shipped by Android. You have to link it statically into the plugin. Have a look at the [build recipe for F-Droid](https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/org.strongswan.android.yml).

PR #385, opened 2021-05-07
testing: Replace qemu-kvm with qemu-system-x86_64 (because qemu-kvm d…

Thermi:
…oes not exist on non-debian based distros)

Also, the "kvm" script in PATH is only a compatibility thing, says the man page:
`man kvm`:
```
The kvm wrapper script is used to provide compatibility with old qemu-kvm package which has been merged into qemu as of version 1.3.
```


tobiasbrunner:
Thanks, applied to master.

PR #393, opened 2021-05-10
Vendor ids

Thermi:


tobiasbrunner:
> because now there's effective code duplication in the IKEv1 and IKEv2 vendor-id handling code.

Can't fully avoid that between the IKE versions (hopefully we can eliminate the IKEv1 code sometime in the future).

> The Heartbeat VIDs, like all VIDs, are from https://github.com/royhills/ike-scan/blob/master/ike-vendor-ids. Googling for "Heartbeat_Notify" gives hits for openswan and libreswan source, as well as Cisco device logs [here](https://community.cisco.com/t5/routing/ios-xe-16-06-04-and-nim-4g-lte-interface-not-obtaining-ip/td-p/4001176), so at the very least Cisco seems to use them.

Now that you mention it, the old pluto daemon had that one too (but seemed to be from Juniper Netscreen). So this is probably  something very old (maybe a precursor to DPD, which was a relatively late extension for IKEv1 and the pluto README talked about heartbeats). Not sure if there is any point in adding them (definitely not for IKEv2). By the way, the "KAME/racoon" VID is also useless for IKEv2 as racoon only support IKEv1.

> I do not know if anything else uses what corresponds to the effective prefix for the Sonicwall VIDs, so I'd rather stick close to the original source for them, than apply the prefix.

Sure, but I doubt that a "Sonicwall 1" supported IKEv2. The earliest IKEv2 log I found was with "Sonicwall 7" and that was posted 8 years ago, so I'm going to remove the earlier ones from the IKEv2 list.

> We know that a lot of proprietary IKEv2 implementations are just revised IKEv1 ones.

Unfortunately, true.

> They do not break anything. Thus I think it's okay. If there are changes desired later so they are identified in more detail, they can be done.

Sure, they don't break anything. But I'm also not a fan of adding clutter just for the sake of it. In the future, I'd rather add additional ones when they are actually observed in the wild (or there might even be a reason to recognize a particular one).

tobiasbrunner:
e1664238565616f0a69962722c5b03389e04d813 fixes an issue with IKEv1 NAT VIDs.

Thermi:
Seen. I thought I had double checked the order of the struct members when writing it and even changed it once because I thought I had it wrong, and in the end still managed to screw it up.

PR #434, opened 2021-06-06
Expose the dh_group on add_sa

deiger:
This would enable add_sa implementations to validate and expose externally the used DH group.

### DIFF
--- a/src/libcharon/kernel/kernel_ipsec.h
+++ b/src/libcharon/kernel/kernel_ipsec.h
@@ -83,6 +83,8 @@ struct kernel_ipsec_add_sa_t {
 	uint16_t int_alg;
 	/** Integrity protection key */
 	chunk_t int_key;
+	/** Diffie Hellman group */
+	uint16_t dh_group;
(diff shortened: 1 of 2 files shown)


tobiasbrunner:
What purpose does this have?

deiger:
We'd like to expose the selected DH group to the user. Currently that data is only available in the log.

tobiasbrunner:
> We'd like to expose the selected DH group to the user.

In the kernel layer?

> Currently that data is only available in the log.

Not sure what you are talking about exactly,  because that information is stored in the CHILD_SA object, so it could be retrieved from there (it's what vici etc. do to report this to users - why not use that interfaces?), it can also be retrieved via hooks (e.g. `child_updown` or `child[_derived]_keys`). In general, it doesn't provide any useful information to kernel plugins. Also note that the DH group is optional (in particular during IKE_AUTH where there never is one) and that `MODP_NULL` isn't the right value to indicate that.

deiger:
OK, thanks.

PR #505, opened 2021-07-15
gitignore: Add /nbproject

Thermi:
/nbproject is the Apache Netbeans IDE project file directory. Lots of prople probably use Netbeans with the (simple) C plugin to work on strongSwan with it. It makes sense to add it into the .gitignore file here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@ coverage/
 /*.files
 /*.includes
 test-driver
+nbproject/


tobiasbrunner:
Would maybe `nbproject/` be a better choice, in case somebody opens a subdir (e.g. `src/frontends/gnome`) in NetBeans?

> Lots of prople probably use Netbeans with the (simple) C plugin to work on strongSwan with it.

Never heard of somebody doing that, but OK :) (IDEs I know people have used with strongSwan in the past are Eclipse, CLion and Qt Creator, I guess I have to add NetBeans to that list.)

Thermi:
Done.

tobiasbrunner:
Thanks, applied to master.

PR #541, opened 2021-08-05
Support for trapping and restarting at the same time

Thermi:
Needed to tell strongSwan to trap and start a connection when the daemon starts.
Otherwise one needs to code that manually using a script and that sucks.

### DIFF
--- a/src/libcharon/config/child_cfg.c
+++ b/src/libcharon/config/child_cfg.c
@@ -22,10 +22,11 @@
 
 #include <daemon.h>
 
-ENUM(action_names, ACTION_NONE, ACTION_RESTART,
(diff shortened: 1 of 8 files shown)


tobiasbrunner:
Sorry, I've currently no time for a proper review. But this seems like a good recipe to create duplicate CHILD_SAs (similar to combining trap policies with `dpd_action=restart` or reauthentication), at least until we have something like the patch in the _child-sa-dedup_ branch applied (which is more of a workaround).

I also don't like the name of the vici keyword (maybe `trap-start` or `trap+start` with a `...restart` alias).

Some commits could probably also be squashed (e.g. documenting and adding the keyword(s)).

Thermi:
Not needed anymore because we now have keywords from it in the latest release(s).

/close

PR #580, opened 2021-09-03
Usage of agreed PRF function in pubkey auth

steweg:
This patch adjust pubkey authentication to use agreed PRF function for RSA KEY types.

### DIFF
--- a/src/libcharon/sa/ikev2/authenticators/pubkey_authenticator.c
+++ b/src/libcharon/sa/ikev2/authenticators/pubkey_authenticator.c
@@ -439,9 +439,31 @@ static status_t sign_classic(private_pubkey_authenticator_t *this,
 	switch (private->get_type(private))
 	{
 		case KEY_RSA:
-			scheme = SIGN_RSA_EMSA_PKCS1_SHA1;
+		{
+			chunk_t skd = chunk_empty;
+			keymat_v2_t *keymat = (keymat_v2_t*)this->ike_sa->get_keymat(this->ike_sa);
+			pseudo_random_function_t prf_alg = keymat->get_skd(keymat, &skd);
 			auth_method = AUTH_RSA;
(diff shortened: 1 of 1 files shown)


steweg:
@tobiasbrunner: is there anything else needed to get this one in? It's needed to get support of EAP AnyConnect

tobiasbrunner:
The problem is that this is not based on any standard. So this will cause interoperability issues with other implementations (even older strongSwan versions).

There was an attempt to make classic RSA signatures more flexible and configurable (see the last three commits in the _sig-hash-cfg_ branch), but that was never completed (only the _gmp_ plugin was changed) and was later superseded by a proper extension for IKEv2. [RFC 7427](https://datatracker.ietf.org/doc/html/rfc7427) solves the problem of using stronger hashes and signature schemes with RSA in a proper and interoperable way. That's what Cisco should be implementing since 7 years (strongSwan supports it since 5.3.0, which was released in March of 2015, although support for RSA-PSS was only added with 5.6.1 in 2017).

Note that I have no interest in EAP AnyConnect due to its proprietary nature, the consequential lack of documentation and its potential legal implications.

steweg:
I agree that AnyConnect is proprietary mechanism, but at the same time, there is currently no alternative client. So this can be a benefit for strongswan. So I see that we have two main options. Option A: I will cancel all my PRs, and simply keep everything in my own fork. Or option B: We will proceed with getting EAP AnyConnect to official repo, and to do that I will adjust the code in such way that it will be backward compatible.

So which way do you want?

tobiasbrunner:
> I agree that AnyConnect is proprietary mechanism, but at the same time, there is currently no alternative client.

That's how Cisco wants it, if they didn't, they would have standardized whatever extensions they employ.

> So which way do you want?

It has to be option A for the time being, sorry.

steweg:

PR #581, opened 2021-09-03
Refactor EAP vendor from uint32_t to pen_t

steweg:


steweg:
Is there anything I can do to actually get this pull request merged?

tobiasbrunner:
I don't think this change is correct. EAP types will be vendor-specific if there is a vendor ID. Printing names for those would rather require changes like the ones in the (very old, but never merged) _eap-vendor-names_ branch.

Also note that if compiled with `-fshort-enums`, `eap_vendor_t` wouldn't be able to hold all possible 24-bit PEN values (I suppose that's similarly true for `eap_type_t`, to some degree, as extended types are 32-bit, although many vendor-specific types might be < 65536 or even < 256). However, there already is a `pen_t` enum type for PENs, that could possibly be used for the EAP vendor IDs throughout the codebase (it even has an `enum_name_t` that could be used to print them). I've pushed such a change to the _581-eap-vendor_ branch.

steweg:
Ok, so I will cancel this one. But can you please check the others that I have opened quite some time ago, as in the end they are prerequisites for the final patches to support EAP AnyConnect

PR #612, opened 2021-09-23
Fix of credential manager

steweg:
This fixes the issue, when the same key is used for client
connection and at the same time it is received from other client
during authentication on server connection


steweg:

PR #613, opened 2021-09-23
Extend IKE_SA of other OCTET ID

steweg:
This enhancement allows various pubkey and eap authenticators to use different identification for keymat octets calculations then the one that is from received certificate


steweg:

PR #632, opened 2021-09-28
vici: add asynchronous session in python module

TheDevOpsExpert:
Add a new Python 3.6 package allowing vici to send asynchronous
requests. This feature is only available in Python 3 so we add a
condition in the setup.py in order to not break the Python 2 package.


Thermi:
You missed the opportunity to call it `avinci`

TheDevOpsExpert:
Hello,
Can someone take a look to this PR.
We are using this code in production and it works great.

Thanks

ismaillatona:
It's been 4 months. Looks like you could make a fork project. I'd suggest you name it `Aviici` ;)

PR #779, opened 2021-11-23
feat: Improved simplified Chinese Translation/ 完善了简体中文翻译

275873859:


tobiasbrunner:
Looks good, thanks. I've pushed this commit to master, but it might be a while until the next Android release.

PR #832, opened 2021-12-30
conf: fix typo (plugin -> plugins)

Thermi:


tobiasbrunner:
Thanks, applied to master.

PR #868, opened 2022-01-31
proposal: Add ESN transforms to default ESP AEAD proposal

martinwilli:
The commit mentioned below adds an AES-GCM default proposal for ESP. That
proposal does not include any ESN or non-ESN transform to indicate if extended
sequence numbers are supported.

A standards-compliant peer will include one or more ESN support transforms,
and will be unable to select this proposal due to a proposal mismatch.

Fix the default CGM proposal by adding both ESN and NOESN transforms. Given
that ESN is supported in Linux for more than ten years now, it probably
makes sense to indicate ESN support for use with AES-GCM.

Fixes: c7bef954eec6 ("proposal: Add AES-GCM to the ESP default AEAD proposal")


tobiasbrunner (commented):
Thanks.  I'm not so sure about enabling ESN by default, though. While it's probably safe to assume ESN is usable on Linux, FreeBSD only added support relatively recently (end of 2019) and macOS might still not support it, neither does _libipsec_ (on Android we use hardcoded proposals, but not in other setups via _kernel-libipsec_). And a kernel feature wouldn't help as this file is in libstrongswan. So it's probably safer to only send _noesn_.

Typo in the commit message: "CGM proposal" -> "GCM proposal" (or "AEAD proposal").

martinwilli:
Thanks, Tobias. Yes, makes sense, I've updated the commit accordingly.

tobiasbrunner:
Thanks, applied to master.

PR #1041, opened 2022-05-07
ikev1: Remove outbound policy of rekeyed CHILD_SA

leonshaw:
Remove outbound policy of rekeyed CHILD_SA since only one policy is
valid. Otherwise, during update-SA job (when NAT mapping changed),
CHILD_SA are updated and installed one by one, leaving a window where
old SAs are being used. There are also circumstances where the new SA is
not processed last.

### DIFF
--- a/src/libcharon/sa/ikev1/tasks/quick_mode.c
+++ b/src/libcharon/sa/ikev1/tasks/quick_mode.c
@@ -411,6 +411,8 @@ static bool install(private_quick_mode_t *this)
 		/* rekeyed CHILD_SAs stay installed until they expire or are deleted
 		 * by the other peer */
 		old->set_state(old, CHILD_REKEYED);
+		/* remove from outbound since only one outbound policy is valid */
+		old->remove_outbound(old);
 		/* as initiator we delete the CHILD_SA if configured to do so */
 		if (this->initiator && this->delete)
 		{


tobiasbrunner:
Thanks. Probably makes sense to remove the old outbound SA. We really only keep a rekeyed CHILD_SA around to process inbound packets in case the other peer decides to continue to use it until it expires.

> There are also circumstances where the new SA is not processed last.

What does that mean exactly? And what are these circumstances?

Also, why are you still using IKEv1?

tobiasbrunner:
> I see the policy of the new CHILD_SA overwritten by the old one. Looks like the order of CHILD_SA went wrong.

Interesting. Looks like that's caused by this: https://github.com/strongswan/strongswan/blob/7022fdc72e51d8768a5b1772bd46128c864ea7a1/src/libcharon/processing/jobs/adopt_children_job.c#L215-L219

That should be `remove_first()` instead, so the older SAs are added first to the new IKE_SA. I've pushed a fix to the _1041-ikev1-remove-rekeyed_ branch, which also contains your change.

> We want to support connecting by username/password from OS built-in VPN clients, without configuring certificates.

You mean with PSK? Please [don't do that for roadwarrior connections](https://docs.strongswan.org/docs/5.9/support/faq.html#_pre_shared_key_authentication). If the server certificate is issued by a trusted third-party CA (e.g. Let's Encrypt), you won't need to configure/install any certificates on the clients. And even if not, this shouldn't be a reason to continue to use a severely inferior protocol (especially for mobile clients, see e.g. [MOBIKE](https://docs.strongswan.org/docs/5.9/features/mobike.html)).

leonshaw:
> That should be `remove_first()` instead, so the older SAs are added first to the new IKE_SA. I've pushed a fix to the _1041-ikev1-remove-rekeyed_ branch, which also contains your change.

Great.

> You mean with PSK? Please [don't do that for roadwarrior connections]

Yes, PSK. We will consider getting a cert from trusted CA. Thanks!

PR #1044, opened 2022-05-09
proposal: Add explicit break in AUTH_HMAC_MD5_96 case

simondeziel:


tobiasbrunner:
Isn't the comment enough to silence compilers?

simondeziel:
Yes but all the other `case` have the explicit `break` so I figured it might as well be added. Mostly cosmetic I agree.

tobiasbrunner:
Ah, I see, makes sense. Since I was updating some stuff around there in the last commit of the _1026-kdf-plugin-check_ branch anyway, I've added it to that commit.

PR #1095, opened 2022-06-13
Refactor `add_ecdh_features` to use dynamically allocated array

thatnealpatel:
This change aims to increase safety of code by removing use of variable length array.

### DIFF
--- a/src/libstrongswan/plugins/openssl/openssl_plugin.c
+++ b/src/libstrongswan/plugins/openssl/openssl_plugin.c
@@ -345,23 +345,29 @@ static bool ecdh_group_supported(EC_builtin_curve *curves, size_t num_curves,
 static void add_ecdh_features(plugin_feature_t *features,
 							  plugin_feature_t *to_add, int count, int *pos)
 {
-	size_t num_curves;
-	int i;
+    size_t num_curves;
+    int i;
+    EC_builtin_curve* curves = NULL;
 
-	num_curves = EC_get_builtin_curves(NULL, 0);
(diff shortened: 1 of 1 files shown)


thatnealpatel:
It looks like all recent pull requests fail the same step in CI pipeline for a seemingly identical reason. Is this a known issue with the pipeline?

I tried to reproduce this error on local, but I was unable to do so.

tobiasbrunner:
Do you have a concrete example where the current code causes problems?

thatnealpatel:
### Portability
* There exists a non-trivial set of compilation / build tools will not allow VLA and suppressing with `-Wno-vla` is not necessarily a correct "fix" (e.g. LLVM compiler) [1]

### Security
* While this is mostly reserved for user inputs (i.e. not this particular case), variable length arrays open an attack vector in memory.

Most famously recently, the Linux kernel project has put significant effort towards refactoring away from VLAs. 

### Performance
* VLAs are not as performant as dynamically allocated arrays: extra instructions are required for stack management [1]
* [open question] If limit of `num_curves` is known and can be proven, why not use this value? From what I can understand, if unsupported curves exist in a particular version of openssl (or a fork of it), then using the maximum possible still works.
* Even if `-fstack-clash-protection` is enabled, compiler probing code adds extra complexity and size [1]

### Conclusion
I realize that this somewhat trivial change is opening the door to a larger potential change and refactor (since VLAs are used in other parts of strongSwan as well). 

As always, if this change does not fit the strongSwan philosophy, then feel free to close the PR :]


[1] https://nullprogram.com/blog/2019/10/27/

PR #1113, opened 2022-06-28
Fix liboqs build

jgertm:
liboqs does not define a `_init` method.


tobiasbrunner:
Thanks. I think `_init` (and `_fini`) is a symbol that gets automatically added to shared libs by some compilers/linkers. But since relying on that is probably not ideal, checking an actually defined and exported symbol is a better idea. I've integrated the fix directly into the now rebased _ikev2-qske-multi-ke_ branch (parts of the original branch have now been merged into master).

PR #1127, opened 2022-07-06
BF-5.9.6

ferasbi:
Added full HW offload support for strongSwan 5.9.6 (like BF-5.9.0).


ferasbi:
> Added full HW offload support for strongSwan 5.9.6 (like BF-5.9.0).

delete pull request, done by mistake

PR #1187, opened 2022-08-04
configure.ac: libsystemd-daemon and libsystemd-journal do not exist

Thermi:
All functionality seems to have been merged into libsystemd since some
version of systemd. Arch Linux already patches these two lines like
this commit does. This commit would enable Arch Linux to just drop the
patch.


tobiasbrunner:
Why is there even a fallback to that? Shouldn't the `libsystemd >= 209` check avoid that?

Thermi:
Thank you, it seems we don't need the patch anymore because it builds without it now.

PR #1244, opened 2022-08-20
connmark: fix wrong plugin name in log message

Thermi:


tobiasbrunner:
Thanks, applied to master.

PR #1266, opened 2022-09-01
Add missing testing css dir to make dist content

boisletterink:
Add the css dir to the EXTRA_DIST variable in the makefile for the test
environment. This dir was missing when generating distribution tarballs.
Adding it enables successful builds of the test environment from the
dist tarballs.

### DIFF
--- a/testing/Makefile.am
+++ b/testing/Makefile.am
@@ -1,6 +1,6 @@
 EXTRA_DIST = do-tests make-testing ssh start-testing stop-testing \
-             testing.conf ssh_config config hosts images scripts tests \
-			 README
+             testing.conf ssh_config config hosts images css scripts \
+			 tests README
 
 # exclude all files ignored by Git from the tarball
 dist-hook:


boisletterink:
Note, this is my first pull request ever - tips welcome. I followed a guideline I found as best as I could :)

tobiasbrunner:
Thanks! There are three things I changed in the commit I pushed to master:

1. the indentation, as we generally use tabs (4 chars wide), which actually wasn't your doing, it was wrong before on that line
2. the prefix, we usually add a prefix to the summary line of the commit message that indicates the modified component/area (in this case `testing:`), also slightly rephrased that line
3. the `Fixes:` tag pointing to the commit that introduced the `css` dir but missed adding it to `EXTRA_DIST`

boisletterink:
Thanks for the feedback!

I just tried to clean up my own copy by removing the dist-missing-testenv-css branch, and I now see that this is visible in the StrongSwan repo. Does it matter? I'll restore it right now in case it would cause problems :/

tobiasbrunner:
Yeah, removing that branch is exactly what you should do :) The message here is just because it was associated with this PR.

PR #1271, opened 2022-09-02
kernel-netlink: increase debug level of the "querying [...]" from 2 to 3

Thermi:
When watching the output of `swanctl -l` during debugging these debug prints cause a lot of noise in the logs (level 2 for DBG_KNL still has actually useful information that we want to see in the logs) and they're not very useful. Increasing the level to DBG3 for these prints would silence that and make the logs less noisy.


Thermi:
@tobiasbrunner thoughts?

tobiasbrunner:
Sounds reasonable, particularly for the messages in `query_sa()` and `query_policy()`. The others are only used when an IKE_SA is updated due to MOBIKE, so that's rare, but there is another message on level 2, so increasing the level there too is fine. Applied to master, thanks! (I've also pushed a similar change for the _kernel-pfkey_ plugin.)

PR #1284, opened 2022-09-07
Libipsec improvements

Thermi:
This branch contains some of the changes that Martin Willi did for libipsec to constrain the queue size of libipsec with a codel queue, as well as some changes from me to fix types, a small linking issue, and implementation of acquires for libipsec.

### DIFF
--- a/src/libipsec/Makefile.am
+++ b/src/libipsec/Makefile.am
@@ -5,6 +5,7 @@ ipsec.c ipsec.h \
 esp_context.c esp_context.h \
 esp_packet.c esp_packet.h \
 ip_packet.c ip_packet.h \
+codel_queue.c codel_queue.h \
 ipsec_event_listener.h \
 ipsec_event_relay.c ipsec_event_relay.h \
 ipsec_policy.c ipsec_policy.h \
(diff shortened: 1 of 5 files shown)


tobiasbrunner:
Regarding the CoDel queue: I've asked Martin and as far as he remembers, he did some (successful) experimenting at the time,  but never any full blown tests (stress or otherwise) or formal measurements. Did you do any kind of testing/benchmarking? The patches would also need some reviewing. And it might be interesting to check [RFC 8289](https://datatracker.ietf.org/doc/html/rfc8289) to see if anything could be improved (as a later improvement, [RFC 8290](https://datatracker.ietf.org/doc/html/rfc8290) could be an option, but that would probably require quite some work).

Regarding the acquires: I've implement asynchronous relaying of acquires from libipsec to the daemon a long time ago (`libipsec-trap` branch). What's missing from that is a call in the processor. But I've wanted to implement some rate limiting first, so we don't trigger an acquire for every matching packet. I think whoever was interested in this at the time disappeared, so this was never completed.

Thermi:
I did not do any performance comparisons simply because the overhead caused by the crypto is so absurdly high in comparison. There are plans in the making to improve the performance generally and also make it more efficient but up to now having it simply not break in corner cases is the better way forward right now.

Yes, the patches need some review too. But the general acceptance is more important right now than the exact implementation.

I used your implementation I think as basis. I did not implement any rate limiting yet. I could of course do something like the kernel does. That should be a good enough example implementation to basically copy from.

There might be more improvements coming, I intend to get them all into upstream here.

PR #1303, opened 2022-09-15
testing: Add test set location config option

boisletterink:
Adds configurable test set location, making it easier to specify a limited test set or include your own without changing files tracked by the repo. In other words, you can now specify your own dir with tests to run, from testing.conf.local.

Also renames the variable containing this location in the do-tests script to something that may be more descriptive.


tobiasbrunner:
I've not yet looked at the change, but you can add custom tests in the `testing/tests/local` directory (it is ignored by Git). And then just run `testing/do-tests local` to only run those.

PR #1329, opened 2022-10-08
swanctl.conf: better describe if_id_in and if_id_out options, their v…

Thermi:
…alid values, and their exact behavior

Up to now the valid values and the exact behavior was not documented. This commit adds it.

### DIFF
--- a/src/swanctl/swanctl.opt
+++ b/src/swanctl/swanctl.opt
@@ -296,16 +296,18 @@ connections.<conn>.pools =
 	from either the **pools** section or an external pool.
 
 connections.<conn>.if_id_in = 0
-	Default inbound XFRM interface ID for children.
-
-	XFRM interface ID set on inbound policies/SA, can be overridden by child
-	config, see there for details.
+	XFRM interface ID set on inbound policies/SA, can be overridden by child config,
+	see connections.<conn>.children.<child>.if_id_in for valid values,
+	behaviour, and other details.
+	When the ID is set to **%unique** or **%unique-dir** the same if_id is
+	set for the inbound policies for all CHILD_SAs managed by the particular IKE_SA.
 
 connections.<conn>.if_id_out = 0
-	Default outbound XFRM interface ID for children.
-
-	XFRM interface ID set on outbound policies/SA, can be overridden by child
-	config, see there for details.
+	XFRM interface ID set on inbound policies/SA, can be overridden by child config,
+	see connections.<conn>.children.<child>.if_id_out for valid values,
+	behaviour, and other details.
+	When the ID is set to **%unique** or **%unique-dir** the same if_id is
+	set for the outbound policies for all CHILD_SAs managed by the particular IKE_SA.
 
 connections.<conn>.mediation = no
 	Whether this connection is a mediation connection.
@@ -999,6 +1001,9 @@ connections.<conn>.children.<child>.if_id_in = 0
 	ID. The special value _%unique_ sets a unique interface ID on each CHILD_SA
 	instance, beyond that the value _%unique-dir_ assigns a different unique
 	interface ID for each CHILD_SA direction (in/out).
+	This value can be hexadecimal (start with 0x), decimal (no special prefix), or octal (start with 0).
+	The conversion of the numerical values is done using the **strtoul** function thus all quirks and
+	implementation details of it apply here. The if_id is a 64 bit value.
 
 connections.<conn>.children.<child>.if_id_out = 0
 	Outbound XFRM interface ID.
@@ -1008,6 +1013,10 @@ connections.<conn>.children.<child>.if_id_out = 0
 	ID. The special value _%unique_ sets a unique interface ID on each CHILD_SA
 	instance, beyond that the value _%unique-dir_ assigns a different unique
 	interface ID for each CHILD_SA direction (in/out).
+	This value can be hexadecimal (start with 0x), decimal (no special prefix), or octal (start with 0).
+	The conversion of the numerical values is done using the **strtoul** function thus all quirks and
+	implementation details of it apply here. The if_id is a 64 bit value.
+
 
 	The daemon will not install routes for CHILD_SAs that have this option set.
 


Thermi:
@tobiasbrunner thoughts?

src/swanctl/swanctl.opt
```diff
@@ -296,16 +296,18 @@ connections.<conn>.pools =
...
 connections.<conn>.if_id_in = 0
-	Default inbound XFRM interface ID for children.
-
-	XFRM interface ID set on inbound policies/SA, can be overridden by child
-	config, see there for details.
+	XFRM interface ID set on inbound policies/SA, can be overridden by child config,
+	see connections.<conn>.children.<child>.if_id_in for valid values,
+	behaviour, and other details.
+	When the ID is set to **%unique** or **%unique-dir** the same if_id is
+	set for the inbound policies for all CHILD_SAs managed by the particular IKE_SA.
```
tobiasbrunner, line 303: That's just the general behavior of this option, that's already described above, it has nothing to do with `%unique` or `%unique-dir`. Or what's your point here.

src/swanctl/swanctl.opt
```diff
@@ -999,6 +1001,9 @@ connections.<conn>.children.<child>.if_id_in = 0
 	ID. The special value _%unique_ sets a unique interface ID on each CHILD_SA
 	instance, beyond that the value _%unique-dir_ assigns a different unique
 	interface ID for each CHILD_SA direction (in/out).
+	This value can be hexadecimal (start with 0x), decimal (no special prefix), or octal (start with 0).
```
tobiasbrunner, line 1004: That's just how any integer values can be configured in these config files. If anything, we should probably add a general section about number/time formats to the strongswan.conf man page (there is a section about time formats in swanctl.conf, which should probably be moved to strongswan.conf, which contains all other information about the general format).

src/swanctl/swanctl.opt
```diff
@@ -999,6 +1001,9 @@ connections.<conn>.children.<child>.if_id_in = 0
...
+	The conversion of the numerical values is done using the **strtoul** function thus all quirks and
```
tobiasbrunner, line 1005: That could maybe be added to the above section. But what quirks are you referring to exactly?

src/swanctl/swanctl.opt
```diff
@@ -999,6 +1001,9 @@ connections.<conn>.children.<child>.if_id_in = 0
...
+	implementation details of it apply here. The if_id is a 64 bit value.
```
tobiasbrunner, line 1006: That's not true. It's an unsigned 32-bit value.

src/swanctl/swanctl.opt
Thermi: The point is that the sa,e value is assigned to all CHILD_SAs instead of the value being carried over to the CHILD_SAs verbatim (them all getting %unique or %unique-dir).

src/swanctl/swanctl.opt
Thermi: Yes, taht would be very very useful.

src/swanctl/swanctl.opt
Thermi: Whatever the documentation for strtoul does on the system strongSwan is running. It can be different on 'BSD and Linux, as the man page for it implies. That's why that's noted here.

src/swanctl/swanctl.opt
tobiasbrunner: Ah, I see what you mean. Yes, a unique ID (or two) is currently allocated for each IKE_SA that's based on such a config, which is then used for all CHILD_SAs of a particular IKE_SA (unless the child config overrides it). I've pushed text that describes the behavior to the _1329-number-formats_ branch.

src/swanctl/swanctl.opt
tobiasbrunner: I've pushed a commit with some general information about the format to the _1329-number-formats_ branch (note that we explicitly don't allow octal format to avoid confusion, that `0` prefix was a really weird idea).

src/swanctl/swanctl.opt
tobiasbrunner: Since it's a standard C function, I don't think there are major differences between platforms (the only note regarding BSD I found in the man page I have here is that `strtouq()` is provided there, but that's obviously not relevant as we don't call that function). There could be implementation-specific differences that occur in non-C or -POSIX locales, but that mainly broadens the accepted input (e.g. regarding thousands separators), so that shouldn't matter much either.

src/swanctl/swanctl.opt
Thermi: Thank you, looks good.

src/swanctl/swanctl.opt
Thermi: It'd be good if the locale dependence was documented so if it ever happens to be a problem, the man page mentions it.

src/swanctl/swanctl.opt
Thermi: Well could we use o instead? (Just to accept Octal too. Or maybe explicitely mention that octal is not possible because it's still a common number format for some things)

tobiasbrunner: I think it's fine documenting just the supported formats (and decimal and hex should cover practically all use cases). The only place I ever encountered octal numbers is `chmod` (although, considering the range of supported values, they could also be hex). Never related to IKE, IPsec or networking. Do you have an example?

Thermi: No, it's just for completeness. To make sure it's evident, because ineviteably that question will come up if we don't document that it isn't supported.

tobiasbrunner: I don't think that's true. It hasn't been supported thus far and nobody ever asked for it (probably because it really isn't common for these use cases). Also, explicitly documenting stuff that isn't supported is tricky because there are always people who read that incorrectly (e.g. miss the _not_), and then assume something is supported because they saw the keyword they expected (and they rarely bother to re-read the documentation if something doesn't work, but immediately complain). So I feel it's better to not bring up octal numbers at all.

Thermi: I agree with you on the point with people not reading texts correctly. I guess it's better to leave it out, as you proposed.

PR #1414, opened 2022-11-14
Fixed apply_jitter function in case of jitter bigger than rekey value

d06alexandrov:
`uint64_t apply_jitter(uint64_t rekey, uint64_t jitter)` function could return values greater than rekey if jitter is greater than rekey.
 
Return value (`rekey - jitter * (random() / (RAND_MAX + 1.0))`) must not be less than zero (because function returns unsigned value), so quick fix is to limit jitter by rekey.

### DIFF
--- a/src/libcharon/config/child_cfg.c
+++ b/src/libcharon/config/child_cfg.c
@@ -435,6 +435,10 @@ static uint64_t apply_jitter(uint64_t rekey, uint64_t jitter)
 		return rekey;
 	}
 	jitter = (jitter == UINT64_MAX) ? jitter : jitter + 1;
+	if (jitter > rekey)
+	{
+		jitter = rekey;
+	}
 	return rekey - jitter * (random() / (RAND_MAX + 1.0));
 }
 #define APPLY_JITTER(l) l.rekey = apply_jitter(l.rekey, l.jitter)


src/libcharon/config/child_cfg.c
```diff
@@ -435,6 +435,10 @@ static uint64_t apply_jitter(uint64_t rekey, uint64_t jitter)
 		return rekey;
 	}
 	jitter = (jitter == UINT64_MAX) ? jitter : jitter + 1;
+	if (jitter > rekey)
+	{
+		jitter = rekey;
+	}
 	return rekey - jitter * (random() / (RAND_MAX + 1.0));
```
tobiasbrunner, line 442:
```suggestion
	return rekey - min(jitter, rekey) * (random() / (RAND_MAX + 1.0));
```

tobiasbrunner (changes requested):
Thanks. Could be simplified a bit.

src/libcharon/config/child_cfg.c
d06alexandrov: Now I'm not sure that those solutions are correct. Such jitter function could return 0 if after subtract we'll get floating point number between 0 and 1 (e.g., in case of random() = RAND_MAX). And "0" here is greater than non-zero rekey, because 0 means that rekey is disabled.
To prevent such problem we have to cast subtrahend to int.
```suggestion
	return rekey - (int)(min(jitter, rekey) * (random() / (RAND_MAX + 1.0)));
```

tobiasbrunner: While I agree that this is technically possible, and I don't oppose fixing it, I wouldn't expect this to be an issue in practice, for two reasons:

1. It's only a problem if `jitter >= rekey`, which is a misconfiguration to begin with. The jitter value should be nowhere near the rekey value or even exceed it (i.e. even your original fix should never actually come into play).
2.  For usual rekey values, it's extremely rare that the subtraction will actually result in 0. Chances are basically `1/rekey`, so for the default `rekey_time` for CHILD_SAs of 3600 seconds that's roughly `0.03%`.

And while returning 1 is probably better than returning 0 and disabling rekeying, it's pretty much useless as it will immediately trigger a rekeying (again confirming 1. above).

Also, it should probably be a cast to `uint64_t` (`rekey_bytes` could be a large number):
```suggestion
	return rekey - (uint64_t)(min(jitter, rekey) * (random() / (RAND_MAX + 1.0)));
```

tobiasbrunner: Ah, I actually wondered how you triggered this issue. The _vici_ plugin currently assumes that the main config option is `rekey_*` and derives defaults for the other two from that (see e.g. [here](https://docs.strongswan.org/docs/5.9/config/rekeying.html#_settings_2)). It doesn't change `rekey_*` if any of the others have been modified manually. I suppose we could change that, see the possible fix in the _1414-vici-rekey_ branch (also includes a squashed version of your fix).

PR #1416, opened 2022-11-14
Python vici client event listen timeout

martinwilli:
Add an optional timeout parameter to `Session.listen()`, so a Thread can periodically do other things, such as checking for cancellation. Add some new unit-tests to check the old and new behavior. While at it, drop support for Python 2.


tobiasbrunner (approved):
LGTM, thanks!

PR #1428, opened 2022-11-20
github: add CIFuzz integration

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where strongSwan already runs. CIFuzz can help detect catch regressions and fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'strongswan'
+        dry-run: false
+        language: c++
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'strongswan'
+        fuzz-seconds: 300
+        dry-run: false
+        language: c++
+    - name: Upload Crash
+      uses: actions/upload-artifact@v3
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: artifacts
+        path: ./out/artifacts


tobiasbrunner:
Interesting idea, thanks.

I tried to reproduce crashes by reverting commits for issues that OSS-Fuzz found previously. Unfortunately, that didn't work (see the logs [here](https://github.com/strongswan/strongswan/actions/workflows/cifuzz.yml)). In particular the last one is weird because there is at least one input in our seed corpus that explicitly triggers it (see [the failure here](https://github.com/strongswan/strongswan/actions/runs/3594098330/jobs/6051932170)). Some of the inputs that crashed are stored in a separate `certs-crash` directory (which contains another input that triggers this crash) that the [build.sh](https://github.com/google/oss-fuzz/blob/master/projects/strongswan/build.sh) script currently doesn't include, but this particular one actually is in the regular seed corpus. Any idea why that doesn't lead to a fail on CIFuzz?

Also, with or without bugs, the logs always show lines similar to these for all five fuzz targets:

```
INFO - Fuzzer: fuzz_certs. Detected bug.
INFO - Trying to reproduce crash using: /tmp/tmp1_76pyx6/tmpsudcyv32.
INFO - Reproduce command returned: 0. Not reproducible on /github/workspace/build-out/fuzz_certs.
INFO - Crash is not reproducible.
```

Do you know if that's to be expected?

arthurscchan:
The error you see above is marked as not reproducible, which means that the issue occurred only once and the fuzzer is not even able to reproduce it. These kinds of unreproducible issues are sometimes related to instability or random timeout in the fuzzers, where a single payload fails to crash or trigger the issue again. I have not been able to trace this now, as the CI run record is quite old. However, if new non-reproducible issues arise, I can help investigate and look into making the fuzzer more stable.

arthurscchan:
I think this PR can be closed in favour of #2993.

DavidKorczynski:
Closing in favour of https://github.com/strongswan/strongswan/pull/2993

PR #1462, opened 2022-12-15
Add support for full HW offload

ferasbi:
Added support for full HW offload since now IPsec full HW offload has been accepted by the upstream linux kernel.
This patch allows users to configure full HW offload by setting the hw_offload parameter to "full" or "auto" in the swanctl.conf files. The configuration will work if both the kernel and the device support it ("auto" fallbacks to crypto offload if full doesn't work, and if crypto offload doesn't work then it doesn't configure hw offload).
Also changed HW_OFFLOAD_YES to HW_OFFLOAD_CRYPTO to distinguish from HW_OFFLOAD_FULL, and added "crypto" option to the hw_offload parameter which is the same as "yes" was previously (kept "yes" option as legacy for backward compatibility).

Signed-off-by: Feras Bisharat <fbisharat@nvidia.com>

### DIFF
--- a/src/include/linux/xfrm.h
+++ b/src/include/linux/xfrm.h
@@ -503,6 +503,7 @@ struct xfrm_user_offload {
 };
 #define XFRM_OFFLOAD_IPV6	1
 #define XFRM_OFFLOAD_INBOUND	2
+#define XFRM_OFFLOAD_FULL	4
 
 #ifndef __KERNEL__
 /* backwards compatibility for userspace */
--- a/src/libcharon/plugins/kernel_netlink/kernel_netlink_ipsec.c
+++ b/src/libcharon/plugins/kernel_netlink/kernel_netlink_ipsec.c
@@ -1523,17 +1523,19 @@ static bool netlink_detect_offload(const char *ifname)
 #endif
 
 /**
- * There are 3 HW offload configuration values:
- * 1. HW_OFFLOAD_NO   : Do not configure HW offload.
- * 2. HW_OFFLOAD_YES  : Configure HW offload.
- *                      Fail SA addition if offload is not supported.
- * 3. HW_OFFLOAD_AUTO : Configure HW offload if supported by the kernel
- *                      and device.
- *                      Do not fail SA addition otherwise.
+ * There are 4 HW offload configuration values:
+ * 1. HW_OFFLOAD_NO   	: Do not configure any HW offload.
+ * 2. HW_OFFLOAD_CRYPTO	: Configure crypto HW offload.
+ *                        Fail SA addition if crypto HW offload is not supported.
+ * 3. HW_OFFLOAD_FULL	: Configure full HW offload.
+ * 			  Fail SA addition if full HW offload is not supported.
+ * 4. HW_OFFLOAD_AUTO 	: Configure full HW offload if supported by the kernel and device.
+ * 			  If not, configure crypto HW offload if supported by the kernel and device.
+ *                        Do not fail SA addition if offload is not supported.
  */
 static bool config_hw_offload(kernel_ipsec_sa_id_t *id,
 							  kernel_ipsec_add_sa_t *data, struct nlmsghdr *hdr,
-							  int buflen)
+							  int buflen, bool hw_offload_full)
 {
 	host_t *local = data->inbound ? id->dst : id->src;
 	struct xfrm_user_offload *offload;
@@ -1546,7 +1548,7 @@ static bool config_hw_offload(kernel_ipsec_sa_id_t *id,
 		return TRUE;
 	}
 
-	hw_offload_yes = (data->hw_offload == HW_OFFLOAD_YES);
+	hw_offload_yes = (data->hw_offload == HW_OFFLOAD_CRYPTO) || (data->hw_offload == HW_OFFLOAD_FULL);
 
 	if (!charon->kernel->get_interface(charon->kernel, local, &ifname))
 	{
@@ -1571,6 +1573,12 @@ static bool config_hw_offload(kernel_ipsec_sa_id_t *id,
 	offload->ifindex = if_nametoindex(ifname);
 	offload->flags |= data->inbound ? XFRM_OFFLOAD_INBOUND : 0;
 
+	/* check if full HW offload is requested and activate it */
+	if (hw_offload_full)
+	{
+		offload->flags |= XFRM_OFFLOAD_FULL;
+	}
+
 	ret = TRUE;
 
 out:
@@ -1590,6 +1598,7 @@ METHOD(kernel_ipsec_t, add_sa, status_t,
 	uint16_t icv_size = 64, ipcomp = data->ipcomp;
 	ipsec_mode_t mode = data->mode, original_mode = data->mode;
 	traffic_selector_t *first_src_ts, *first_dst_ts;
+	bool full_hw_offload = FALSE;
 	status_t status = FAILED;
 
 	/* if IPComp is used, we install an additional IPComp SA. if the cpi is 0
@@ -2008,8 +2017,11 @@ METHOD(kernel_ipsec_t, add_sa, status_t,
 			sa->replay_window = data->replay_window;
 		}
 
+		/* try full HW offload if requested */
+		full_hw_offload = (data->hw_offload == HW_OFFLOAD_FULL) || (data->hw_offload == HW_OFFLOAD_AUTO);
+
 		DBG2(DBG_KNL, "  HW offload: %N", hw_offload_names, data->hw_offload);
-		if (!config_hw_offload(id, data, hdr, sizeof(request)))
+		if (!config_hw_offload(id, data, hdr, sizeof(request), full_hw_offload))
 		{
 			DBG1(DBG_KNL, "failed to configure HW offload");
 			goto failed;
@@ -2017,6 +2029,21 @@ METHOD(kernel_ipsec_t, add_sa, status_t,
 	}
 
 	status = this->socket_xfrm->send_ack(this->socket_xfrm, hdr);
+
+	/* if fails when full HW offload is configured and we are in auto mode
+	 * then fallback to crypto offload (disable full HW offload) and retry */
+	if (status != SUCCESS && id->proto != IPPROTO_COMP && full_hw_offload && data->hw_offload == HW_OFFLOAD_AUTO)
+	{
+		DBG1(DBG_KNL, "failed to configure full HW offload, trying to fallback to crypto");
+		full_hw_offload = FALSE;
+		if (!config_hw_offload(id, data, hdr, sizeof(request), full_hw_offload))
+		{
+			DBG1(DBG_KNL, "failed to configure HW offload");
+			goto failed;
+		}
+		status = this->socket_xfrm->send_ack(this->socket_xfrm, hdr);
+	}
+
 	if (status == NOT_FOUND && data->update)
 	{
 		DBG1(DBG_KNL, "allocated SPI not found anymore, try to add SAD entry");
--- a/src/libcharon/plugins/vici/vici_config.c
+++ b/src/libcharon/plugins/vici/vici_config.c
@@ -1040,9 +1040,11 @@ CALLBACK(parse_hw_offload, bool,
 	action_t *out, chunk_t v)
 {
 	enum_map_t map[] = {
-		{ "no",		HW_OFFLOAD_NO	},
-		{ "yes",	HW_OFFLOAD_YES	},
-		{ "auto",	HW_OFFLOAD_AUTO	},
+		{ "no",		HW_OFFLOAD_NO		},
+		{ "yes",	HW_OFFLOAD_CRYPTO	},
+		{ "crypto",	HW_OFFLOAD_CRYPTO	},
+		{ "full",	HW_OFFLOAD_FULL		},
+		{ "auto",	HW_OFFLOAD_AUTO		},
 	};
 	int d;
 
--- a/src/libstrongswan/ipsec/ipsec_types.c
+++ b/src/libstrongswan/ipsec/ipsec_types.c
@@ -41,6 +41,8 @@ ENUM(ipcomp_transform_names, IPCOMP_NONE, IPCOMP_LZJH,
 ENUM(hw_offload_names, HW_OFFLOAD_NO, HW_OFFLOAD_AUTO,
 	"no",
 	"yes",
+	"crypto",
+	"full",
 	"auto",
 );
 
--- a/src/libstrongswan/ipsec/ipsec_types.h
+++ b/src/libstrongswan/ipsec/ipsec_types.h
@@ -126,7 +126,9 @@ extern enum_name_t *ipcomp_transform_names;
 enum hw_offload_t {
 	HW_OFFLOAD_NO = 0,
 	HW_OFFLOAD_YES = 1,
-	HW_OFFLOAD_AUTO = 2,
+	HW_OFFLOAD_CRYPTO = 2,
+	HW_OFFLOAD_FULL = 3,
+	HW_OFFLOAD_AUTO = 4,
 };
 
 /**


ferasbi:
Related to this discussion we had previously:
https://github.com/strongswan/strongswan/discussions/1194

tobiasbrunner:
Thanks.

The constant in `xfrm.h` should be the same as in the kernel i.e. `XFRM_OFFLOAD_PACKET`. And maybe that could also be used for the rest (i.e. "packet" instead of "full")? (Was there a discussion somewhere regarding "packet" vs. "full"?)

What's the reason support for full/packet offload can't be queried via `ethtool`? Space limitation? Since that's currently not the case, is there a way to detect the error specifically? (e.g. a specific error code or an extack that points to the `XFRMA_OFFLOAD_DEV` attribute?) Note that the code in your commit might not work correctly when retrying as it just adds an additional such attribute to the request, it doesn't update the existing one that's already there.

What about policy offloading?

tobiasbrunner:
I've pushed an alternative implementation to the _1462-hw-offload-packet_ branch that uses "packet" and should handle the fallback correctly (we only need to remove the flag as the rest is the same and the hardware should support crypto offloading if we came to the point of trying to enable full packet offloading). Still, if we could detect that beforehand or with a more specific error that would be preferable. Please let me know if that works for you.

ferasbi:
Thanks for the fast reply.

Regarding the XFRM flag, indeed the `XFRM_OFFLOAD_PACKET` seems to be giving the same offload parameters for the IPsec state as `XFRM_OFFLOAD_FULL`. `XFRM_OFFLOAD_FULL` was used by us (NVIDIA) until recently when it was added to IPsec upstream it was changed to `XFRM_OFFLOAD_PACKET`.

Regarding the `ethtool`, I'm not sure. As far as I'm aware, when a user wants to configure full/packet HW offload for IPsec, the user also needs to enable TC offloading on the network device through `ethtool`, e.g.:
`ethtool -K $PF hw-tc-offload on `
Where PF is the network/PCI device

Regarding detecting beforehand, unfortunately, `xfrm` stack doesn't have such capabilities. As I understood, the new version of the linux kernel should print a more a specific error when packet/full HW offload fails.

The alternative implementation seems fine by me. Indeed, the hardware should support crypto if reached the point of enabling full/packet HW offloading, so removing the flag is sufficient. My only concern is the naming of the HW offload (packet vs full). When we talk about this type of HW offloading, we always referred to it as full HW offload. I need to check about this and see if it's going to be called packet HW offload from now on.

When we reach a conclusion regarding the naming, will this feature be added to the master and then to a release (5.9.x or 6.x )?

tobiasbrunner:
> Regarding detecting beforehand, unfortunately, `xfrm` stack doesn't have such capabilities.

Yeah, that's unfortunate. To detect basic offload support we currently query `ethtool`'s `esp-hw-offload` bit (see `netlink_find_offload_feature()` and `netlink_detect_offload()`). Something similar that indicates whether packet/full offload is supported and enabled for a specific device would be nice (and maybe similarly for the policy offload, unless that's mandatory for packet offload).

> My only concern is the naming of the HW offload (packet vs full). When we talk about this type of HW offloading, we always referred to it as full HW offload. I need to check about this and see if it's going to be called packet HW offload from now on.

The name `packet` was apparently also used for the flag in the `iproute2` patches. So it might be more consistent to use that throughout. In the updated `swanctl.conf` documentation I now wrote "full packet offloading" to clarify what the `packet` option means.

> When we reach a conclusion regarding the naming, will this feature be added to the master and then to a release (5.9.x or 6.x)?

Sure.

What about policy offload? Should that be added as an additional option (either as flag for `hw_offload` or as separate boolean flag)? Or should it always be enabled if packet/full offload is configured (or is it an optional feature for hardware with packet/full offload)?  Due to the `auto` option we probably need a fallback during policy installation similar to the SA installation (unless there was a `ethtool` bit we could query).

ferasbi:
> > Regarding detecting beforehand, unfortunately, `xfrm` stack doesn't have such capabilities.
> 
> Yeah, that's unfortunate. To detect basic offload support we currently query `ethtool`'s `esp-hw-offload` bit (see `netlink_find_offload_feature()` and `netlink_detect_offload()`). Something similar that indicates whether packet/full offload is supported and enabled for a specific device would be nice (and maybe similarly for the policy offload, unless that's mandatory for packet offload).
> 
Indeed `ethtool` added the `ETHTOOL_GFEATURES` that allows users (amongst other things) to detect basic offload. `xfrm`  currently doesn't support such features. If in the future they add such features that would be great and then we can add it.

> > My only concern is the naming of the HW offload (packet vs full). When we talk about this type of HW offloading, we always referred to it as full HW offload. I need to check about this and see if it's going to be called packet HW offload from now on.
> 
> The name `packet` was apparently also used for the flag in the `iproute2` patches. So it might be more consistent to use that throughout. In the updated `swanctl.conf` documentation I now wrote "full packet offloading" to clarify what the `packet` option means.
> 
Indeed I think using full is not correct and we should be aligned with the upstream. So from now on we will refer to this type offload as packet HW offload (or packet offload). Thanks for adding the clarification in the documentation!

> > When we reach a conclusion regarding the naming, will this feature be added to the master and then to a release (5.9.x or 6.x)?
> 
> Sure.
> 
Great thanks!

> What about policy offload? Should that be added as an additional option (either as flag for `hw_offload` or as separate boolean flag)? Or should it always be enabled if packet/full offload is configured (or is it an optional feature for hardware with packet/full offload)? Due to the `auto` option we probably need a fallback during policy installation similar to the SA installation (unless there was a `ethtool` bit we could query).
>
I consulted the experts and indeed policy offload is mandatory for packet offload (it should always be enabled when packet offload is configured) so there is no need for a separate flag for this, `hw_offload = packet` (and `auto`) should configure packet offload for SAs and policies. I'm currently trying to add this to the policy in the `add_policy_internal` function (similar to what we did in `add_sa`). For this to work we also need the latest iproute2 code that added the packet HW offload (see https://git.kernel.org/pub/scm/network/iproute2/iproute2-next.git/commit/?id=19b358e1ded31189f6081ee7668d003a49ebf96b). I'll try this and get back to you.

tobiasbrunner:
Thanks for the update.

> I consulted the experts and indeed policy offload is mandatory for packet offload (it should always be enabled when packet offload is configured) so there is no need for a separate flag for this, `hw_offload = packet` (and `auto`) should configure packet offload for SAs and policies.

Looking at the kernel code, the problem is that a policy installation with `XFRMA_OFFLOAD_DEV` attribute fails if the `XFRM_OFFLOAD_PACKET` is not set (i.e. it's not just ignored and results in a software installation of the policy). That makes retrying policy installation trickier as we need a message without (and not just a modified) attribute for the second try.

I'll have a look at this and will push an update to the branch later today. Would be great if you could then try if that works on actual hardware.

ferasbi:
> I'll have a look at this and will push an update to the branch later today. Would be great if you could then try if that works on actual hardware.
>
Great, thanks.
When you update the branch I'll run on it on the actual hardware to see if it works.

tobiasbrunner:
> When you update the branch I'll run on it on the actual hardware to see if it works.

Great. I've force pushed the branch with changes that enable offloading for policies if `hw_offload` is either `packet` or `auto`. I hope these work as intended.

ferasbi:
So this what I got:
When we have packet offload capability enabled on our BlueField2 (DPU), configuring `hw_offload = packet` using the new patch indeed configures IPsec with packet offload (as does `auto`).
When we have packet offload capability disabled, configuring `hw_offload = packet` fails (as intended) and `auto` fallbacks to crypto offload (as intended).
The only thing I'm not sure about is the policy offloading, for that I need to install the newest Linux kernel which has the newest patch (currently I'm using internal implementation of the packet offload which is different from the upstream Linux kernel). This may take a while as installing new kernel on the DPU is new to me. I'll try doing this by this week and update you.

tobiasbrunner:
Sounds great, thanks for testing.

ferasbi:
Upgrading to new kernel seems to be problematic. In the next week or so we are expecting that our internal implementation of packet offload to be rebased with the upstream kernel (which has the new implementation of packet offload). When that happen I'll try again to see if the policy offload works as expected and I'll update you. Please keep this thread open.

ferasbi:
Hi again,

I installed the new version of Linux kernel which has the packet HW offload ([Linux 6.2-rc1](https://github.com/torvalds/linux/commit/1b929c02afd37871d5afb9d498426f83432e71c2)) and also the newest version of iproute2 (git://git.kernel.org/pub/scm/network/iproute2/iproute2-next.git) and tried the new patch with these configurations:
First machine has the attached `BFL.swanctl.conf` file with eth3 interface having IP address `192.168.150.1` , note `hw_offload = auto`

BFL swanctl

Second machine has the attached `BFR.swanctl.conf` file with eth3 interface having IP address `192.168.150.2`, note `hw_offload = auto`

BFR swanctl

After loading the conf files on both machines, I initiated a the bf-out connection from the first machine, and this is what I got:


```
# swanctl -i --child bf-out

[IKE] initiating IKE_SA BFL-BFR[1] to 192.168.150.2
[ENC] generating IKE_SA_INIT request 0 [ SA KE No N(NATD_S_IP) N(NATD_D_IP) N(FRAG_SUP) N(HASH_ALG) N(REDIR_SUP) ]
[NET] sending packet: from 192.168.150.1[500] to 192.168.150.2[500] (240 bytes)
[NET] received packet: from 192.168.150.2[500] to 192.168.150.1[500] (248 bytes)
[ENC] parsed IKE_SA_INIT response 0 [ SA KE No N(NATD_S_IP) N(NATD_D_IP) N(FRAG_SUP) N(HASH_ALG) N(CHDLESS_SUP) N(MULT_AUTH) ]
[CFG] selected proposal: IKE:AES_CBC_128/HMAC_SHA2_256_128/PRF_HMAC_SHA2_256/CURVE_25519
[IKE] authentication of 'host1' (myself) with pre-shared key
[IKE] establishing CHILD_SA bf-out{1}
[ENC] generating IKE_AUTH request 1 [ IDi N(INIT_CONTACT) IDr AUTH N(USE_TRANSP) SA TSi TSr N(MULT_AUTH) N(EAP_ONLY) N(MSG_ID_SYN_SUP) ]
[NET] sending packet: from 192.168.150.1[500] to 192.168.150.2[500] (256 bytes)
[NET] received packet: from 192.168.150.2[500] to 192.168.150.1[500] (224 bytes)
[ENC] parsed IKE_AUTH response 1 [ IDr AUTH N(USE_TRANSP) SA TSi TSr ]
[IKE] authentication of 'host2' with pre-shared key successful
[IKE] IKE_SA BFL-BFR[1] established between 192.168.150.1[host1]...192.168.150.2[host2]
[CFG] selected proposal: ESP:AES_GCM_16_128/EXT_SEQ
[KNL] received netlink error: Device failed to offload this state (22)
[KNL] failed to install SA with packet HW offload, trying with crypto HW offload
[IKE] CHILD_SA bf-out{1} established with SPIs cb4346e4_i c579b188_o and TS 192.168.150.1/32[udp] === 192.168.150.2/32[udp/vxlan]
initiate completed successfully
```

It seems that it failed to configure with packet offload (note that new error message from the new kernel: `received netlink error: Device failed to offload this state (22)`)

And when using iproute on the first machine to check the states I get this:
```
# ip xfrm state

src 192.168.150.1 dst 192.168.150.2
	proto esp spi 0xc579b188 reqid 1 mode transport
	replay-window 0 flag esn
	aead rfc4106(gcm(aes)) 0x9421ca148487db8f0eb99bf10c4c323b31e95284 128
	anti-replay esn context:
	 seq-hi 0x0, seq 0x0, oseq-hi 0x0, oseq 0x0
	 replay_window 1, bitmap-length 1
	 00000000 
	crypto offload parameters: dev eth3 dir out mode crypto
	sel src 192.168.150.1/32 dst 192.168.150.2/32 
src 192.168.150.2 dst 192.168.150.1
	proto esp spi 0xcb4346e4 reqid 1 mode transport
	replay-window 0 flag esn
	aead rfc4106(gcm(aes)) 0xdc6154b2c632a9e1541a40844a18de7d86b3c512 128
	anti-replay esn context:
	 seq-hi 0x0, seq 0x0, oseq-hi 0x0, oseq 0x0
	 replay_window 32, bitmap-length 1
	 00000000 
	crypto offload parameters: dev eth3 dir in mode packet
	sel src 192.168.150.2/32 dst 192.168.150.1/32 
```

and the policies:
```
# ip xfrm policy

src 192.168.150.1/32 dst 192.168.150.2/32 proto udp dport 4789 
	dir out priority 366911 ptype main 
	tmpl src 0.0.0.0 dst 0.0.0.0
		proto esp spi 0xc579b188 reqid 1 mode transport
	crypto offload parameters: dev eth3 mode packet
src 192.168.150.2/32 dst 192.168.150.1/32 proto udp sport 4789 
	dir in priority 366911 ptype main 
	tmpl src 0.0.0.0 dst 0.0.0.0
		proto esp reqid 1 mode transport
	crypto offload parameters: dev eth3 mode packet 
```

Which seems weird, because one of the states has `packet` and the other has `crypto` in the offload parameters, while in the policies both have `packet`.
Any ideas what this happens? Note that adding states and policies with packet HW offlaod using `ip xfrm` tool works as expected.

ferasbi:
Also forgot to mention that packet HW offload does not support fwd policies, please take that into account in the patch.

ferasbi:
After debugging, it seems that packet offload doesn't with esn, I'm trying without it and the problem seems to be a bypassing problem. Need to check if they took that into account when implementing the packet offload. Will update you

tobiasbrunner:
> Also forgot to mention that packet HW offload does not support fwd policies, please take that into account in the patch.

Even in tunnel mode? So how are we supposed to handle FWD policies? Install them in software? Not install them at all (that might not work as FWD policies could be required in in- and outbound direction, depending on other policies)?

> After debugging, it seems that packet offload doesn't with esn, I'm trying without it and the problem seems to be a bypassing problem.

The SA that failed was apparently the outbound SA. For those we set the replay window explicitly to 0 (no need for replay protection on outbound SAs) if ESN is not used and to 1 if ESN is enabled (IIRC, the kernel doesn't accept an empty `struct xfrm_replay_state_esn` so we use the minimal size accepted). Maybe the latter is somehow an issue with the hardware. Note that you can [increase the log level](https://docs.strongswan.org/docs/5.9/config/logging.html) for `knl` to 2 to see more details when SAs are installed.

ferasbi:
> Even in tunnel mode? So how are we supposed to handle FWD policies? Install them in software? Not install them at all (that might not work as FWD policies could be required in in- and outbound direction, depending on other policies)?
> 
Yes, install them in software. In general, the kernel doesn't support FWD policies with `packet offload` (in `crypto offload`, the policies aren't offloaded at all).

> The SA that failed was apparently the outbound SA. For those we set the replay window explicitly to 0 (no need for replay protection on outbound SAs) if ESN is not used and to 1 if ESN is enabled (IIRC, the kernel doesn't accept an empty `struct xfrm_replay_state_esn` so we use the minimal size accepted). Maybe the latter is somehow an issue with the hardware. Note that you can [increase the log level](https://docs.strongswan.org/docs/5.9/config/logging.html) for `knl` to 2 to see more details when SAs are installed.
>
The ESN problem seems like an internal capability that we configured incorrectly (in our setups). The current problem seems to be a bypassing problem which we discovered that the firmware didn't take that into consideration. We are fixing that and then we may need to add to the strongSwan code a bypassing policy with `packet offload`. The bypassing wasn't a problem in `crypto offload` because the policies weren't offloaded, in `packet offload` the policies are offloaded.
I'll update you once we solve these problems.

tobiasbrunner:
> > Even in tunnel mode? So how are we supposed to handle FWD policies? Install them in software? Not install them at all (that might not work as FWD policies could be required in in- and outbound direction, depending on other policies)?
> 
> Yes, install them in software. In general, the kernel doesn't support FWD policies with `packet offload` (in `crypto offload`, the policies aren't offloaded at all).

OK, that's fixed now in the branch.

> The ESN problem seems like an internal capability that we configured incorrectly (in our setups).

Since it worked for the inbound SA, is this related to how we configure the window size (1 packet)? If so, we could add an exception if offloading is used and configure the window on the outbound SA the same size as for the inbound SA.

> We are fixing that and then we may need to add to the strongSwan code a bypassing policy with `packet offload`. The bypassing wasn't a problem in `crypto offload` because the policies weren't offloaded, in `packet offload` the policies are offloaded.

There is already a setting in strongswan.conf ( `charon.plugins.kernel-netlink.port_bypass`) that causes the installation of UDP port-specific bypass policies instead of the usual socket policies. We could extend that so that the setting also takes e.g. `offload` as valid option to offload them to the hardware. A problem I see with this is determining the interface as we don't have an address available (the sockets are bound to `0.0.0.0`). And it might even be necessary to offload such policies to multiple devices. So it might instead be necessary to take a list of interface names (possibly via a separate option, e.g. `charon.plugins.kernel-netlink.port_bypass_offload`). These interfaces probably also have to be configured on the selector of these policies (`ifindex`) so the kernel accepts these "duplicate" policies (I've not tried if it actually does accept that, otherwise, that would be a problem and might require the option to add multiple `XFRMA_OFFLOAD_DEV` attributes on policies, or a kernel fix that allows such duplicates). What do you think?

ferasbi:
Sorry for the delay, we were discussing this issue and checking different approaches internally.

> There is already a setting in strongswan.conf ( `charon.plugins.kernel-netlink.port_bypass`) that causes the installation of UDP port-specific bypass policies instead of the usual socket policies. We could extend that so that the setting also takes e.g. `offload` as valid option to offload them to the hardware. A problem I see with this is determining the interface as we don't have an address available (the sockets are bound to `0.0.0.0`). And it might even be necessary to offload such policies to multiple devices. So it might instead be necessary to take a list of interface names (possibly via a separate option, e.g. `charon.plugins.kernel-netlink.port_bypass_offload`). These interfaces probably also have to be configured on the selector of these policies (`ifindex`) so the kernel accepts these "duplicate" policies (I've not tried if it actually does accept that, otherwise, that would be a problem and might require the option to add multiple `XFRMA_OFFLOAD_DEV` attributes on policies, or a kernel fix that allows such duplicates). What do you think?
>
Regarding the interface determining issue, we think that a good approach could be installing a bypass policy per connection and deleting it upon termination. Such approach allows us to delay the bypass policies installation to a later stage where we have the interfaces names.
We saw that libreSwan uses a similar approach to overcome the above issue and we have a POC of the approach (in libreSwan).

The other approach that you mentioned (adding a list of interface names) could also work but we think that it would add complexity to the user configurations (the user would need to add a list of interfaces before initializing the connections). So we prefer the first approach although it may require more code changes.

What are your thoughts about this approach? Do you think this approach should be applied only for the `packet offload` connections or it should be the main approach for all connections?

ferasbi:
> I just realize that drop policies (or manually configured bypass policies) currently can't be offloaded (we don't have an address to determine the interface, but we could probably do that via interface setting). And I don't know how the XFRM default policies translate to the hardware if at all. Do you? 
>
We don't support offloading default policies for now, although it could be translated to a match-all rule (I don't think it's implemented yet). 

> It might also be enough if such drop policies are only installed in software? (In which case our regular bypass mechanism should probably work.)
>
Yes, it should work, since by definition drop policies have a low priority, and the offloading mechanism gives higher priority for hardware offloaded policies hence not offloading any drop policies will fallback to software. So, we think drop policies should stay in software.

> I guess we could also implement this approach in a way that users don't have to configure interfaces manually. For instance, enumerate all interfaces, check if they support offloading and if so try to install bypass policies for the IKE ports on them. 
>
We believe you are right, this approach should work.

> One concern here could be that interfaces might come up (or even get added to the system) after the daemon is already started. If that's actually a problem in practice we'd probably have to handle RTM_NEWLINK events from the kernel and install the bypass policies as interfaces come up.
>
Definitely, we should take care of this case (when new interfaces comes up, we should install the bypass policies in case offloading is supported).

> Do you know what happens with offloaded SAs and policies if an interface goes down and comes back up (e.g. via ip link set dev <iface> down)? Are they flushed or will they still be there?
>
We checked this, offloaded policies are flushed once the interface is down. 

> Is it possible to install policies on an interface that's "down"?
>
It is possible but we think we should prevent this when using strongSwan.

tobiasbrunner:
> since by definition drop policies have a low priority

Unfortunately, that's not true. That's only a convention used for our default priorities. Users are free to configure custom priorities (via `priority` option in swanctl.conf) that are higher than those of other policies (e.g. to use drop policies like firewall rules to block certain traffic that lies within tunneled traffic selectors). If that's the case, the question is if these policies are still enforced if they are only installed in software (i.e. even if the hardware reports that the packet is fully validated).

> We checked this, offloaded policies are flushed once the interface is down.

I guess the same is true for SAs? If guess if that happens, the kernel would trigger acquires for outbound traffic as it won't find an SA, or does it not yet know that the SA is gone (does it learn that at all)?

ferasbi:
> Unfortunately, that's not true. That's only a convention used for our default priorities. Users are free to configure custom priorities (via `priority` option in swanctl.conf) that are higher than those of other policies (e.g. to use drop policies like firewall rules to block certain traffic that lies within tunneled traffic selectors). If that's the case, the question is if these policies are still enforced if they are only installed in software (i.e. even if the hardware reports that the packet is fully validated).
> 
Sorry for not being clear, we meant that the default drop policies which have no IP address must has lower priority by definition since we want to hit it In case we don’t find any other matching policy.
I also want to clarify that today's kernel gives HW policies higher priority by design, and it's up to user to use the offload in a correct way, so if the user would like to have a policy for dropping packets with a specific IP addresses (to act like a firewall) he must offload those policies and we can do that since we have an IP address. However today we don’t have a way to match on a policy in HW while having a matching policy in SW with higher priority.

> I guess the same is true for SAs? If guess if that happens, the kernel would trigger acquires for outbound traffic as it won't find an SA, or does it not yet know that the SA is gone (does it learn that at all)?
>
The offloaded SAs are also deleted (so both SAs and policies are getting deleted by the stack) and regarding kernel acquiring, do you mean when we install trap policies (via `start_action = trap` in the .conf files)? If that is so, then I guess also the trap policies also get deleted since I think they are also offloaded.

ferasbi:
Sorry closed the thread by mistake

tobiasbrunner:
> Sorry for not being clear, we meant that the default drop policies which have no IP address must has lower priority by definition since we want to hit it In case we don’t find any other matching policy.

Ah, I see. That makes sense.

I wonder if we should offload bypass and drop policies with `hw_offload=packet` (or `auto`) that don't specify a specific interface to all interfaces that support offloading (as well as install them without interface in the selector in software), or if we should make an interface mandatory.

> The offloaded SAs are also deleted (so both SAs and policies are getting deleted by the stack) and regarding kernel acquiring, do you mean when we install trap policies (via `start_action = trap` in the .conf files)? If that is so, then I guess also the trap policies also get deleted since I think they are also offloaded.

Not only with trap policies. But as far as I understand, offloaded policies are always installed in software also. When the interface goes down and the offloaded SAs and policies disappear, does that trigger XFRM events from the kernel? And does it automatically affect the software equivalents of the offloaded policies somehow? If the latter is not the case, then the software OUT policies will trigger acquires as no SAs will be found with matching reqid/SPI (unless the kernel has some kind of meta-SA that stores information about the offloaded SA, which will now still be found but won't work anymore with the offloaded SA gone). The daemon won't be able to process such acquires unless a trap policy was installed earlier (at least it would prevent traffic leaks). However, the problem is, of course, that the daemon is not aware that the SAs are gone at all (similar to a manual `ip xfrm state del` on the console as the daemon currently doesn't process `XFRM_MSG_DELSA` messages from the kernel). It still has the CHILD_SA installed and it will refuse to establish another CHILD_SA with the same traffic selectors even when it is prepared to process an acquire (the user would have to terminate the CHILD_SA first). I guess it's a corner case, but not sure what the expected behavior by the keying daemon is here.

tobiasbrunner:
I've updated the _1462-hw-offload-packet_ branch and added changes that should allow installing [custom bypass and drop policies](https://docs.strongswan.org/docs/5.9/config/quickstart.html#_passthroughbypass_policies) to specific interfaces (by setting `interface` and `hw_offload=packet` or `auto`). There is currently no "all interfaces" mode that automatically offloads such policies on all interfaces (even those activated after installing the policy initially).

However, I've also pushed prototypical changes that should cause bypass policies for the IKE ports to get offloaded to all interfaces that support it (even those activated after the daemon created and bypassed the IKE sockets). Would be great if you could test these changes on actual hardware. I don't think it's necessary to  enable `charon.plugins.kernel-netlink.port_bypass`, but if you could test with and without, that would be great.

ferasbi:
> I wonder if we should offload bypass and drop policies with hw_offload=packet (or auto) that don't specify a specific interface to all interfaces that support offloading (as well as install them without interface in the selector in software), or if we should make an interface mandatory.
>
If I understand correctly, you're talking about the IKE bypass policies, so regarding those we want to add them to all interfaces that support `packet offload` to allow the key exchange. Regarding the drop policies, user must specify an interface since the kernel expects an interface and it checks for that so we think that an interface must be mandatory similar to how IProute2 asks for an interface.

> Not only with trap policies. But as far as I understand, offloaded policies are always installed in software also.
>
That is not the case, the approach that was taken in the kernel is to separate HW and software policies while giving any HW offloaded policy a higher priority than any software policy. So the offloaded policies don't have a corresponding software policy to fallback to. 

> When the interface goes down and the offloaded SAs and policies disappear, does that trigger XFRM events from the kernel? And does it automatically affect the software equivalents of the offloaded policies somehow? If the latter is not the case, then the software OUT policies will trigger acquires as no SAs will be found with matching reqid/SPI (unless the kernel has some kind of meta-SA that stores information about the offloaded SA, which will now still be found but won't work anymore with the offloaded SA gone). The daemon won't be able to process such acquires unless a trap policy was installed earlier (at least it would prevent traffic leaks). However, the problem is, of course, that the daemon is not aware that the SAs are gone at all (similar to a manual ip xfrm state del on the console as the daemon currently doesn't process XFRM_MSG_DELSA messages from the kernel). It still has the CHILD_SA installed and it will refuse to establish another CHILD_SA with the same traffic selectors even when it is prepared to process an acquire (the user would have to terminate the CHILD_SA first). I guess it's a corner case, but not sure what the expected behavior by the keying daemon is here.
> 
We think that the mentioned approach above prevent such issues (a similar issue was in socket policies hence the separation of software and HW policies).

> I've updated the 1462-hw-offload-packet branch and added changes that should allow installing [custom bypass and drop policies](https://docs.strongswan.org/docs/5.9/config/quickstart.html#_passthroughbypass_policies) to specific interfaces (by setting interface and hw_offload=packet or auto). There is currently no "all interfaces" mode that automatically offloads such policies on all interfaces (even those activated after installing the policy initially).

However, I've also pushed prototypical changes that should cause bypass policies for the IKE ports to get offloaded to all interfaces that support it (even those activated after the daemon created and bypassed the IKE sockets). Would be great if you could test these changes on actual hardware. I don't think it's necessary to enable charon.plugins.kernel-netlink.port_bypass, but if you could test with and without, that would be great.
>

I'll test and update you

tobiasbrunner:
> If I understand correctly, you're talking about the IKE bypass policies, so regarding those we want to add them to all interfaces that support `packet offload` to allow the key exchange. Regarding the drop policies, user must specify an interface since the kernel expects an interface and it checks for that so we think that an interface must be mandatory similar to how IProute2 asks for an interface.

I was referring to general bypass and drop policies there, which we could theoretically offload to all interfaces (similar to the approach I've now implemented for the IKE bypass policies).  But it does require a lot more bookkeeping, so for now the approach that requires an interface in the config is a lot simpler and probably adequate for most use cases.

> > Not only with trap policies. But as far as I understand, offloaded policies are always installed in software also.
> 
> That is not the case, the approach that was taken in the kernel is to separate HW and software policies while giving any HW offloaded policy a higher priority than any software policy. So the offloaded policies don't have a corresponding software policy to fallback to.

Really? How can outbound traffic be handled? How does the kernel know that that traffic matches an offloaded policy and should be handled by the hardware? Or is there a lookup with some data (addresses, ports, protocol) to the hardware first? If so, how does the kernel know which hardware to query with that?

Also for inbound traffic, what if traffic that matches an offloaded inbound policy does not arrive over the interface that handles the offloading? The kernel should drop that traffic as it wasn't encrypted. But if the offloaded policy is not also held in software how could it do that? Or is there again some lookup via hardware? (If so, same questions as above.)

> I'll test and update you

Great, thanks!

ferasbi:
> Really? How can outbound traffic be handled? How does the kernel know that that traffic matches an offloaded policy and should be handled by the hardware? Or is there a lookup with some data (addresses, ports, protocol) to the hardware first?
>
The approach that was taken in the kernel is to give higher priority for HW policies by sorting the policies list (the database) and making sure that HW policies (if exist) are prior to SW policies.
Regarding handling of outbound traffic, there is always a lookup with some data (addresses, ports and protocols like you mentioned, please see `xfrm_state_find` function in https://elixir.bootlin.com/linux/v6.2-rc6/source/net/xfrm/xfrm_state.c#L1137 ).

>  If so, how does the kernel know which hardware to query with that?
>
This data (which hardware to query with) comes from the packet.

> Also for inbound traffic, what if traffic that matches an offloaded inbound policy does not arrive over the interface that handles the offloading? The kernel should drop that traffic as it wasn't encrypted. But if the offloaded policy is not also held in software how could it do that? Or is there again some lookup via hardware? (If so, same questions as above.)
>
Good question. Most likely we have a lookup via hardware. I need to take it with the feature owner to verify that this is the case (I'll get back to you).

>  I don't think it's necessary to enable charon.plugins.kernel-netlink.port_bypass, but if you could test with and without, that would be great.
>
So we tested the new patch with various configurations and with both port_bypass and without it and here are the results:
Without start_action everything works as expected, we manage to bypass the IKE traffic and succefully creating a connection with `packet offload`.
Setting start_action = start also worked for us, same as above.
Setting start_action = trap failed (when pinging we get "ping: connect: Invalid argument"). We debugged the issue and found out that the root cause is our driver denying to create the initial state (empty state) which strongSwan updates once an acquire is received. We are currently working on fixing that and will update you.
Note that we didn't have to add to the swanctl.conf file the interface name for the above to work but we tested with and without the interface configure option, and we got similar results in both cases. 
Also, as expected, bringing up an interface (that supports `packet offload`) automatically adds the bypass policies, and setting the interface down delete those policies (great !!).

ferasbi:
Sorry closed the thread by mistake again

tobiasbrunner:
> The approach that was taken in the kernel is to give higher priority for HW policies by sorting the policies list (the database) and making sure that HW policies (if exist) are prior to SW policies.

Yes, but doesn't that mean there is some kind of skeleton policy for each offloaded policy with `xdo.type == XFRM_DEV_OFFLOAD_PACKET`. Is that policy updated (or removed) if the corresponding policy disappears on the interface?

What I fear is something like this:
 1. SAs and policies are offloaded to the hardware, everything is properly encrypted
 2. Now an attacker (or even an unsuspecting user) pulls out the network cable, the interface goes down and the offloaded policies and SAs disappear
 3. The cable is plugged back in, however, the traffic now leaves the interface without any protection as the policies and SAs are gone, while the kernel still "offloads" packets to the hardware due to the skeleton policy/state

> Regarding handling of outbound traffic, there is always a lookup with some data (addresses, ports and protocols like you mentioned, please see `xfrm_state_find` function in https://elixir.bootlin.com/linux/v6.2-rc6/source/net/xfrm/xfrm_state.c#L1137 ).

That's the state lookup and matching policy and state seems to happen in software (i.e. for each offloaded state there is a skeleton state, too, that's matched there on line 1178). And I guess before that there was a policy lookup in software as well that found the OUT skeleton policy with `xdo.type == XFRM_DEV_OFFLOAD_PACKET` in the first place. So if the skeleton state is successfully found and the kernel is done and offloads the packet to the interface, I guess the hardware will just do another policy and state lookup. Or do I misunderstand this?

> > If so, how does the kernel know which hardware to query with that?
> 
> This data (which hardware to query with) comes from the packet.

From the packet how? Isn't the interface stored in `xdo.dev` of the policy and state (i.e. only after the kernel found the state in software will it offload the packet to the device stored there)?

> Good question. Most likely we have a lookup via hardware. I need to take it with the feature owner to verify that this is the case (I'll get back to you).

OK, thanks for looking into it. It might work if the kernel matches the packet against the skeleton policies and drops it if it doesn't find any indication that it was decrypted by the hardware (I guess there is some flag for it?).

> So we tested the new patch with various configurations and with both port_bypass and without it and here are the results:

Thank for the tests.

> We debugged the issue and found out that the root cause is our driver denying to create the initial state (empty state) which strongSwan updates once an acquire is received.

Hm, not sure what state you are referring to. When installing a trap policy, strongSwan will not install a state, only policies. However, the kernel will create a state once it matches a packet to a policy but doesn't find a corresponding state (it does so right before it sends the acquire, which then will also prevent further acquires until that temporary state expires). Do you mean that state?

> Note that we didn't have to add to the `swanctl.conf` file the interface name for the above to work but we tested with and without the interface configure option, and we got similar results in both cases.

That's only necessary for bypass and drop policies as they don't have a local address associated with them (see the docs for [example bypass policies](https://docs.strongswan.org/docs/5.9/config/quickstart.html#_passthroughbypass_policies), drop policies are the same, they just use `mode = drop`, a custom priority can be configured via `priority`).  For regular CHILD_SAs we just use the interface on which the local address resides, so for these it's not necessary to do anything special.

> Also, as expected, bringing up an interface (that supports `packet offload`) automatically adds the bypass policies, and setting the interface down delete those policies (great !!).

Actually, the code currently doesn't actively remove the IKE bypass policies when the interface goes down. But considering that there are skeleton policies in software that might be something we need to do (I guess we could also ignore the EEXIST error returned by the kernel once we install the policies again if the interface comes back up - but not sure if the kernel would actually offload the policy again to the hardware if it still has the skeleton).

ferasbi:
> Yes, but doesn't that mean there is some kind of skeleton policy for each offloaded policy with `xdo.type == XFRM_DEV_OFFLOAD_PACKET`. Is that policy updated (or removed) if the corresponding policy disappears on the interface?
> 
> What I fear is something like this:
> 
> 1. SAs and policies are offloaded to the hardware, everything is properly encrypted
> 2. Now an attacker (or even an unsuspecting user) pulls out the network cable, the interface goes down and the offloaded policies and SAs disappear
> 3. The cable is plugged back in, however, the traffic now leaves the interface without any protection as the policies and SAs are gone, while the kernel still "offloads" packets to the hardware due to the skeleton policy/state
> 

When a change occurs to the interface (by an event) , all the policies and states are deleted. (Those "skeleton" policy/state are not skeleton for say , those policies only exist if HW policies/state exist and deleted if HW policies/states are deleted , the XFRM stack listens to the interface events and act accordingly to delete the "skeleton" policy/state).
Please note that we don’t save those policies/state as a backup to bring back the offload when the interface is back up, hence making the interface go down and back up will cancel IPsec.

> That's the state lookup and matching policy and state seems to happen in software (i.e. for each offloaded state there is a skeleton state, too, that's matched there on line 1178). And I guess before that there was a policy lookup in software as well that found the OUT skeleton policy with `xdo.type == XFRM_DEV_OFFLOAD_PACKET` in the first place. So if the skeleton state is successfully found and the kernel is done and offloads the packet to the interface, I guess the hardware will just do another policy and state lookup. Or do I misunderstand this?
> 

IIUC we look up for a state/policy and if we find a HW state/policy ("skeleton state") then the kernel does not encrypt this and leave it for HW. If we don’t have offload then we will find the SW state and hence we will handle IPsec in the kernel.

> From the packet how? Isn't the interface stored in `xdo.dev` of the policy and state (i.e. only after the kernel found the state in software will it offload the packet to the device stored there)?
> 
During configuration time SA and policy are installed in HW plus the "skeleton policy/state" is added in SW with indication flag that its HW policy/state (xdo.type == XFRM_DEV_OFFLOAD_PACKET)
regarding the device it comes from the user (from the command line) when installing , regarding the inbound the driver add the sec path (code:  sp = secpath_set(skb) ) with decrypted status for offloaded ipsec packets.

> OK, thanks for looking into it. It might work if the kernel matches the packet against the skeleton policies and drops it if it doesn't find any indication that it was decrypted by the hardware (I guess there is some flag for it?).
> 
Exactly, the kernel matches the packet against the skeleton policies and drops it if it doesn't find any indication that it was decrypted by the hardware 
(the indication flag is  xo->flags = CRYPTO_DONE)

> Hm, not sure what state you are referring to. When installing a trap policy, strongSwan will not install a state, only policies. However, the kernel will create a state once it matches a packet to a policy but doesn't find a corresponding state (it does so right before it sends the acquire, which then will also prevent further acquires until that temporary state expires). Do you mean that state?
> 
Yes,  that state.


> Actually, the code currently doesn't actively remove the IKE bypass policies when the interface goes down. But considering that there are skeleton policies in software that might be something we need to do (I guess we could also ignore the EEXIST error returned by the kernel once we install the policies again if the interface comes back up - but not sure if the kernel would actually offload the policy again to the hardware if it still has the skeleton).
>
As explained above, any change occurs to the interface, all the policies and states are deleted

tobiasbrunner:
Thanks for the update.

> When a change occurs to the interface (by an event) , all the policies and states are deleted. (Those "skeleton" policy/state are not skeleton for say , those policies only exist if HW policies/state exist and deleted if HW policies/states are deleted , the XFRM stack listens to the interface events and act accordingly to delete the "skeleton" policy/state).
> Please note that we don’t save those policies/state as a backup to bring back the offload when the interface is back up, hence making the interface go down and back up will cancel IPsec.

OK, the latter is unfortunate. In particular in regards to the policies (I get why you don't want to store the SA keys in the skeleton states, but the policies are lightweight and all the information is there in the skeleton policies). But if they are removed in the kernel at least any software drop policies (or a default XFRM drop policy) should get enforced.

And I guess this confirms that re-installation of the IKE bypass policies on re-activated interfaces is definitely necessary (but can be done safely as there won't be a conflict with existing policies). For manually offloaded drop/bypass policies we'd theoretically have to do something similar to keep them installed persistently. But I'll leave that for now as it would require significant changes and might only be necessary in corner cases.

Also, is this behavior specific to your driver or is this inherently part of the XFRM offloading framework and will apply to all devices that support packet offload?

> regarding the inbound the driver add the sec path (code:  sp = secpath_set(skb) ) with decrypted status for offloaded ipsec packets

OK, so I guess the skeleton policy (as long as it exists) should drop the packet if it comes from a different interface (i.e. was not decrypted and has no sec path). If they're gone, that again would require software drop policies.

> As explained above, any change occurs to the interface, all the policies and states are deleted

OK, but does that trigger any events from the kernel? For instance, for the states it might be interesting if the kernel could send a hard expire (`XFRM_MSG_EXPIRE` with `struct xfrm_user_expire::hard != 0`), similar to when the kernel deleted an SA because it expired (i.e. if it was not replaced with a rekeying for some reason and its lifetime limits were exceeded). That at least would trigger the recreation of the CHILD_SA in the IKE daemon and not let them linger without actual SA in the kernel. There is also `XFRM_MSG_POLEXPIRE` for policies, but we currently don't handle that (we don't configure any lifetimes on the policies).

I guess `XFRM_MSG_DELSA` and `XFRM_MSG_DELPOLICY` could also be an option, however, these are currently only triggered as notifies if the kernel acts on deletes from userland (and the IKE daemon does currently not listen for such events as it assumes full control over SAD/SPD).

ferasbi:
> OK, the latter is unfortunate. In particular in regards to the policies (I get why you don't want to store the SA keys in the skeleton states, but the policies are lightweight and all the information is there in the skeleton policies). But if they are removed in the kernel at least any software drop policies (or a default XFRM drop policy) should get enforced.
And I guess this confirms that re-installation of the IKE bypass policies on re-activated interfaces is definitely necessary (but can be done safely as there won't be a conflict with existing policies). For manually offloaded drop/bypass policies we'd theoretically have to do something similar to keep them installed persistently. But I'll leave that for now as it would require significant changes and might only be necessary in corner cases.
> 
Agree.

> Also, is this behavior specific to your driver or is this inherently part of the XFRM offloading framework and will apply to all devices that support packet offload?
> 
This is inherently part of the XFRM offloading framework and will apply to all devices that support packet offload (not just our drivers).


> OK, but does that trigger any events from the kernel? For instance, for the states it might be interesting if the kernel could send a hard expire (`XFRM_MSG_EXPIRE` with `struct xfrm_user_expire::hard != 0`), similar to when the kernel deleted an SA because it expired (i.e. if it was not replaced with a rekeying for some reason and its lifetime limits were exceeded). That at least would trigger the recreation of the CHILD_SA in the IKE daemon and not let them linger without actual SA in the kernel. There is also `XFRM_MSG_POLEXPIRE` for policies, but we currently don't handle that (we don't configure any lifetimes on the policies).
I guess `XFRM_MSG_DELSA` and `XFRM_MSG_DELPOLICY` could also be an option, however, these are currently only triggered as notifies if the kernel acts on deletes from userland (and the IKE daemon does currently not listen for such events as it assumes full control over SAD/SPD).
>
Good point , I'm not sure about any events triggered from the kernel, we also think the solution you're describing should work however it will require efforts from both kernel and strongSwan side,
So we think that it is best as a first step support to add support for what is currently supported in kernel and after that we start taking care of corner cases to eventually get to fully support everything.

tobiasbrunner:
Alright, I've rebased the _1462-hw-offload-packet_ branch. Do you think that's good enough for now? Or are there any pressing open issues?

ferasbi:
From our side, after testing, we see that the patches are more than good enough. In the future if we discover issues/corner cases that need to be raised, we can handle them step by step.
Thanks a lot Tobias!

tobiasbrunner:
I've merged the changes to master. Please let me know if anything else comes up.

PR #1488, opened 2023-01-03
Replace destroy in the failed SA with DESTROY_IF

TNLTom:
On testing, we get random segfaults on this particular destroy call. I can't really reproduce it but replacing the destroy call with DESTROY_IF should not hurt and since then we have not seen the segfault anymore.


tobiasbrunner:
You should investigate when this happens exactly, because there should not be a call to this function with `this->child_sa` being `NULL`. It's called from `process_i*` implementations, that is, after `build_i`, which always allocates `this->child_sa` (and that can't have failed as there would have been a crash earlier). So there has to be some other issue here. What failure triggers the call to `delete_failed_sa()`?

Also note that the _ikev2-qske-multi-ke_ branch will not be developed any further, the current branch with multi-KE support is _six-beta_.

PR #1511, opened 2023-01-15
Include IPSEC_CONFDIR variable replacement in ipsec man

pemensik:
Fedora chosen different default directory to avoid conflicts with libreswan. Use ${sysconfdir} variable to provide correct location.


pemensik:
Solves downstream bug: https://bugzilla.redhat.com/show_bug.cgi?id=2106120

tobiasbrunner:
Thanks. I've pushed these to master with minor changes to the commit messages. I've also pushed additional commits to fix the paths in the strongswan.conf and swanctl.conf man pages.

pemensik:
Just for the record. My commits applied on commit ee046552bb1f3c98d89837d58f7da7d83c8fbb82 with bunch of commits following.

PR #1570, opened 2023-03-02
vici: generalize timeout support

### LINKED ISSUE #1562: Python timeout
Since the merge of https://github.com/strongswan/strongswan/pull/1416, any timeout set directly on the socket is systematically reset by `vici.Transport.receive()` unless called by `vici.Session.listen()` thus making it impossible to configure some default timeout from the outside.

There could be at least 2 solutions for this:
1. Add a notion of "default timeout" on the Session that would be forwarded and applied by the Transport and let the exception propagate.
2. Add an explicit optional "timeout" parameter on all network method (`vici.Session.request()`, `vici.Session.streamed_request()`) as well as each of their wrapper in `CommandWrappers`.

If either of these look acceptable to the project, I can open a PR.

yadutaf:
Since #1416, timeout set directly on the socket are reset by `vici.Transport.receive()` unless called by `vici.Session.listen()`. This prevents configuring a default timeout directly on the socket.

However, setting a timeout directly on the socket also had drawbacks since it can cause `vici.Transport.receive()` to raise a timeout error while a subset of the data have been received, with no way to recover.

This commit merges both approach by considering the timeout configured on the socket by default (when no timeout is explicitly set) and keeping the switch to blocking receive once the first byte has been received. When the full expected data have been received, the timeout configured on the socket is restored.

Closes #1562

### DIFF
--- a/src/libcharon/plugins/vici/python/vici/protocol.py
+++ b/src/libcharon/plugins/vici/python/vici/protocol.py
@@ -8,6 +8,9 @@ from collections import OrderedDict
 from .exception import DeserializationException
 
 
+RECV_TIMEOUT_DEFAULT = object()
+
+
 class Transport(object):
     HEADER_LENGTH = 4
     MAX_SEGMENT = 512 * 1024
@@ -18,7 +21,7 @@ class Transport(object):
     def send(self, packet):
         self.socket.sendall(struct.pack("!I", len(packet)) + packet)
 
-    def receive(self, timeout=None):
+    def receive(self, timeout=RECV_TIMEOUT_DEFAULT):
         raw_length = self._recvall(self.HEADER_LENGTH, timeout)
         length, = struct.unpack("!I", raw_length)
         payload = self._recvall(length)
@@ -28,17 +31,21 @@ class Transport(object):
         self.socket.shutdown(socket.SHUT_RDWR)
         self.socket.close()
 
-    def _recvall(self, count, timeout=None):
+    def _recvall(self, count, timeout=RECV_TIMEOUT_DEFAULT):
         """Ensure to read count bytes from the socket"""
         data = b""
-        if count > 0:
+        old_timeout = self.socket.gettimeout()
+        if timeout is not RECV_TIMEOUT_DEFAULT:
             self.socket.settimeout(timeout)
-        while len(data) < count:
-            buf = self.socket.recv(count - len(data))
-            self.socket.settimeout(None)
-            if not buf:
-                raise socket.error('Connection closed')
-            data += buf
+        try:
+            while len(data) < count:
+                buf = self.socket.recv(count - len(data))
+                self.socket.settimeout(None)
+                if not buf:
+                    raise socket.error('Connection closed')
+                data += buf
+        finally:
+            self.socket.settimeout(old_timeout)
         return data
 
 
--- a/src/libcharon/plugins/vici/python/vici/session.py
+++ b/src/libcharon/plugins/vici/python/vici/session.py
@@ -1,12 +1,29 @@
 import socket
 
 from .exception import SessionException, CommandException, EventUnknownException
-from .protocol import Transport, Packet, Message
+from .protocol import Transport, Packet, Message, RECV_TIMEOUT_DEFAULT
 from .command_wrappers import CommandWrappers
 
 
 class Session(CommandWrappers, object):
     def __init__(self, sock=None):
+        """Establish a session with an IKE daemon.
+        By default, the session will connect to the `/var/run/charon.vici` Unix
+        domain socket.
+
+        If there is a need to connect a socket in another location or set
+        specific settings on the socket (like a timeout), create and connect
+        a socket and pass it to the `sock` parameter.
+
+        .. note::
+
+            In case a timeout is set on the socket, the internal read code
+            will temporarily disable it after receiving the first byte to avoid
+            partial read corruptions.
+
+        :param sock: socket connected to the IKE daemon (optional)
+        :type sock: socket.socket
+        """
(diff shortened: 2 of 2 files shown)


tobiasbrunner:
@cla-bot check

yadutaf:
@tobiasbrunner Could you hint me at how to sign the CLA? I tried to find the information in https://docs.strongswan.org/docs/5.9/devs/contributions.html but maybe missed it.

tobiasbrunner:
Thanks for sending the CLA. @cla-bot check

src/libcharon/plugins/vici/python/README.rst
```diff
@@ -22,3 +22,16 @@ Example Usage
...
+
+A timeout can be configured directly on the underlying socket. Please note that
+`vici` does not support resuming partial read. Thus, to avoid corrupting
+messages, the socket temporarily switches back to blocking read after the first
+byte has been received. The initially configured timeout is restored once all
+expected bytes have been received.
+
+.. code-block:: python
+
+   >>> s.transport.socket.settimeout(10)
```
tobiasbrunner, line 37: Hm, I don't really like this code example that accesses the internal socket.

src/libcharon/plugins/vici/python/README.rst
```diff
@@ -22,3 +22,16 @@ Example Usage
...
     ('online', b'0'), ('offline', b'0')]))])
+
+Timeout support
+---------------
+
+A timeout can be configured directly on the underlying socket. Please note that
+`vici` does not support resuming partial read. Thus, to avoid corrupting
+messages, the socket temporarily switches back to blocking read after the first
+byte has been received. The initially configured timeout is restored once all
+expected bytes have been received.
```
tobiasbrunner, line 33: Documenting this so prominently might give the wrong impression that setting a timeout is something users should or even have to do (the text currently does also not explain why one would set such a timeout on the socket outside of `listen()` where it already can be configured without direct access to the socket).

src/libcharon/plugins/vici/python/vici/protocol.py
```diff
@@ -8,6 +8,8 @@
 from .exception import DeserializationException
 
 
+RECV_TIMEOUT_DEFAULT = object()
+
 class Transport(object):
```
tobiasbrunner, line 13: Please add another blank line between these two elements.

src/libcharon/plugins/vici/python/vici/protocol.py
```diff
@@ -28,17 +30,22 @@ def close(self):
...
+        try:
+            while len(data) < count:
+                buf = self.socket.recv(count - len(data))
+                self.socket.settimeout(None)
+                if not buf:
+                    raise socket.error('Connection closed')
+                data += buf
+        finally:
+            self.socket.settimeout(old_timeout)
+
```
tobiasbrunner, line 48: I don't think this blank line is necessary.

tobiasbrunner (changes requested):
I'm not sure about the documentation of this. I've added some comments.

src/libcharon/plugins/vici/python/README.rst
yadutaf: I think it's important to at least mention it, but I agree that the Readme is too prominent for an edge use-case. Would it be better to mention it in the docstring of `Session.__init__()` ?

It could be like:

```python
    def __init__(self, sock=None):
        """Establish a session with a StrongSwan daemon. By default, the Session
        will connect to StrongSwan via the `/var/run/charon.vici` Unix Domain
        Socket.

        If there is a need to connect a socket in another location or set
        specific settings on the socket like a timeout, create a connection and
        pass it to the `sock` parameter.

        :param sock: Socket connected to StrongSwan (optional)
        :type sock: socket.socket
        """
```

This would address both comments.

tobiasbrunner: Yes, I like that a lot more. I'll add some editorial suggestions.

src/libcharon/plugins/vici/python/vici/session.py
```diff
@@ -1,12 +1,29 @@
...
+        specific settings on the socket like a timeout, create a connection and
+        pass it to the `sock` parameter.
+
+        .. note::
+
+            In case a timeout is set on the socket, the internal read code
+            will temporarily disable it after receiving the first byte to avoid
+            partial read corruptions.
+
+        :param sock: Socket connected to StrongSwan (optional)
```
tobiasbrunner, line 24:
```suggestion
        """Establish a session with an IKE daemon.

        By default, the session will connect to the `/var/run/charon.vici` Unix
        domain socket.

        If there is a need to connect a socket in another location or set
        specific settings on the socket (like a timeout), create and connect
        a socket and pass it to the `sock` parameter.

        .. note::

            In case a timeout is set on the socket, the internal read code
            will temporarily disable it after receiving the first byte to avoid
            partial read corruptions.

        :param sock: socket connected to the IKE daemon (optional)
```

tobiasbrunner (changes requested):

src/libcharon/plugins/vici/python/vici/session.py
yadutaf: Applied, thanks.

tobiasbrunner:
Looks good, thanks a lot. Is it OK for you if I add a co-authored-by tag to the commit?

yadutaf:
Please do. Thanks for asking.

tobiasbrunner:
Great, thanks again. The change is now in master (the "Closes" tag references #1562  and not this PR, which would probably have been better, but doesn't really matter). I've also updated [the package on PyPI](https://pypi.org/project/vici/).

PR #1615, opened 2023-03-21
pkcs7: allow SHA hashes with RSA encryption in openssl pkcs7 validation

hargut:
Hello together,

this PR allows validation for SHA hashed RSA encrypted pkcs7 signatures. As of now pcks7 only validates using `openssl` in case the signature is RSA encrypted.

During validation of the according RFC's I could not locate any areas that would restrict the usage of hashes within signatures. It is also not clear for me why only RSA is currently supported. This might be related to other architectural constraints that I've not had in my view.

As C is not my typical language I'd be happy to receive any feedback on the implementation.

references:
CMS [RFC 5652, Section 10.1.2](https://www.rfc-editor.org/rfc/rfc5652.html#section-10.1.2)
SCEP [RFC 8894, Section 3.5.2](https://www.rfc-editor.org/rfc/rfc8894.html#name-ca-capabilities-response-fo)
SCEP [RFC 8894, Section 7.9](https://www.rfc-editor.org/rfc/rfc8894.html#name-use-of-sha-1)

Many thanks for your consideration.

Best regards,
Harald

### DIFF
--- a/src/libstrongswan/plugins/openssl/openssl_pkcs7.c
+++ b/src/libstrongswan/plugins/openssl/openssl_pkcs7.c
@@ -36,6 +36,13 @@
 
 typedef struct private_openssl_pkcs7_t private_openssl_pkcs7_t;
 
+static int allowed_signature_algorithms[] = {
+		OID_RSA_ENCRYPTION,
+		OID_SHA1_WITH_RSA,
+		OID_SHA256_WITH_RSA,
+		OID_SHA512_WITH_RSA,
+};
+
 /**
  * Private data of an openssl_pkcs7_t object.
  */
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
@cla-bot check

tobiasbrunner (commented):
Yes, looks like we should update this code to come closer to actual CMS support. What's now called `signatureAlgorithm` (`SignatureAlgorithmIdentifier`) in the `SignerInfo` structure was called `digestEncryptionAlgorithm` (`DigestEncryptionAlgorithmIdentifier`) in PKCS#7 and only `rsaEncryption` was defined back then.

But I feel like your change is not enough and some more changes are required to support signatures properly (e.g. to also support ECDSA). And the same would also be required in the _pkcs7_ plugin, which has the same limitation. Something like `signature_scheme_from_oid(openssl_asn1_known_oid(sig->algorithm))` might be a starting point here, then derive the key type from that (to support RSA-PSS signatures additional work is required to handle encoded parameters).

I might have a look at this tomorrow.

hargut:
Thank you for the feedback and additional details.

I had assumed that the limitation for RSA only was related to an older specification.

hargut:
Hello @tobiasbrunner,

I've successfully tested the initial scenario with branch `1615-pkcs7-signatures`  and the following configuration:
```
./configure --enable-curl --enable-openssl --enable-pkcs7

 strongSwan will be built with the following plugins
-----------------------------------------------------
libstrongswan: aes des rc2 sha2 sha1 md5 mgf1 random nonce x509 revocation constraints pubkey pkcs1 pkcs7 pkcs12 pgp dnskey sshkey pem openssl pkcs8 fips-prf gmp curve25519 xcbc cmac hmac kdf gcm drbg curl
libcharon:     attr kernel-netlink resolve socket-default stroke vici updown xauth-generic counters
libtnccs:     
libtpmtss: 
```

The `CertRep` contained in this case:
```
messageType:  CertRep
pkiStatus:    SUCCESS
...
L1 - signatureAlgorithm:
L2 - algorithmIdentifier:
L3 - algorithm:
  'sha256WithRSAEncryption'
```

Thank you very much for your effort in this matter. :+1: 


Kind regards,
Harald

PR #1616, opened 2023-03-21
pki: EST filter whitespace on base64 encoded responses

hargut:
Hello together,

this PR arose from an response parsing issue using the EST protocol. Looking into the details it got clear that the responses base64 data formatting in combination with the method used for base64 decoding caused the issue.

The change only focuses on the `pki est` command, and adds a white space filter as per RFC in case the `Content-Type-Encoding` is `base64`.
It might be a consideration to adjust the implementation in `utils/chunk.c` `chunk_from_base64()` to only accept valid base64 characters before attempting to decode. As it was not clear to me if this section is performance critical I've limited the change to the `pki/est` section.

As C is not my typical language I'd be happy to receive any feedback on the implementation.

reference: EST Clarification [RFC 8951, Section 3.1](https://www.rfc-editor.org/rfc/rfc8951#name-white-space-processing)

Many thanks for your consideration.

Best regards,
Harald

### DIFF
--- a/src/pki/est/est_tls.c
+++ b/src/pki/est/est_tls.c
@@ -15,6 +15,7 @@
  */
 
 #define _GNU_SOURCE /* for asprintf() */
+#include <ctype.h>
 #include <stdio.h>
 #include <unistd.h>
 #include <errno.h>
@@ -198,7 +199,6 @@ static bool parse_http_header(chunk_t *in,  u_int *http_code, u_int *content_len
 	return (*http_code < 300);
 }
 
-
 METHOD(est_tls_t, request, bool,
 	private_est_tls_t *this, est_op_t op, chunk_t in, chunk_t *out,
 	u_int *http_code, u_int *retry_after)
@@ -206,7 +206,7 @@ METHOD(est_tls_t, request, bool,
 	chunk_t http = chunk_empty, data = chunk_empty, response;
 	u_int content_len;
 	char buf[1024];
-	int len;
+	int i, len;
 
 	/* initialize output variables */
 	*out = chunk_empty;
@@ -276,6 +276,15 @@ METHOD(est_tls_t, request, bool,
 				return FALSE;
 			}
 		}
+
+		for (i = 0, len = 0; i < data.len; i++)
+		{
+			if (!isspace(data.ptr[i]))
+			{
+				data.ptr[len++] = data.ptr[i];
+			}
+		}
+		data.len = len;
 
 		*out = chunk_from_base64(data, NULL);
 		chunk_free(&data);


tobiasbrunner:
@cla-bot check

src/pki/est/est_tls.c
```diff
@@ -272,6 +279,22 @@ METHOD(est_tls_t, request, bool,
...
+			for (int i = 0; i < data.len; i++) {
+				char c = data.ptr[i];
+				if (c != '\r' && c != '\n' &&
+					c != '\t' && c != ' ') {
+					filtered[cnt] = data.ptr[i];
+					cnt++;
+				}
+			}
+
+			data = chunk_create(filtered, cnt);
```
tobiasbrunner, line 295: This could be simplified (`i` has also to be delared where `len` is):

```suggestion
			for (i = 0, len = 0; i < data.len; i++)
			{
				if (!isspace(data.ptr[i]))
				{
					data.ptr[len++] = data.ptr[i];
				}
			}
			data.len = len;
```

tobiasbrunner (changes requested):
Note that since a708e969068b32a5abb07bd325de19f6670109f6 the code always assumes the content is Base64 encoded. And since you haven't made the the actual `chunk_from_base64()` call conditional, I feel like the `content_b64enc` flag etc. is not actually necessary.

I think filtering whitespace makes sense, but the code could be simplified (see my comment).

src/pki/est/est_tls.c
hargut: Thank you for the feedback and pointer to the errata and previous removal.

The topic came up as base64 encoded response containig CRLF could not be parsed/validated successfully.

I'll update the PR within the next week.

hargut: I've updated the PR to include the suggested change and have rebased it on the current `master`.

The proposed solution to re-use/modify the existing `chunk` is elegant and simple. :+1:

src/pki/est/est_tls.c
```diff
@@ -193,6 +193,19 @@ static bool parse_http_header(chunk_t *in,  u_int *http_code, u_int *content_len
...
+	{
+		case ' ':
+		case '\n':
+		case '\r':
+		case '\t':
+			return true;
+		default:
+			return false;
+	}
+}
```
tobiasbrunner, line 208: This is not necessary, there is a standard C function called `isspace()` for this purpose (you need to add `#include <ctype.h>` for the declaration).

hargut: Sorry, I've missed that as I was thinking that the `isspace` from the first comment should show that it would be cleaner to read/understand when using a separate function with a clean name.

Updated.

tobiasbrunner: I should have pointed that out, sorry. Is it OK for you if I add myself as co-author to the commit?

PR #1636, opened 2023-04-02
openssl: fix x509 SubjectAlternativeName wildcard validation

hargut:
The patch fixes the x509 SubjectAlternativeName wildcard validation within the openssl plugin.

With inverted arguments the matches_string validation exits early on the length check. Further the wildcard validation does not work as expected.


tobiasbrunner:
This is not a mistake. Wildcards in SANs (or certificates in general) [are not supported](https://docs.strongswan.org/docs/5.9/support/faq.html#_wildcard_certificates).

PR #1663, opened 2023-04-19
ike_reset: update child-sa after update ike initiator spi

pushan01:
**Update child-sa after update ike initiator spi, in case of checkout fail after ike reset.**

If an ike-sa was reset due to retransmit after adopt child-sa from old ike-sa, a new  initiator spi was replaced in the new ike-sa. However, ike-sa-id of child-sa was not updated simultaneously, which may lead to child-sa checkout failed. (`child_sa_manager.c:L260`)

This patch update the ike-sa-id of child-sa after generate new initiator spi when reset ike-sa, solve the issue of checkout failed after reset of ike-sa.

### DIFF
--- a/src/libcharon/sa/ike_sa_manager.c
+++ b/src/libcharon/sa/ike_sa_manager.c
@@ -1732,6 +1732,8 @@ METHOD(ike_sa_manager_t, new_initiator_spi, bool,
 	entry_t *entry;
 	u_int segment;
 	uint64_t new_spi, spi;
+    enumerator_t *enumerator;
+    child_sa_t *child_sa;
 
 	state = ike_sa->get_state(ike_sa);
 	if (state != IKE_CONNECTING)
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks for bringing this to our attention. Note that this is an IKEv1 issue only, `reset()` won't be called on IKEv2 SAs that have CHILD_SAs assigned. Please make sure to use IKEv2 if you can.

I feel updating the CHILD_SAs and triggering events on the bus while holding the segment lock is risky. Another problem is that the `children_migrate()` event will have to be called before updating the `ike_sa_id_t` on the IKE_SA, otherwise, listeners will not be able to migrate from the old ID to the new one. And calling `remove_child_sa()/add_child_sa()` on the same IKE_SA while enumerating might not be the best idea. It's also not really necessary, we can just directly remove/add the CHILD_SAs from `child_sa_manager_t`. However, the latter requires that the ID on the SA is updated, so we have to do that in between these calls. Please have a look at the commit in the _1663-ikev1-reset_ branch.

pushan01:
>Please make sure to use IKEv2 if you can.

Currently, I'm making some tests about IKE/IPSec. I would use IKEv2 if i need in product environment. 

Solution in 1663-ikev1-reset branch is much more better than this PR. I have tested it in my scene and no checkout failure was happend after retrasmit. Thank you very much for your effort in this problem. 👍

tobiasbrunner:
OK, great. Thanks for testing. I'll line that up for the next release.

PR #1748, opened 2023-06-22
[fix][AB-776] crash fix for file descriptors more than 1024

oShcherbininSuper:


zoobibackups:
this is a possible fix for it. we are struggling on this for more then 2 months. but have  no support . @oShcherbininSuper  can you help with us.

PR #1765, opened 2023-06-29
libipsec: Fix routing issue if no local address is within src ts

boisletterink:
Fixes a problem installing traps for source traffic selectors which do not contain/match any of the local addresses.

When installing a route for a source ts that does contain one or more local addresses, a preferred source address should be set in the route (analogous to the `src` option in `ip route add`). This address is used when the host itself sends traffic via that route (in contrast to forwarding it).

When the source ts does not encompass any of the local addresses, the host is not allowed to send traffic via this route and does not need to set the preferred source address. However, the code would just return a failure because it could not find a matching local address. This commit fixes that analogous to how this situation is dealt with by kernel_netlink_ipsec.c from the kernel-netlink module.


boisletterink:
Sorry, need to fix email address

boisletterink:
Should be fixed now. Sorry for the mess.

PR #1766, opened 2023-06-29
libipsec: Fix routing issue if no local address is within src ts (take 2)

boisletterink:
Fixes a problem installing traps for source traffic selectors which do not contain/match any of the local addresses.

When installing a route for a source ts that does contain one or more local addresses, a preferred source address should be set in the route (analogous to the `src` option in `ip route add`). This address is used when the host itself sends traffic via that route (in contrast to forwarding it).

When the source ts does not encompass any of the local addresses, the host is not allowed to send traffic via this route and does not need to set the preferred source address. However, the code would just return a failure because it could not find a matching local address. This commit fixes that analogous to how this situation is dealt with by kernel_netlink_ipsec.c from the kernel-netlink module.

### DIFF
--- a/src/libcharon/plugins/kernel_libipsec/kernel_libipsec_ipsec.c
+++ b/src/libcharon/plugins/kernel_libipsec/kernel_libipsec_ipsec.c
@@ -419,8 +419,16 @@ static bool install_route(private_kernel_libipsec_ipsec_t *this,
 	{
 		traffic_selector_t *multicast, *broadcast = NULL;
 		bool ignore = FALSE;
+		int family;
 
-		this->mutex->unlock(this->mutex);
+		/* At this point, our src ts does not encompass any of our local
+		 * addresses. If the src ts is a broadcast or multicast range, that's
+		 * fine, we just ignore it and do not install a route. If not, we set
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
@cla-bot check

tobiasbrunner:
Thanks for this change and filling out the CLA.

> This commit fixes that analogous to how this situation is dealt with by kernel_netlink_ipsec.c from the kernel-netlink module.

Note that that's not true. The _kernel-netlink_ plugin only installs routes without preferred source IP for passthrough policies. For regular IPsec policies (including trap policies) it won't install a route at all if no IP is found in the local TS. For policy-based IPsec that's generally fine, but not with the route-based approach of _kernel-libipsec_ as without a route nothing will get forwarded via TUN device. So I agree that we should lift the requirement for an IP in the local traffic selector and just install routes in any case.  I've pushed the commit with shortened comment and updated commit message to the _1766-kernel-libipsec-route_ branch. Please let me know if you're fine with those changes.

boisletterink:
Thanks for the review.  I'm fine with the changes!

tobiasbrunner:
Thanks, I've pushed the patch to master.

PR #1856, opened 2023-08-21
Implement Max child SA limit per configuration

anipatel-eng:
This PR seeks to introduce the ability to control the maximum number of child SAs that are allowed to be negotiated by a responder. This is an optional configuration that maintains the current behavior of unlimited child SAs by default.

This is in relation to discussions per https://github.com/strongswan/strongswan/discussions/1725

Duplicate child SA handling and preventing initiating duplicate child SAs was introduced in 5.9.6 and allows the initiator to largely catch and prevent duplicate child SA from being initiated. However, this doesn't help the responder and can hinder a responder that has a limited implementation or a specific use case to only allow a single child SA per IKE SA.

There is currently no provision as confirmed in the discussion thread to limit the number of child SAs negotiated as a responder which leaves the only option to accept or disconnect upon detecting the max child SA count has been exceeded via a listener plugin. This is far from ideal as it only leads to flapping tunnels or undesired multiple child SAs.  Also, the behavior may be caused by misbehavior of a client such as a pre-5.9.6 client that negotiates duplicates not desired by customer config that could lead to the tunnel flapping.

The ideal behavior would be to allow configuration to control the maximum number of child SAs per configuration profile specifically and for this limit to be enforced at negotiation time to reject additional child SAs with an appropriate notification by the responder.

The RFC, in fact, outlines this use case and refers to minimal implementations where limiting to a single Child SA is desired but the current code doesn't allow for this by configuration and this change introduces this as an optional configuration.

 RFC7296 [Section 1.3](https://datatracker.ietf.org/doc/html/rfc7296#section-1.3).  The CREATE_CHILD_SA Exchange:
  ```
    The responder sends a NO_ADDITIONAL_SAS notification to indicate that
     a CREATE_CHILD_SA request is unacceptable because the responder is
     unwilling to accept any more Child SAs on this IKE SA.  This
     notification can also be used to reject IKE SA rekey.  Some minimal
     implementations may only accept a single Child SA setup in the
     context of an initial IKE exchange and reject any subsequent attempts
     to add more.
  ```
  
  Configuration introduced:
    ```connections.<conn>.children.<child>.max_child_sas = 0```

### DIFF
--- a/src/libcharon/config/child_cfg.c
+++ b/src/libcharon/config/child_cfg.c
@@ -183,6 +183,11 @@ struct private_child_cfg_t {
 	 * DS header field copy mode
 	 */
 	dscp_copy_t copy_dscp;
+
+	/**
+	 * Maximum number of child SA pairs to allow on this child_cfg
+	 */
+	uint32_t max_child_sas;
 };
 
 METHOD(child_cfg_t, get_name, char*,
@@ -660,6 +665,12 @@ METHOD(child_cfg_t, set_replay_window, void,
 	this->replay_window = replay_window;
 }
 
+METHOD(child_cfg_t, get_max_child_sas, uint32_t,
+	   private_child_cfg_t *this)
+{
+	return this->max_child_sas;
+}
+
 #define LT_PART_EQUALS(a, b) ({ a.life == b.life && a.rekey == b.rekey && a.jitter == b.jitter; })
 #define LIFETIME_EQUALS(a, b) ({ LT_PART_EQUALS(a.time, b.time) && LT_PART_EQUALS(a.bytes, b.bytes) && LT_PART_EQUALS(a.packets, b.packets); })
 
@@ -717,7 +728,8 @@ METHOD(child_cfg_t, equals, bool,
 		streq(this->updown, other->updown) &&
 		streq(this->interface, other->interface) &&
 		sec_labels_equal(this->label, other->label) &&
-		this->label_mode == other->label_mode;
+        this->label_mode == other->label_mode &&
+        this->max_child_sas == other->max_child_sas;
 }
 
 METHOD(child_cfg_t, get_ref, child_cfg_t*,
@@ -784,6 +796,7 @@ child_cfg_t *child_cfg_create(char *name, child_cfg_create_t *data)
 			.destroy = _destroy,
 			.get_hw_offload = _get_hw_offload,
 			.get_copy_dscp = _get_copy_dscp,
+			.get_max_child_sas = _get_max_child_sas,
 		},
 		.name = strdup(name),
 		.options = data->options,
@@ -815,6 +828,7 @@ child_cfg_t *child_cfg_create(char *name, child_cfg_create_t *data)
 							"%s.replay_window", DEFAULT_REPLAY_WINDOW, lib->ns),
 		.hw_offload = data->hw_offload,
 		.copy_dscp = data->copy_dscp,
+		.max_child_sas = data->max_child_sas,
 	);
 
 	return &this->public;
--- a/src/libcharon/config/child_cfg.h
+++ b/src/libcharon/config/child_cfg.h
@@ -324,6 +324,13 @@ struct child_cfg_t {
 	void (*set_replay_window)(child_cfg_t *this, uint32_t window);
 
 	/**
+	* Get maximum child SA pairs allowed for this child_cfg
+	*
+	* @return              max_child_sas
+	*/
+	uint32_t (*get_max_child_sas)(child_cfg_t *this);
+
+	/**
 	 * Check if an option flag is set.
 	 *
 	 * @param option		option flag to check
@@ -436,6 +443,8 @@ struct child_cfg_create_t {
 	hw_offload_t hw_offload;
 	/** How to handle the DS header field in tunnel mode */
 	dscp_copy_t copy_dscp;
+	/** Maximum number of child SA pairs to allow for this child_cfg */
+	uint32_t max_child_sas;
 };
 
 /**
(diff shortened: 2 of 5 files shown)


tobiasbrunner:
@cla-bot check

Hm, I don't think such a limitation to a specific number is very useful. What would the use case be to set it to a particular value other than 1 or 0? Basing the duplicate check on the name is also far from ideal (there is narrowing and configs could get renamed etc.). And sending back a `NO_ADDITIONAL_SAS` notify does not seem correct either. That would technically not allow the peer to rekey the SA or create CHILD_SAs for other configs/selectors under the same IKE_SA. The RFC says this:
```
   The responder sends a NO_ADDITIONAL_SAS notification to indicate that
   a CREATE_CHILD_SA request is unacceptable because the responder is
   unwilling to accept any more Child SAs on this IKE SA.  This
   notification can also be used to reject IKE SA rekey.  Some minimal
   implementations may only accept a single Child SA setup in the
   context of an initial IKE exchange and reject any subsequent attempts
   to add more.
```
The protocol does not provide a way to convey to a peer that only a single CHILD_SA with particular selectors is wanted. In particular, because the RFC explicitly states:
```
   Note that IKEv2 deliberately allows parallel SAs with the same
   Traffic Selectors between common endpoints.  One of the purposes of
   this is to support traffic quality of service (QoS) differences among
   the SAs.  Hence unlike IKEv1, the combination of the endpoints
   and the Traffic Selectors may not uniquely identify an SA between
   those endpoints, so the IKEv1 rekeying heuristic of deleting SAs on
   the basis of duplicate Traffic Selectors SHOULD NOT be used.
```
So as already suggested in #1725, while not recommended by the RFC, you could technically delete the unwanted CHILD_SA after the fact (the newly created or maybe better the "old" one if the peer lost that and created an additional SA - but it should really be investigated why the duplicate was created in the first place). Of course, that's not ideal if the peer just recreates the SA again (e.g. because of something like `close_action=restart` and a lack of a local duplicate check in older strongSwan versions).

anipatel-eng:
Thanks for the feedback.

Addressing some of the question and comments.

  *   I don’t have a use case necessarily in mind for setting the max child SA limit to something other than 1 necessarily but I thought I’d make the max limit usable in general if someone wanted to be able to limit the number of child SAs
  *   Duplicate check on name: Understood that narrowing and config name changes can occur but the configured intention here is for a named policy to only allow a specific number of child SAs whether narrowed or not and it a policy is renamed the SA count is still per that new policy and would then need to be handled by the application to terminate tunnels as needed.
  *   As for the notify reason being sent back, I now understand the implicit implication on rekeys but didn’t expect that is applies to rekeys but that is may be used for and it wasn’t my intention to block child rekeys so we can shift to using a different notify reason if this really is the case. Any suggestion on an alternate notification type is welcome.

-          Also, I do clearly understand that the RFC specifically allows for multiple parallel tunnels for use cases such as QoS but as you quoted it also allows for minimal implementations:

        Some minimal implementations may only accept a single Child SA setup in the

   context of an initial IKE exchange and reject any subsequent attempts

   to add more.

-          The intention here is for this specific use case to only allow the single Child SA that is created along with the IKE exchange and allow it to be subsequently rekeyed which can clearly be identified vs an additional child SA.

Lastly, as discussed in #1725<https://github.com/strongswan/strongswan/discussions/1725>, terminating unwanted CHILD_SAs after the fact is not ideal and could be better handled by detecting and not allowing it to negotiate in the first place which is the intention of this PR to improve that behavior by allowing configuration of the intention and enforcement of the same in the ideal place. BTW, the cause the issue was already investigated and well understood to be caused by older Strongswan implemented client that can get into a loop of continually negotiating duplicate child SA which was identified as a bug and can be remediated by fixing/updating the client. However, the point of this PR is to allow a server (responder) to protect itself from this behavior that leads a loop of bringing the tunnels up and down and thus unnecessarily expending resources when the behavior can be detected and blocked.

In any case, I’m hoping this provided additional clarity and information to support the PR. I would appreciate any additional feedback and especially the feasibility of this change or a similar modified change to achieve the ideal behavior please.

Thanks,
Animesh

From: Tobias Brunner ***@***.***>
Date: Wednesday, May 15, 2024 at 2:35 AM
To: strongswan/strongswan ***@***.***>
Cc: Animesh Patel (anipatel) ***@***.***>, Mention ***@***.***>
Subject: Re: [strongswan/strongswan] Implement Max child SA limit per configuration (PR #1856)

@cla-bot<https://github.com/cla-bot> check

Hm, I don't think such a limitation to a specific number is very useful. What would the use case be to set it to a particular value other than 1 or 0? Basing the duplicate check on the name is also far from ideal (there is narrowing and configs could get renamed etc.). And sending back a NO_ADDITIONAL_SAS notify does not seem correct either. That would technically not allow the peer to rekey the SA or create CHILD_SAs for other configs/selectors under the same IKE_SA. The RFC says this:

   The responder sends a NO_ADDITIONAL_SAS notification to indicate that

   a CREATE_CHILD_SA request is unacceptable because the responder is

   unwilling to accept any more Child SAs on this IKE SA.  This

   notification can also be used to reject IKE SA rekey.  Some minimal

   implementations may only accept a single Child SA setup in the

   context of an initial IKE exchange and reject any subsequent attempts

   to add more.

The protocol does not provide a way to convey to a peer that only a single CHILD_SA with particular selectors is wanted. In particular, because the RFC explicitly states:

   Note that IKEv2 deliberately allows parallel SAs with the same

   Traffic Selectors between common endpoints.  One of the purposes of

   this is to support traffic quality of service (QoS) differences among

   the SAs.  Hence unlike IKEv1, the combination of the endpoints

   and the Traffic Selectors may not uniquely identify an SA between

   those endpoints, so the IKEv1 rekeying heuristic of deleting SAs on

   the basis of duplicate Traffic Selectors SHOULD NOT be used.

So as already suggested in #1725<https://github.com/strongswan/strongswan/discussions/1725>, while not recommended by the RFC, you could technically delete the unwanted CHILD_SA after the fact (the newly created or maybe better the "old" one if the peer lost that and created an additional SA - but it should really be investigated why the duplicate was created in the first place). Of course, that's not ideal if the peer just recreates the SA again (e.g. because of something like close_action=restart and a lack of a local duplicate check in older strongSwan versions).

PR #1972, opened 2023-11-01
introduce docker

mcluseau:
Hi strongSwan devs :)

I would like to know if there's interest in docker support upstream.

I've made some work here to build alpine-based container images. There's a Dockerfile for the "build from source case" (`Dockerfile`) and another one for the "build from release" (`Dockerfile.release`) with a build-arg to select the release to build.

They have a lot in common so they're an assembly built by `docker/Makefile`.

There's some preliminary work to try to use the existing tests in a docker environment, but it's not finished.

I'm opening this PR to get some feedback, as it's in WIP state.

Thanks!

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+Dockerfile
+Dockerfile.release
+docker/Dockerfile*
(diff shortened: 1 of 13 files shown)


tobiasbrunner:
@cla-bot check

PR #2005, opened 2023-11-20
Extend strongSwan Android frontend with managed configuration support

mpfeiffermway:
This adds support for [managed configuration](https://developer.android.com/work/managed-configurations) to the Android frontend of strongSwan. This allows administrators to remotely configure the strongSwan app, when installed on an Android device enrolled through Android Enterprise.

Additionally, the app has been extended with the ability to install user certificates through the [DevicePolicyManager](https://developer.android.com/reference/android/app/admin/DevicePolicyManager), which is possible on enrolled devices, if the app has been granted the `CERT_INSTALL` [delegate scope](https://developers.google.com/android/management/reference/rest/v1/enterprises.policies#delegatedscope)

The changes also make it possible for an administrator to hide the option to create or use existing unmanaged VPN profiles.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ fuzzing-corpora/
 *.tar.bz2
 *.tar.gz
 .DS_Store
+._.DS_Store
 coverage/
 *.gcno
 *.gcda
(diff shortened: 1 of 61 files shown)


tobiasbrunner:
Refactored versions of these changes were merged with 4be75c5ab14a9983cc075a7a65ce4a4f4717f805.

PR #2016, opened 2023-11-29
Adding support for RFC 4806

jfhren-stormshield:
Hello,

As evoked in https://github.com/strongswan/strongswan/discussions/1845, we added support for RFC 4806.
A new token `connections.<conn>.ocsp` was added and can take 4 values: `never`, `reply`, `request` or `both`.
If `never` is set, the functionality is disabled entirely.
If `reply` is set, request for OCSP status are replied to.
If `request` is set, request for OCSP status are sent.
If `both` is set, request for OCSP status are replied to and request are sent.

A request is answered through the revocation plugin making use of the caching feature.
When a reply is received, it is cached in the same fashion that CRL in CERT payload. It is later checked for validity while taking into account the certificate hash received in the OCSP certreq payload (if any).

As described in RFC 4806 and suggested by @strongX509, OCSP signer certificates from the group `b`  (see section 3.1) can be added in `swanctl/x509ocsp` and while added to OCSP certreq payload.

### DIFF
--- a/src/libcharon/config/ike_cfg.c
+++ b/src/libcharon/config/ike_cfg.c
@@ -97,6 +97,11 @@ struct private_ike_cfg_t {
 	bool certreq;
 
 	/**
+	 * should we send an OCSP status request?
+	 */
+	bool ocsp_certreq;
+
+	/**
 	 * enforce UDP encapsulation
 	 */
 	bool force_encap;
@@ -132,6 +137,12 @@ METHOD(ike_cfg_t, send_certreq, bool,
 	private_ike_cfg_t *this)
 {
 	return this->certreq;
+}
+
+METHOD(ike_cfg_t, send_ocsp_certreq, bool,
+	private_ike_cfg_t *this)
+{
+	return this->ocsp_certreq;
 }
 
 METHOD(ike_cfg_t, force_encap_, bool,
@@ -388,6 +399,7 @@ METHOD(ike_cfg_t, equals, bool,
 	return
 		this->version == other->version &&
 		this->certreq == other->certreq &&
+		this->ocsp_certreq == other->ocsp_certreq &&
 		this->force_encap == other->force_encap &&
 		this->fragmentation == other->fragmentation &&
 		this->childless == other->childless &&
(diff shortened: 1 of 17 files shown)


jfhren-stormshield:
@cla-bot check

tobiasbrunner:
Thanks a lot for this.

I've applied the commit with several changes and refactorings to master. They are as follows:
* Renamed the enum values to make them clearer and avoid the conflict on Windows.
* Only send a non-empty OCSP cert request if we find **self-signed** OCSP-signer certificates. That's basically case (b) mentioned in [section 3.1](https://datatracker.ietf.org/doc/html/rfc4806#section-3.1) of the RFC (in the other cases it makes no sense to send a hash, which the RFC also points out), with the added condition that only certificates flagged as OCSP-signer are considered (i.e. not all self-signed certificates found locally). Which means they either must have that extended key usage flag set or they must be placed in the `x509ocsp` directory so they get flagged when loaded. There was also some weird duplicate hash generation/addition when sending certificate requests.
* Only log received hash values (or certificate subjects) when handling OCSP cert requests, don't apply them as constraints. The sender of an OCSP cert request really doesn't know who'll sign the OCSP responses (e.g. if it has multiple self-signed OCSP-signer certificates, it won't know which will apply to which connection, could even be multiple over the whole chain, or none at all). So using whatever it sends as hard constraint is too restrictive. With self-signed certificates, the hashes could theoretically be useful to select the right (pre-generated) OCSP response or signing certificate if there are multiple during a key rollover, but strongSwan doesn't support this at the moment (and OCSP itself doesn't seem to support such a feature anyway, so this would only be useful if the VPN server generated OCSP responses itself e.g. via `ocsp_responder_t` interface).
* Removed the manual resolution of the certificate chain when sending OCSP responses (which might have been test code or from an earlier iteration) and the corresponding `credential_manager_t::get_issuer_cert()` method. The chain is just enumerated via local auth config.
* Changed the log messages.
* @strongX509 added integration tests in 6f8275ababa99238887071d175ca76cd3801fa35.

PR #2026, opened 2023-12-09
ipsec: strongSwan is not exclusive to Linux kernels

jlduran:
The project is now called "the strongSwan project" rather than "Linux strongSwan" to better reflect this.


tobiasbrunner:
@cla-bot check

PR #2047, opened 2023-12-24
Uses refactored packet filtering code to implement the DHCP plugin on FreeBSD & macOS

sddj:
- Pulled the packet filtering code out of the farp_spoofer.c into shared files pf_handler.c & pf_handler.h
- Updated farp_spoofer.c to use the shared code.
- Updated dhcp_socket.c to use the shared code.
- Implemented the glue code for DHCP in dhcp_socket.c for FreeBSD & macOS
- Updated curl_plugin.c to compile with recently released curl changes

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1853,6 +1853,7 @@ AM_CONDITIONAL(USE_ATTR, test x$attr = xtrue)
 AM_CONDITIONAL(USE_ATTR_SQL, test x$attr_sql = xtrue)
 AM_CONDITIONAL(USE_COUNTERS, test x$counters = xtrue)
 AM_CONDITIONAL(USE_SELINUX, test x$selinux = xtrue)
+AM_CONDITIONAL(USE_PF_HANDLER, test x$dhcp = xtrue -o x$farp = xtrue )
 
 #  other options
 # ---------------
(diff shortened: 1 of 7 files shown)


tobiasbrunner:
Thanks a lot. I've fixed code style issues and refactored it a bit. The updated commit can be found in the _2047-dhcp-on-bsd_ branch. I'm not sure yet about the log messages/levels.

Please let me know what you think.

tobiasbrunner:
@cla-bot check

sddj:
Thanks for the review, refactoring, and reformatting.  The update in the branch looks good to me.  That is definitely a better directory for the pf_handler.* files.

For the debug messages, the information is written so I am happy with the changes.  In my local tests the messages are not frequent, though I only have 3-4 mobile clients connecting.

Do you use a tool to reformat the code?  I would happily use the tool to help keep the formatting consistent.

tobiasbrunner:
OK, great, thanks for the feedback. I've applied the change to master.

> In my local tests the messages are not frequent, though I only have 3-4 mobile clients connecting.

Yeah, I guess we'll have to see if users with larger setups complain :)

> Do you use a tool to reformat the code?

No, I did that manually. There is an `.editorconfig` but that only specifies the very basic whitespace settings. Other than that we have a small [style guide](https://docs.strongswan.org/docs/5.9/devs/programmingStyle.html) in the docs. I guess you used an editor/IDE that used different settings and that automatically reformatted the code, because even copy-n-pasted lines were modified.

PR #2062, opened 2024-01-15
Handle unique identifier allocation overflows gracefully

martinwilli:
Different users in the strongSwan code base use the refcount helpers to allocate incrementing unique values. So far the risk of overflows for these unsigned 32-bit values has been considered mostly theoretical, as it requires a longer uptime and a lot of activity to hit such an overflow.

At least for the Netlink sequence numbers, this is not only theoretical, though, and an overflow has been hit on a productive setup. Unfortunately, the consequences are rather unpleasant, as the response with a zero sequence number can't be matched to the request. This results in the offending thread to block indefinitely while holding the Netlink mutex.

So add a helper to allocate incrementing unique identifiers that checks for overflows and never returns 0. Use it for Netlink sequence numbers and some other potential users affected, namely those allocating IKE_SA/CHILD_SA unique identifiers, marks and interface identifiers.


src/libstrongswan/ipsec/ipsec_types.c
```diff
@@ -150,6 +150,22 @@ bool mark_from_string(const char *value, mark_op_t ops, mark_t *mark)
 	return TRUE;
 }
 
+/*
+ * Described in header
+ */
+uint32_t allocate_unique_mark(void)
+{
+	static refcount_t unique_mark = 0;
```
tobiasbrunner, line 158: Why define this here and not pass the counter? (Would make it e.g. testable.)

tobiasbrunner (commented):
Thanks for this. Looks good, I've just one question (see above).

src/libstrongswan/ipsec/ipsec_types.c
martinwilli: Mostly to align it with the other functions herein, namely `allocate_unique_if_ids()`. It makes the use of these functions somewhat simpler by hiding the underlying counter.

I'm fine with either approach, let me know if I shall change it.

tobiasbrunner: I see. However, that immediately begs the question whether an approach similar to `allocate_unique_mark()` would also make sense for `allocate_unique_if_ids()`. It's probably similarly unlikely to be a problem as with the marks, but would be consistent.

martinwilli: I've updated the changes to use the same mechanic for both mark and interface ID allocation, and handle overflows gracefully in both.

The code duplication is somewhat more obvious now, but I don't see a simple way for unification that is worth the hassle.

tobiasbrunner: Thanks, looks good. I'll merge it to master.

> The code duplication is somewhat more obvious now, but I don't see a simple way for unification that is worth the hassle.

Agreed.

tobiasbrunner:
Thanks, merged to master.

PR #2082, opened 2024-01-26
PKCS11 RSA-PSS Support

strongX509:
Updated the pkcs11.h header based on the latest OpenSC version in order to include new algorithm and struct definitions. Added support for PSS padding in on-chip RSA signatures.


tobiasbrunner:
Looks good. The only thing I noticed is that, similar to the classic schemes, and if supported by the token, there are hash-specific mechanisms (`CKM_SHA*_RSA_PKCS_PSS`) where the token creates the hash. I guess it's a bit slower due to the data that has to be transmitted, but it can avoid having to provide the hash algorithm separately via some other lib. We currently prefer the specific mechanisms (the fallback to software hashing was added via #168), so maybe we could do that here too for consistency.

strongX509:
Yeah, the problem is that OpenSC up to version 0.23 doensn't correctly implement the hash-specific mechanisms (CKM_SHA*_RSA_PKCS_PSS) . Therefore we introduce the **`pkcs11`** plugin option **`use_rsa_pss_hashes`**  which is set to **`no`** by default.

src/libstrongswan/plugins/pkcs11/pkcs11_private_key.c
```diff
@@ -204,11 +213,140 @@ CK_MECHANISM_PTR pkcs11_signature_scheme_to_mech(pkcs11_library_t *p11,
...
+
+			if (scheme == SIGN_RSA_EMSA_PSS)
+			{
+				rsa_pss_params = (rsa_pss_params_t*)params;
+				INIT(rsa_pkcs_pss_params,
+					.sLen = rsa_pss_params->salt_len
+				);
+
+				if (lib->settings->get_bool(lib->settings,
+						"%s.plugins.pkcs11.use_rsa_pss_hashes", FALSE, lib->ns))
```
tobiasbrunner, line 227:
```suggestion
						"%s.plugins.pkcs11.use_rsa_pss_hashers", FALSE, lib->ns))
```

src/libstrongswan/plugins/pkcs11/pkcs11_private_key.c
```diff
@@ -204,11 +213,140 @@ CK_MECHANISM_PTR pkcs11_signature_scheme_to_mech(pkcs11_library_t *p11,
...
+						break;
+					case HASH_SHA3_224:
+						rsa_pkcs_pss_params->mgf = CKG_MGF1_SHA3_224;
+						break;
+					default:
+						free(rsa_pkcs_pss_params);
+						free(mechanism);
+						return NULL;
+				}
+				hash_alg = rsa_pss_params->hash;
```
tobiasbrunner, line 337: I think this should only be assigned if `mechanism->mechanism` is `CKM_RSA_PKCS_PSS`, not when the mechanism is a specific one that includes the hash. Otherwise, the token would get a hash to hash again and wouldn't produce the correct result. `hash_alg` has to be initialized to `HASH_UNKNOWN` for this to work.

tobiasbrunner (changes requested):

tobiasbrunner (approved):

PR #2084, opened 2024-01-30
charon-nm: Allow configuring local traffic selector

dicristina:
Here is a rather trivial patch that allows one to set the local traffic selector by setting a value for the `local-ts` configuration key.  In the absence of `local-ts`  the behavior is the same as before: the local traffic selector is created dynamically.


tobiasbrunner:
Pushed an alternative implementation.

PR #2090, opened 2024-02-07
pki: Added key and cert handles to --ocsp command

strongX509:
OCSP responders might store their signing keys and certificates on a smartcard or a TPM 2.0 device.

### DIFF
--- a/src/pki/commands/ocsp.c
+++ b/src/pki/commands/ocsp.c
@@ -233,7 +233,7 @@ static int ocsp()
 	ocsp_responder_t *index_responder = NULL;
 	linked_list_t *responses = NULL;
 	array_t *index_responders = NULL;
-	chunk_t encoding = chunk_empty, nonce = chunk_empty;
+	chunk_t encoding = chunk_empty, nonce = chunk_empty, handle = chunk_empty;
 	chunk_t issuerNameHash, issuerKeyHash, serialNumber;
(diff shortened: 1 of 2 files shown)


strongX509:
The other **`pki`** commands assign **`-x`**  to **`--keyid`** but this short option is already taken by **`--index`**, so **`-K`** is used instead.

strongX509:
The loading of the OpenSC PKCS#11 module takes about 10 seconds, so that the **`pki --ocsp`** command with smartcard key storage wouldn't be suitable for a static CGI script.

tobiasbrunner (commented):
> The other **`pki`** commands assign **`-x`** to **`--keyid`** but this short option is already taken by **`--index`**, so **`-K`** is used instead.

Unfortunate, but `-K` might have been the more consistent choice to begin with (`-k` => `-K`). I guess we could consider changing `-x` to something else (might not be in much use yet), not sure what, though.

> The loading of the OpenSC PKCS#11 module takes about 10 seconds, so that the **`pki --ocsp`** command with smartcard key storage wouldn't be suitable for a static CGI script.

I see. Yeah, that would probably require something like a `--daemon/-d` option that would handle multiple requests.

Except for some whitespace/alignment issues (I think you mentioned something) this looks OK.

tobiasbrunner (approved):

miladsobhanii (approved):

PR #2097, opened 2024-02-13
openssl: Add missing error checking to ED private key get_encoding method

geedo0:
This applies the same logic found in other private key implementations like
`openssl_ec_private_key.c`.


tobiasbrunner:
Thanks, applied to master.

Issue #1907, opened 2023-09-13
Add support for AWS-LC in the openssl plugin


geedo0:
**Is your feature request related to a problem? Please describe.**
[AWS-LC](https://github.com/aws/aws-lc) is fork of BoringSSL/OpenSSL that is largely API compatible with OpenSSL 1.1.1 and is maintained by AWS. strongSwan's openssl plugin should support AWS-LC as the underlying libcrypto.

Potential benefits from using AWS-LC as the crypto provider:
- Assembly level optimization for AWS EC2 hardware (broadly Intel/ARM processors)
- Alternative path for meeting the requirements of compliance programs such as FIPS 140-2.
- Support for the latest cryptography such as Post-Quantum.

**Describe the solution you'd like**
strongSwan should compile, link, and run successfully when the underlying libcrypto library and headers correspond to AWS-LC. This will be accomplished by making changes to AWS-LC to address missing APIs and adding the minimal number of conditional statements within the openssl plugin to make the compilation successful.

**Describe alternatives you've considered**
We can bias towards more conditional statements in strongSwan and making no changes to AWS-LC. This has the short term benefit of adding support for AWS-LC, but increases the maintenance footprint in strongSwan. Long-term it would be better to improve compatibility in AWS-LC for strongSwan and other projects that may want to build against the library.

**Additional context**
[original strongswan discussion](https://github.com/strongswan/strongswan/discussions/1878)

tobiasbrunner:
So how do you want to proceed? Are you going to prepare a PR?

geedo0:
Sorry, I missed your last comment here. We're currently working on implementing the missing symbols/ciphers in the AWS-LC library. Once that is done, we'll put together a PR of what needs to be contributed upstream to strongSwan. I'm hopeful that there will be minimal changes to strongSwan.

geedo0:
Hi there, we recently finished adding all of the missing symbols in AWS-LC and are ready to look into making this all build successfully. In addition to the symbols listed above, we've since added support for `ChaCha20-Poly1305` and `AES-CCM` via the `EVP_CIPHER API`.

[...]

PR #2103, opened 2024-02-16
openssl: Add conditional macros around SHA_CTX for AWS-LC

geedo0:
AWS-LC is an BoringSSL-based libcrypto implementation. SHA_CTX is declared with
the hash data specified as an array rather than as a field in upstream OpenSSL.
Since AWS-LC builds against C99, we are unable to handle this with anonymous
unions like BoringSSL. The workaround I propose is to add these conditional
macros around the accessors within openssl_sha1_prf. After this change,
everything builds successfully with AWS-LC headers.

Related to Issue #1907


tobiasbrunner:
Thanks, applied to master.

PR #2109, opened 2024-02-20
openssl: Condition out unsupported curves for AWS-LC

geedo0:
AWS-LC lacks support for a number of elliptic curve algorithms so this PR
adds some conditional macros to avoid registering the related plugins. Support
for curves ed448 and x448 is completely absent and are not planned for
implementation as they are no longer recommended for use. While ed25519 is
supported by the library, a single missing API for ASN.1 DER encoding of its
private keys is missing which prevents its use in strongSwan. Future work may
remove this limitation, but for now we will disable the functionality.

Related to Issue #1907


tobiasbrunner:
Thanks, applied to master with some minor changes.

PR #2138, opened 2024-03-04
openssl: Handle BoringSSL-style ASN1_INTEGERs in cert serials

geedo0:
OpenSSL stores the serial number for an X509 certificate as an `ASN1_INTEGER` type. Within BoringSSL (and AWS-LC), the library represents the value of zero as an empty array [1] which is different from OpenSSL which represents it as the 1-byte array [0x00]. Though the value of zero for the certificate serial number is illegal under X.509 [2], we need to handle/encode it consistently within strongSwan. From [3], we infer that the canonical representation of the zero serial is [0x00]. To do this, we introduce `openssl_asn1_int2chunk` to complement the existing string version that allows us to handle the special case for zero instead of always returning a reference to the library-dependent encodings.

Related to Issue #1907

[1] https://github.com/google/boringssl/commit/bdc35b63617f78037768f4897d8835696f02181a
[2] https://datatracker.ietf.org/doc/html/rfc5280#autoid-21
[3] https://github.com/strongswan/strongswan/commit/18082ce2b061da2a12125e535484495d3312e598

### DIFF
--- a/src/libstrongswan/plugins/openssl/openssl_crl.c
+++ b/src/libstrongswan/plugins/openssl/openssl_crl.c
@@ -177,7 +177,7 @@ METHOD(enumerator_t, crl_enumerate, bool,
 		revoked = sk_X509_REVOKED_value(this->stack, this->i);
 		if (serial)
 		{
-			*serial = openssl_asn1_str2chunk(
+			*serial = openssl_asn1_int2chunk(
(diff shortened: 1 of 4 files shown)


tobiasbrunner:
Thanks, looks good. I've applied it to master with a couple of minor code changes (style of the `if`-statement and the declaration of `asn1_zero_int`) and a reformatting of the commit message (was all on one line).

PR #2141, opened 2024-03-05
configure: don't set DL_LIBS to 'none required'

orbea:
This copies the AC_SEARCH_LIBS check from the main strongswan configure.ac.

When building networkmanager-strongswan with slibtool if fails.
```
ld: cannot find none: No such file or directory
ld: cannot find required: No such file or directory
```
This is because configure.ac uses AC_SEARCH_LIBS to find dlopen which sets the value of $ac_cv_search_dlopen to 'none required' which then gets set in DL_LIBS and passed to slibtool.

With GNU libtool it silently ignores the unknown arguments.

Gentoo issue: https://bugs.gentoo.org/914100

### DIFF
--- a/src/frontends/gnome/configure.ac
+++ b/src/frontends/gnome/configure.ac
@@ -35,8 +35,9 @@ AC_PROG_GCC_TRADITIONAL
 AC_FUNC_MEMCMP
 AC_CHECK_FUNCS(select socket uname)
 
-AC_SEARCH_LIBS([dlopen], [dl dld], [], [ac_cv_search_dlopen=])
-AC_SUBST([DL_LIBS], "$ac_cv_search_dlopen")
+LIBS=""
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks for the update and sorry for the delay. Since this PR was originally created, I checked with secunet regarding the requirements for CLAs for contributions to GPL-only components like the NetworkManager plugin. The consensus was that no CLA is required, however, they'd like a Developer Certificate of Origin (DCO). Please refer to [the docs](https://docs.strongswan.org/docs/latest/devs/contributions.html#gpl-only) for details (basically "Signed-off-by" tags in every commit message to consent to the [DCO terms](https://developercertificate.org/)).

If you are fine with that, please update the commit accordingly. Please note that the cla-bot will complain when you push as we currently can't disable it for such PRs.

tobiasbrunner:
Thanks, the patch is in master now. I'll publish a new version of the nm plugin after the 6.0.2 release.

PR #2147, opened 2024-03-06
leak-detective: Add whitelist entries for AWS-LC

geedo0:
AWS-LC (and likely BoringSSL) uses thread specific data to store internal
library state which gets freed via a registered destructor when the thread
terminates. If this thread happens to be the main thread, which runs the
leak-detective evaluation, the detective won't observe the corresponding free
of the related memory and erroneously reports it as a leak.

The two places this happens are:
- `RAND_bytes` for storing internal RNG state.
- `ERR_put_error` for storing the per-thread OpenSSL error queue.

References Issue #1907


tobiasbrunner:
Thanks, applied to master.

PR #2151, opened 2024-03-07
github: Add AWS-LC CI job

geedo0:
AWS-LC is an OpenSSL derivative which can be used with the openssl plugin. This
PR adds a CI job that resembles the openssl-3 test case. It downloads the
source tarball for an AWS-LC release, builds that source using CMake/Ninja, and
then builds/tests strongSwan using the same technique used by openssl-3.

Related to Issue #1907

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -91,7 +91,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest, ubuntu-20.04 ]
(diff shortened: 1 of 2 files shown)


tobiasbrunner:
Applied to master, thanks. I noticed that there was a 1.22 release a few days ago, so I updated the commit. I also let the script fail if the AWS-LC build fails.

geedo0:
Thanks!

> Is it possible to implement this via AWS-LC/BoringSSL?

I'll certainly take a look and see what we can do on the AWS-LC side.

PR #2279, opened 2024-06-06
Add `Wants=` dependencies to Systemd units

matteocarnelos:
This PR adds `Wants=` dependencies to the strongSwan Systemd units to prevent `network-online.target` and `syslog.target` from not being enabled if no other services have a `Wants=` dependency pointing to them, resulting in a possibly unwanted service startup order.

### DIFF
--- a/init/systemd-starter/strongswan-starter.service.in
+++ b/init/systemd-starter/strongswan-starter.service.in
@@ -1,6 +1,7 @@
 [Unit]
 Description=strongSwan IPsec IKEv1/IKEv2 daemon using ipsec.conf
 After=syslog.target network-online.target
+Wants=syslog.target network-online.target
 
 [Service]
 ExecStart=@SBINDIR@/@IPSEC_SCRIPT@ start --nofork
--- a/init/systemd/strongswan.service.in
+++ b/init/systemd/strongswan.service.in
@@ -1,6 +1,7 @@
 [Unit]
 Description=strongSwan IPsec IKEv1/IKEv2 daemon using swanctl
 After=network-online.target
+Wants=network-online.target
 
 [Service]
 Type=notify


tobiasbrunner:
@cla-bot check 

Thanks, applied to master.

matteocarnelos:
@tobiasbrunner there's one detail regarding this PR that's worth mentioning: these changes might alter the service startup order on some systems.

In particular, systems that previously had no enabled services with the `Wants=` dependency to `network-online.target` will have the strongSwan service introduce that dependency, making Systemd start `network-online.target` and delay the start of the strongSwan service. If not overridden by custom values, the delay will be, in the worst case, up to the default timeout of `systemd-networkd-wait-online.service`, i.e., [120 seconds](https://www.freedesktop.org/software/systemd/man/latest/systemd-networkd-wait-online.service.html#--timeout=SECS). Systems that already have enabled services with `Wants=` dependencies to `network-online.target` will not be affected.

In general terms, it's correct and common to specify both the `After=` and `Wants=` dependencies, as stated in the [Systemd manual](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#Before=):
> It is a common pattern to include a unit name in both the `After=` and `Wants=` options, in which case the unit listed will be started before the unit that is configured with these options.

However, I don't know if in this case this can be considered a breaking change and therefore should not be shipped in the next minor release.

tobiasbrunner:
What's a system without `network-online.target` doing starting a network service like strongSwan? (I really don't know much about systemd, so please excuse my ignorance.)

> However, I don't know if in this case this can be considered a breaking change and therefore should not be shipped in the next minor release.

For what it's worth, the next release won't be a minor release.

matteocarnelos:
> What's a system without `network-online.target` doing starting a network service like strongSwan? (I really don't know much about systemd, so please excuse my ignorance.)

It's likely a "misconfigured" system, where `network-online.target` doesn't work as it should.

In conclusion, given also the fact that the next release will be a major release, I think we should keep this change. Thanks for the help!

PR #2297, opened 2024-06-18
error-notify: disconnect socket on ETIMEDOUT

t-beckmann:
When encountering a timeout the error notify socket needs to be closed so that resources can be reclaimed.


tobiasbrunner:
Hm, `ETIMEDOUT` is not documented as possible error for Unix sockets. So how did you trigger this? Did you switch to TCP?

t-beckmann:
Yes. Actually we use the error-notify socket as a vehicle for health checking, just ensuring the process keeps running. The check connects and then drops the connection. On next occasion the timeout is detected which needs to release it's end of the connection.

Chances are this is a somewhat questionable use. Yet, can't see an argument against it.

Issue #2309, opened 2024-06-24
test ./testing/ failed in current master [d02aea9c2c]


muvarov:
**System (please complete the following information):**
 - OS:  Ubuntu 24.04 LTS
 - Kernel version: Host Ubuntu 24.04 LTS, Guest images: linux-5.15.1 + ha-5.15-abicompat.patch + ./testing/config/kernel/config-5.15
 - strongSwan version(s):  Git 5.9.14+  current commit: commit d02aea9c2cef7c9509fe61f674d3b468e5c7a12f
 - Tested/confirmed with the latest version:  yes (issue with the latest master).

**Describe the bug**
All testing/ tests are failed.  Test build all env, start VMs, can ping each other. But no IPsec tunnel. SAs are missing from the log

**To Reproduce**
1. cd testing
2. ./make-testing
3. ./do-tests or ssh to moon and check no IPsec connection created.

**Expected behavior**
ipsec statusall - should print out SAs and connections.

**Logs/Backtraces**
Jun 24 09:14:17 moon systemd[1]: Started strongswan.service - strongSwan IPsec IKEv1/IKEv2 daemon using swanctl.
Jun 24 09:14:17 moon charon-systemd: 00[DMN] SIGTERM received, shutting down
Jun 24 09:14:17 moon systemd[1]: Stopping strongswan.service - strongSwan IPsec IKEv1/IKEv2 daemon using swanctl...
Jun 24 09:14:17 moon charon-systemd[2211]: No leaks detected, 963 suppressed by whitelist
Jun 24 09:14:17 moon systemd[1]: strongswan.service: Deactivated successfully.
Jun 24 09:14:17 moon systemd[1]: Stopped strongswan.service - strongSwan IPsec IKEv1/IKEv2 daemon using swanctl.
Jun 24 09:14:17 moon systemd[1]: Starting strongswan.service - strongSwan IPsec IKEv1/IKEv2 daemon using swanctl...
Jun 24 09:14:17 moon charon-systemd: 00[DMN] Starting charon-systemd IKE daemon (strongSwan 5.9.14, Linux 5.15.1, x86_64)
Jun 24 09:14:17 moon (udev-worker)[2212]: Network interface NamePolicy= disabled on kernel command line.
Jun 24 09:14:17 moon charon-systemd: 00[CFG] loading ca certificates from '/etc/ipsec.d/cacerts'
[...]

PR #2310, opened 2024-06-24
testing: fix scp

muvarov:
On current ssh to test machines works, but
scp fails. Using lagacy scp protocol -O option
makes it work. Also this error was hidden due
to not cheking scp return code and mute all
errors. Add -e to trap script on any errors.

This fixes: https://github.com/strongswan/strongswan/issues/2309

### DIFF
--- a/testing/hosts/default/etc/ssh/sshd_config
+++ b/testing/hosts/default/etc/ssh/sshd_config
@@ -13,3 +13,4 @@ PrintLastLog no
 UsePAM no
 AcceptEnv LANG LC_*
 AcceptEnv LEAK_DETECTIVE_LOG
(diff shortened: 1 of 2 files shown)


tobiasbrunner:
Seems the default has changed with newer versions of OpenSSH. Please try if adding

    Subsystem	sftp	/usr/lib/openssh/sftp-server

to `testing/hosts/default/etc/ssh/sshd_config` works instead (requires rebuilding the guest images via `testing/scripts/build-guestimages`). That seems the more correct fix than forcing the clients to the old mode.

muvarov:
yes "Subsystem	sftp" looks like better variant. Patches updated.

PR #2384, opened 2024-08-04
Decoupling server code from client. Fix ~6 year old issue.

jobs-git:
Fixes: https://askubuntu.com/questions/1222523/how-to-connect-to-an-ipsec-vpn-with-a-short-20chars-psk

Server should implement config based password restriction


tobiasbrunner:
This is a duplicate of #145.

PR #2396, opened 2024-08-10
ikev2: ike_config: fill attribute also as initiator

lynxis:
The the attribute array of the ike_sa as initiator. Some plugins may need to know what the client requested to pass the requested config requests towards an additional backend.

### DIFF
--- a/src/libcharon/sa/ikev2/tasks/ike_config.c
+++ b/src/libcharon/sa/ikev2/tasks/ike_config.c
@@ -111,7 +111,7 @@ static configuration_attribute_t *build_vip(host_t *vip)
 /**
  * Handle a received attribute as initiator
  */
-static void handle_attribute(private_ike_config_t *this,
+static void handle_attribute_initiator(private_ike_config_t *this,
 							 configuration_attribute_t *ca)
 {
 	attribute_handler_t *handler = NULL;
@@ -139,6 +139,13 @@ static void handle_attribute(private_ike_config_t *this,
 							handler, ca->get_type(ca), ca->get_chunk(ca));
 }
 
+static void handle_attribute_responder(private_ike_config_t *this,
+							 configuration_attribute_t *ca)
+{
+	this->ike_sa->add_configuration_attribute(this->ike_sa,
+							NULL, ca->get_type(ca), ca->get_chunk(ca));
+}
+
 /**
  * process a single configuration attribute
  */
@@ -158,7 +165,7 @@ static void process_attribute(private_ike_config_t *this,
 		{
 			if (this->initiator && !this->vip_requested)
 			{
-				handle_attribute(this, ca);
+				handle_attribute_initiator(this, ca);
 				return;
 			}
 
@@ -191,7 +198,11 @@ static void process_attribute(private_ike_config_t *this,
 		{
 			if (this->initiator)
 			{
-				handle_attribute(this, ca);
+				handle_attribute_initiator(this, ca);
+			}
+			else
+			{
+				handle_attribute_responder(this, ca);
 			}
 		}
 	}


lynxis:
I fixed the commit message.
I mean if there is already an API for the initiator, I would try to use it also for the responder
Should I change to use the message() hook?

The ePDG need to know in advance which attributes has been requested to pass it further to the backend.

lynxis:
>> The API has a different purpose, though. It stores the assigned attributes on the initiator, not the requested ones.

But why can't it extended to support the other side as well? IMHO it would fit the concept here.

> > The ePDG need to know in advance which attributes has been requested to pass it further to the backend.
> 
> What does it need to know exactly? The type? The value as well (if there are actual values)? Is that documented somewhere?

The UE (smartphone) requests DNS servers, P-CSCF server (voip endpoint), IPv4 address or IPv6 address over it. It is a part of the IKE_AUTH Request and must be forwarded to the AAA server.

The TS 33.400 Figure 8.2.2-1: Tunnel full authentication and authorization explains it (a little bit).

Screenshot_20250710_232517

## References

* [rfc7651](https://www.rfc-editor.org/rfc/rfc7651.txt)
* [TS 33.400](https://www.etsi.org/deliver/etsi_ts/133400_133499/133402/18.01.00_60/ts_133402v180100p.pdf)

tobiasbrunner:
> But why can't it extended to support the other side as well? IMHO it would fit the concept here.

It's a different concept (storing requested attributes instead of assigned attributes). It would require a new or modified API for something that's generally not needed.

> It is a part of the IKE_AUTH Request and must be forwarded to the AAA server.

What exactly must be forwarded? And how (e.g. how would that request be encoded in RADIUS)? And what is returned and in what form?

Generally, configuration attributes can always just be assigned to clients without having received an explicit request. The trigger is the request for a virtual IP (IPv4 or IPv6, or both), but stuff like DNS servers or P-CSCF servers can just be sent along with it. For instance, defined via `p-cscf` keyword of the [`attr` plugin](https://docs.strongswan.org/docs/latest/plugins/attr.html). Interestingly, the `vici` plugin currently doesn't support that keyword in [pools sections](https://docs.strongswan.org/docs/latest/swanctl/swanctlConf.html#_pools), so it would have to configured numerically there but it's still possible. So maybe you are unnecessarily complicating things.

How to get additional attributes from a RADIUS server to the IKE client is a different issue. The `eap-radius` plugin currently only [handles virtual IPs and DNS/NBNS server attributes](https://docs.strongswan.org/docs/latest/plugins/eap-radius.html#_radius_attribute_forwarding) (plus some Cisco Unity ones). So that might need some code to handle P-CSCF servers if there exists a RADIUS attribute that could be converted.

lynxis:
To give a little bit more context.
At osmocom, we've created a combined ePDG/AAA which is called [osmo-epdg](https://osmocom.org/projects/osmo-epdg).
It contains the following components:

- strongswan (IKEv2, EAP, netlink to the kernel for ESP)
- osmo-epdg, a strongswan plugin (to communicates towards erlang: auth data, config data and management)
- osmo-epdg, a erlang daemon (Diameter, GTPv2)
- (some ip routing and firewall configuration).

The term AAA is ambiguous in this context. AAA has a different meaning in the ePDG/VoWifi context, even it does
similar things as a AAA in the Radius context.
The AAA in the context of the ePDG/VoWifi is speaking Diameter and GTPv2 and not Radius.

> What exactly must be forwarded? And how (e.g. how would that request be encoded in RADIUS)? And what is returned and in what form?

I mixed some parts here, it is not required while doing the EAP auth.
**To clarify:**
To authenticate an initiator/phone, it must forwarded the APN (given as Identification Payload, Responder) and
many of EAP attributes (but this is done in a different way, the osmo-epdg plugin will use the eap-aka api).

But before strongswan can sent the final IKE_AUTH Response (direct after the EAP Auth Success), it must forwarded all requested CFG Attributes,
including IPv4/IPv6 DNS, P-CSCF, ADDRESS over GTPv2 (GPRS Tunneling Protocol v2) to an external component (PGW) and wait for the GTPv2 Response.

So far I understand, this is required because 3GPP specs says the final IKE_AUTH_RESP contains all 
CFG Attributes replies.

You could think of GTPv2 in this regard as DHCP. Lets imagine strongswan would need to do a DHCP Request and to do a proper DHCP Request, it must map all CFG Attributes to DHCP options.
GTPv2 behaves similar (with a lot more complexity and features).

For sure, I could just assign Attribute, but I should forward the attributes between the initiator and the backend to comply to the spec.

> The trigger is the request for a virtual IP (IPv4 or IPv6, or both), but stuff like DNS servers or P-CSCF servers can just be sent along with it.

I think the virtual IP hook is too late, because it is called when the phone/initiator requests a CHILD_SA, correct?

> For instance, defined via p-cscf keyword of the [attr plugin](https://docs.strongswan.org/docs/latest/plugins/attr.html). Interestingly, the vici plugin currently doesn't support that keyword in [pools sections](https://docs.strongswan.org/docs/latest/swanctl/swanctlConf.html#_pools), so it would have to configured numerically there but it's still possible. So maybe you are unnecessarily complicating things.

P-CSCF is usually assigned dynamic via GTPv2. In the case you would like to setup a VoWifi without all the 3GPP specific protocols, it could be useful to support p-cscf keyword there.

To come back to your recommendation of using a message hook(). I guess I've to copy most of the parts of  process_payloads(), even it would be great to have Cfg Request parsing somehow in the library. But fine for me.

### Reference

The 3GPP spec says that the IKE_AUTH_RESP after the EAP Authentication succeeds, which finally authenticates the phone/initiator must contain already the CFG Attributes. This is not well documented.

([TS 33.402](https://www.etsi.org/deliver/etsi_ts/133400_133499/133402/18.01.00_60/ts_133402v180100p.pdf), Figure 8.2.2-1, Step 13-15).

Screenshot_20250716_135151

```
13. The ePDG checks the correctness of the AUTH received from the UE. At this point the UE is authenticated. In
case S2b is used, PMIP signalling between ePDG and PDN GW can now start, as specified in TS 23.402 [5].
The ePDG continues with the next step in the procedure described here only after successful completion of the
PMIP binding update procedure.
14. The ePDG calculates the AUTH parameter which authenticates the second IKE_SA_INIT message. The ePDG
shall send the assigned Remote IP address in the configuration payload (CFG_REPLY).
15 The AUTH parameter is sent to the UE together with the configuration payload, security associations and the
rest of the IKEv2 parameters and the IKEv2 negotiation terminates
```

tobiasbrunner:
> For sure, I could just assign Attribute, but I should forward the attributes between the initiator and the backend to comply to the spec.

Just send one in your request to the backend if that's necessary. What's the point of "forwarding" an empty request?  Or are there clients that only request some of the attributes? Also, won't the backend provide attributes that were not explicitly requested if configured for a client? And is this requirement and the conversion from configuration attribute specified somewhere?

> I think the virtual IP hook is too late, because it is called when the phone/initiator requests a CHILD_SA, correct?

No, the virtual IP is requested and assigned during IKE_AUTH (it's one of those configuration attributes and the only one TS 33.402 mentions).


> To come back to your recommendation of using a message hook(). I guess I've to copy most of the parts of process_payloads(), even it would be great to have Cfg Request parsing somehow in the library. But fine for me.

You don't need to do any parsing. You just have to get the configuration payload (`message->get_payload(message, PLV2_CONFIGURATION)`) and if there is one, cast it to `cp_payload_t` and enumerate the attributes (`create_attribute_enumerator()`). But again, you might not need to do that.

PR #2447, opened 2024-09-10
botan: add missing include for build flags

fgirlich:
`botan/build.h` defines all `BOTAN_HAS_*` build feature flags. With botan-2 this used to included indirectly via `botan_util.h` -> `botan/ffi.h`.

Botan-3 no longer includes `botan/build.h` from `botan/ffi.h`. It must now be included explicitly. This is already done in all other places that access botans feature flags. It seemingly was forgotten here. This omission causes loading of certificates and keys to fail, even though botan supports the required crypto algorithms.


tobiasbrunner:
Ran into this myself, fixed with 0a3889086decb7364642cf0c7849b28a40dadd32.

PR #2475, opened 2024-10-01
Add support to log timestamp in microseconds

laura-zelenku:


tobiasbrunner:
@cla-bot check

Thanks, looks pretty good. 

I feel the configuration could be simplified if we'd have only a single option for those additions (e.g. `time_precision = ms|us`, default empty) as being able to enable both seems weird. The old option could still be supported as fallback.

I also don't like the growing list of unnamed `bool` arguments for that `set_options()` method.

So I've pushed a commit that changes both things on top of yours to the `2475-logger-microseconds` branch, which I additionally rebased to the `2222-json-filelog` branch as there were some conflicts.

laura-zelenku:
> @cla-bot check
> 
> Thanks, looks pretty good.
> 
> I feel the configuration could be simplified if we'd have only a single option for those additions (e.g. `time_precision = ms|us`, default empty) as being able to enable both seems weird. The old option could still be supported as fallback.
> 
> I also don't like the growing list of unnamed `bool` arguments for that `set_options()` method.
> 
> So I've pushed a commit that changes both things on top of yours to the `2475-logger-microseconds` branch, which I additionally rebased to the `2222-json-filelog` branch as there were some conflicts.

Thank you @tobiasbrunner

laura-zelenku:
@tobiasbrunner do I need to prepare PR based on your branch `2475-logger-microseconds` or do you proceed with it yourself?

tobiasbrunner:
Thanks for asking. No, you don't have to do anything, I'll close this PR once I merge the branch.

Issue #2228, opened 2024-04-30
Add support for the post-quantum ML-KEM KE algorithm in openssl plugin


geedo0:
[Original Discussion Thread](https://github.com/strongswan/strongswan/discussions/2175)

**Is your feature request related to a problem? Please describe.**

Post-Quantum (PQ) cryptography aims to mitigate the threat quantum computing poses to confidentiality of our data in transit. In mid-2024, NIST is expected to standardize the PQ algorithm ML-KEM (formerly known as Kyber). An [IETF draft](https://datatracker.ietf.org/doc/draft-kampanakis-ml-kem-ikev2/) has been submitted for registering this algorithm in IANA's IKEv2 parameter registry. We would like to contribute an additional implementation of this algorithm in the openssl plugin.

Currently, strongSwan-6 supports this algorithm in the oqs plugin by using the Open Quantum Safe library `liboqs` directly. This library comes with a strong [disclaimer](https://github.com/open-quantum-safe/liboqs?tab=readme-ov-file#limitations-and-security) around usage in a production. While there are [talks](https://www.linuxfoundation.org/press/announcing-the-post-quantum-cryptography-alliance-pqca) to transition the OQS project to something more production ready with its recent membership in the Linux Foundation, it's still unclear what exactly that means (new library, supporting PQ in existing work, or improving liboqs itself). Some signs signal towards continued development of OQS's [OpenSSL-3 provider](https://github.com/open-quantum-safe/oqs-provider) which uses the OpenSSL KEM API.

This feature request proposes using [AWS-LC](https://github.com/aws/aws-lc) (an OpenSSL derivative) as a means to provide PQ confidentiality while providing a path toward future PQ algorithm implementations in upstream OpenSSL.

**Describe the solution you'd like**

The openssl plugin can be updated ([here](https://github.com/strongswan/strongswan/blob/7859f5ddd8a9b73d69c8911003a1e32e7e8b7797/src/libstrongswan/plugins/openssl/openssl_plugin.c#L374)) to conditionally provide support for the 3 different [ML-KEM algorithms](https://github.com/strongswan/strongswan/blob/six-beta/src/libstrongswan/crypto/key_exchange.h#L71-L74) KE's. It will register an implementation which will be able to handle arbitrary KEM algorithms.

[...]

PR #2490, opened 2024-10-09
Add ML-KEM support to openssl plugin with AWS-LC

geedo0:
This registers support for the ML_KEM_{512,768,1024} key exchange algorithms in the `openssl` plugin when built using AWS-LC as the libcrypto. To do this, we introduce the `openssl_kem` source files which implement the key exchange algorithm using the Key Encapsulation Mechanism (KEM) API. Future KEM algorithms can be implemented generically using this interface by substituting the appropriate NIDs. It also supports both seeded (via DRBG) and unseeded modes depending on the user's requirements for KATs or entropy sources. It should be noted that this does not add support for KEM algorithms within upstream OpenSSL and is API incompatible. Future work will need to condition out the incompatibilities as-appropriate. However, the high-level logic should be the same for all KEMs and KEM APIs.

Related to Issue #2228


src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
+/*
+ * Copyright (C) 2024 Amazon.com Inc. or its affiliates.
```
tobiasbrunner, line 2: We only have your individual CLA on file (no entity CLA for Amazon), so I feel it would be better if you used your own name here.

src/libstrongswan/plugins/openssl/openssl_plugin.c
```diff
@@ -549,6 +550,13 @@ METHOD(plugin_t, get_features, int,
 			PLUGIN_PROVIDE(KE, MODP_768_BIT),
 			PLUGIN_PROVIDE(KE, MODP_CUSTOM),
 #endif
+#ifdef OPENSSL_IS_AWSLC
```
tobiasbrunner, line 553: Is there a specific `OPENSSL_NO_*` constant for ML-KEM? Or is AWS-LC always built with ML-KEM (in versions that support it)?

src/libstrongswan/plugins/openssl/openssl_kem.h
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (C) 2024 Amazon.com Inc. or its affiliates.
```
tobiasbrunner, line 2: Same as with the `.c` file.

src/libstrongswan/plugins/openssl/openssl_kem.h
```diff
@@ -0,0 +1,47 @@
...
+#define OPENSSL_KEM_H_
+
+typedef struct openssl_kem_t openssl_kem_t;
+
+#include <crypto/key_exchange.h>
+
+/**
+ * Implementation of KEM algorithms using OpenSSL.
+ */
+struct openssl_kem_t {
```
tobiasbrunner, line 32: This indirection is not really necessary. `openssl_kem_create()` can just return a `key_exchange_t*` (see e.g. the example in `wolfssl_kem.h` or `botan_kem.h`). Also see the comments regarding the implementation.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+
+
+#include "openssl_kem.h"
+
+#include <openssl/evp.h>
+
+#ifdef OPENSSL_IS_AWSLC
+#include <openssl/experimental/kem_deterministic_api.h>
+
+#include <crypto/key_exchange.h>
```
tobiasbrunner, line 25: Already included in the header.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+#include <crypto/drbgs/drbg.h>
```
tobiasbrunner, line 26: Also gets pulled in via the header as `key_exchange.h` includes it.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	{
+		EVP_PKEY_CTX_free(ctx);
+		return FALSE;
+	}
+	if (drbg)
+	{
+		if (!EVP_PKEY_keygen_deterministic(ctx, &this->pkey, NULL, &seed_length))
+		{
+			EVP_PKEY_CTX_free(ctx);
+			EVP_PKEY_free(this->pkey);
```
tobiasbrunner, line 117: Does this function actually modify the key for this use case (determining the seed length) at all (and especially if it fails)?

More importantly, the key is freed in `destroy()` anyway, which probably results in a double free if any of these function calls fails (the pointer is not set to NULL). So please remove all these `EVP_PKEY_free(this->pkey)` calls.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+			return FALSE;
+		}
+		seed = chunk_alloc(seed_length);
+		drbg->generate(drbg, seed.len, seed.ptr);
```
tobiasbrunner, line 121: Probably should add this to the `if` statement to check the return value.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+		{
+			EVP_PKEY_CTX_free(ctx);
+			EVP_PKEY_free(this->pkey);
+			chunk_clear(&seed);
+			return FALSE;
+		}
+	}
+	else
+	{
+		if (!EVP_PKEY_keygen(ctx, &this->pkey))
```
tobiasbrunner, line 132: Combine these lines via `else if`.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+			EVP_PKEY_free(this->pkey);
+			return FALSE;
+		}
+	}
+	EVP_PKEY_CTX_free(ctx);
+	chunk_clear(&seed);
+	return TRUE;
+}
+
+static bool openssl_kem_get_encoded_public_key(private_openssl_kem_t *this,
```
tobiasbrunner, line 144: Add a short comment.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+																							 chunk_t *out)
```
tobiasbrunner, line 145: Indentation is off here (we use a tab size of 4 in case you maybe used 2).

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+{
+	chunk_t pkey_chunk = chunk_empty;
+	size_t public_key_length = 0;
+	if (!this->pkey)
```
tobiasbrunner, line 149: Add an empty line before this.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	{
+		return FALSE;
+	}
+	EVP_PKEY_get_raw_public_key(this->pkey, NULL, &public_key_length);
+	pkey_chunk = chunk_alloc(public_key_length);
+	if (!EVP_PKEY_get_raw_public_key(this->pkey, pkey_chunk.ptr,
+																	 &public_key_length))
+	{
+		chunk_clear(&pkey_chunk);
```
tobiasbrunner, line 158: `chunk_free()` is fine here for the public key.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	chunk_t pkey_chunk = chunk_empty;
+	size_t public_key_length = 0;
+	if (!this->pkey)
+	{
+		return FALSE;
+	}
+	EVP_PKEY_get_raw_public_key(this->pkey, NULL, &public_key_length);
+	pkey_chunk = chunk_alloc(public_key_length);
+	if (!EVP_PKEY_get_raw_public_key(this->pkey, pkey_chunk.ptr,
+																	 &public_key_length))
```
tobiasbrunner, line 156: Indentation again.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	if (this->ciphertext.ptr != NULL)
+	{
+		*value = chunk_clone(this->ciphertext);
+		return TRUE;
+	}
+
+	/* initiator action */
+	if (!this->pkey)
+	{
+		if (!openssl_kem_generate_pkey(this))
```
tobiasbrunner, line 178: Can be combined.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+{
+	if (this->shared_secret.ptr != NULL)
+	{
+		*secret = chunk_clone(this->shared_secret);
+		return TRUE;
+	}
+	return FALSE;
+}
+
+static bool openssl_kem_decapsulate(private_openssl_kem_t *this,
```
tobiasbrunner, line 197: Comment again.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+		if (!openssl_kem_generate_pkey(this))
+		{
+			return FALSE;
+		}
+	}
+	return openssl_kem_get_encoded_public_key(this, value);
+}
+
+METHOD(key_exchange_t, get_shared_secret, bool, private_openssl_kem_t *this,
+			 chunk_t *secret)
```
tobiasbrunner, line 187: Indentation again.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	}
+	return FALSE;
+}
+
+static bool openssl_kem_decapsulate(private_openssl_kem_t *this,
+																		chunk_t ciphertext)
+{
+	EVP_PKEY_CTX *ctx = NULL;
+	size_t shared_secret_length = 0;
+	ctx = EVP_PKEY_CTX_new(this->pkey, NULL);
```
tobiasbrunner, line 202: Add an empty line before this.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+		*secret = chunk_clone(this->shared_secret);
+		return TRUE;
+	}
+	return FALSE;
+}
+
+static bool openssl_kem_decapsulate(private_openssl_kem_t *this,
+																		chunk_t ciphertext)
+{
+	EVP_PKEY_CTX *ctx = NULL;
```
tobiasbrunner, line 200: Not necessary to initialize as it's directly assigned below.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	size_t shared_secret_length = 0;
+	ctx = EVP_PKEY_CTX_new(this->pkey, NULL);
+	if (!ctx)
+	{
+		return FALSE;
+	}
+	EVP_PKEY_decapsulate(ctx, NULL, &shared_secret_length, ciphertext.ptr,
```
tobiasbrunner, line 207: Check the return value?

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+											 ciphertext.len);
```
tobiasbrunner, line 208: Indentation

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	this->shared_secret = chunk_alloc(shared_secret_length);
+	if (!EVP_PKEY_decapsulate(ctx, this->shared_secret.ptr,
+														&this->shared_secret.len, ciphertext.ptr,
+														ciphertext.len))
```
tobiasbrunner, line 212: Indentation.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+		EVP_PKEY_CTX_free(ctx);
+		chunk_clear(&this->shared_secret);
+		return FALSE;
+	}
+	EVP_PKEY_CTX_free(ctx);
+	return TRUE;
+}
+
+static bool openssl_kem_encapsulate(private_openssl_kem_t *this,
+																		chunk_t public_key)
```
tobiasbrunner, line 223: Comment and indentation.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+{
+	EVP_PKEY_CTX *ctx = NULL;
+	EVP_PKEY *pkey = NULL;
```
tobiasbrunner, line 226: Not necessary.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	size_t shared_secret_length = 0;
+	size_t ciphertext_length = 0;
+	size_t seed_length = 0;
+	chunk_t seed = chunk_empty;
+	drbg_t *drbg = this->thread_drbg->get(this->thread_drbg);
+
+	pkey = EVP_PKEY_kem_new_raw_public_key(openssl_kem_get_nid(this),
+																				 public_key.ptr, public_key.len);
```
tobiasbrunner, line 234: Indentation.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	if (!ctx)
+	{
+		EVP_PKEY_free(pkey);
+		return FALSE;
+	}
+	if (drbg)
+	{
+		EVP_PKEY_encapsulate_deterministic(ctx, NULL, &ciphertext_length, NULL,
+																			 &shared_secret_length, NULL,
+																			 &seed_length);
```
tobiasbrunner, line 249: Indentation.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	}
+	ctx = EVP_PKEY_CTX_new(pkey, NULL);
+	if (!ctx)
+	{
+		EVP_PKEY_free(pkey);
+		return FALSE;
+	}
+	if (drbg)
+	{
+		EVP_PKEY_encapsulate_deterministic(ctx, NULL, &ciphertext_length, NULL,
```
tobiasbrunner, line 247: Check the return value?

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+																			 &shared_secret_length, NULL,
+																			 &seed_length);
+		this->shared_secret = chunk_alloc(shared_secret_length);
+		this->ciphertext = chunk_alloc(ciphertext_length);
+		seed = chunk_alloc(seed_length);
+		drbg->generate(drbg, seed.len, seed.ptr);
```
tobiasbrunner, line 253: Add this to the `if` statement.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+		if (!EVP_PKEY_encapsulate_deterministic(ctx, this->ciphertext.ptr,
+																						&this->ciphertext.len,
+																						this->shared_secret.ptr,
+																						&this->shared_secret.len, seed.ptr,
+																						&seed.len))
```
tobiasbrunner, line 258: Indentation.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+			chunk_clear(&this->shared_secret);
+			chunk_clear(&this->ciphertext);
+			chunk_clear(&seed);
+			return FALSE;
+		}
+	}
+	else
+	{
+		EVP_PKEY_encapsulate(ctx, NULL, &ciphertext_length, NULL,
+												 &shared_secret_length);
```
tobiasbrunner, line 271: Indentation.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+			EVP_PKEY_CTX_free(ctx);
+			chunk_clear(&this->shared_secret);
+			chunk_clear(&this->ciphertext);
+			chunk_clear(&seed);
+			return FALSE;
+		}
+	}
+	else
+	{
+		EVP_PKEY_encapsulate(ctx, NULL, &ciphertext_length, NULL,
```
tobiasbrunner, line 270: Check the return value?

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+												 &shared_secret_length);
+		this->shared_secret = chunk_alloc(shared_secret_length);
+		this->ciphertext = chunk_alloc(ciphertext_length);
+		if (!EVP_PKEY_encapsulate(ctx, this->ciphertext.ptr, &this->ciphertext.len,
+															this->shared_secret.ptr,
+															&this->shared_secret.len))
```
tobiasbrunner, line 276: Indentation.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+		}
+	}
+	EVP_PKEY_free(pkey);
+	EVP_PKEY_CTX_free(ctx);
+	chunk_clear(&seed);
+	return TRUE;
+}
+
+METHOD(key_exchange_t, set_public_key, bool,
+			 private_openssl_kem_t *this, chunk_t value)
```
tobiasbrunner, line 292: Indentation.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+{
+
```
tobiasbrunner, line 294: Remove the empty line.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	if (this->shared_secret.ptr != NULL || this->ciphertext.ptr != NULL)
```
tobiasbrunner, line 295: While our code calls `set_public_key()` only once, this seems like a good call from an API perspective. I might add that to the other KEM implementations as well.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	{
+		return openssl_kem_decapsulate(this, value);
+	}
+
+	/* responder action */
+	return openssl_kem_encapsulate(this, value);
+}
+
+METHOD(key_exchange_t, set_seed, bool,
+			 private_openssl_kem_t *this, chunk_t ignore, drbg_t *seed)
```
tobiasbrunner, line 311: Indentation.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+{
+	drbg_t *drbg = this->thread_drbg->get(this->thread_drbg);
+	if (!seed)
+	{
+		return FALSE;
+	}
+	DESTROY_IF(drbg);
+	this->thread_drbg->set(this->thread_drbg, seed->get_ref(seed));
+	return TRUE;
```
tobiasbrunner, line 320: As mentioned above, here you can just get a reference to the passed `drbg_t`.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+}
+
+METHOD(key_exchange_t, destroy, void,
+private_openssl_kem_t *this)
+{
+	drbg_t *drbg = this->thread_drbg->get(this->thread_drbg);
+	EVP_PKEY_free(this->pkey);
+	chunk_clear(&this->shared_secret);
+	chunk_clear(&this->ciphertext);
```
tobiasbrunner, line 329: `chunk_free()` should be fine for the ciphertext.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+	{
+		return FALSE;
+	}
+	DESTROY_IF(drbg);
+	this->thread_drbg->set(this->thread_drbg, seed->get_ref(seed));
+	return TRUE;
+}
+
+METHOD(key_exchange_t, destroy, void,
+private_openssl_kem_t *this)
```
tobiasbrunner, line 324: Indentation.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+
+/*
+ * Described in header.
+ */
+openssl_kem_t *openssl_kem_create(key_exchange_method_t method)
+{
+	private_openssl_kem_t *this;
+	INIT(this,
+		 .public = {
+			 .ke = {
```
tobiasbrunner, line 343: As commented in the struct definition, `.ke` can be dropped and `.public` would get the methods assigned directly.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,360 @@
...
+				 .get_public_key = _get_public_key,
+				 .set_seed = _set_seed,
+				 .get_method = _get_method,
+				 .destroy = _destroy,
+			 },
+		 },
+		 .group = method,
+		 .pkey = NULL,
+		 .shared_secret = chunk_empty,
+		 .ciphertext = chunk_empty,
```
tobiasbrunner, line 355: Initializing these is not necessary as the `{}` struct assignment ensures that unspecified members are initialized to `0` or `NULL`.

scripts/test.sh
```diff
@@ -226,6 +226,8 @@ openssl*)
 		DEPS=""
 		use_custom_openssl $1
 	elif test "$TEST" = "openssl-awslc"; then
+		CONFIG="${CONFIG} --enable-drbg"
+		TESTS_PLUGINS="${TESTS_PLUGINS} drbg"
```
tobiasbrunner, line 230: We'll need this for OpenSSL too at some point, so you could also add these to the constants directly.

tobiasbrunner (changes requested):
Thanks a lot. Looks pretty good. I've commented some stuff I found (you don't have to worry about the code style issues too much, I can also fix those before applying the change to our branch).

tobiasbrunner:
One thing I'm wondering about is whether we should always use the deterministic versions, just fed by random data either from a `drbg_t` or an `rng_t` (with `RNG_STRONG` - or maybe directly from `RAND_priv_bytes()`). That way we'd reduce the number of code paths and test more of the API that is actually used outside of testing. See the _wolfssl_ plugin's KEM for an example of that approach.

geedo0:
Thanks a lot for taking a look at this, I'm working through the comments now.

> One thing I'm wondering about is whether we should always use the deterministic versions, just fed by random data either from a drbg_t or an rng_t (with RNG_STRONG - or maybe directly from RAND_priv_bytes()).

This is a tricky call to make and may limit your options depending on how libraries decide to interpret NIST's "the interfaces specified in this section should not be made available to applications other than for testing purposes" guidance. At the very least, there's some risk of these APIs not being available/allowed when using a library as an approved FIPS module. At worst, these types of APIs just won't be available at all.

src/libstrongswan/plugins/openssl/openssl_plugin.c
geedo0: It's the latter case. Our consumers tend to use either compile time checks or logic around the version number macro to figure out exactly which features are available. I can add a version check here, but since it's still early days with AWS-LC in this code base, I think we can still get away with avoiding it.

src/libstrongswan/plugins/openssl/openssl_kem.c
geedo0: Thankfully, they all agree that we should align with the project's requirements/preferences. 👍

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,371 @@
...
+		if (!this->drbg->generate(this->drbg, seed.len, seed.ptr))
+		{
+			EVP_PKEY_CTX_free(ctx);
+			return FALSE;
+		}
+		if (!EVP_PKEY_keygen_deterministic(ctx, &this->pkey, seed.ptr,
+										   &seed.len))
+		{
+			EVP_PKEY_CTX_free(ctx);
+			EVP_PKEY_free(this->pkey);
```
tobiasbrunner, line 126: Please remove these remaining `EVP_PKEY_free(this->pkey)` calls. Should the object be allocated, it's freed in `destroy()` anyway, so there is no need to do it here. And as I mentioned before, calling it here as well will cause a double-free because `EVP_PKEY_free()` won't set the pointer to NULL

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,371 @@
...
+			chunk_clear(&seed);
+			return FALSE;
+		}
+	}
+	else if (!EVP_PKEY_keygen(ctx, &this->pkey))
+	{
+		EVP_PKEY_CTX_free(ctx);
+		EVP_PKEY_free(this->pkey);
```
tobiasbrunner, line 134: See above.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,371 @@
...
+											   chunk_t *out)
+{
+	chunk_t pkey_chunk = chunk_empty;
+	size_t public_key_length = 0;
+
+	if (!this->pkey)
+	{
+		return FALSE;
+	}
+	EVP_PKEY_get_raw_public_key(this->pkey, NULL, &public_key_length);
```
tobiasbrunner, line 155: I think I commented on this before. This can't fail? If it could, maybe add the call to the `if` statement above.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,371 @@
...
+			!EVP_PKEY_encapsulate_deterministic(ctx, this->ciphertext.ptr,
+												&this->ciphertext.len,
+												this->shared_secret.ptr,
+												&this->shared_secret.len,
+												seed.ptr, &seed.len))
+		{
+			EVP_PKEY_free(pkey);
+			EVP_PKEY_CTX_free(ctx);
+			chunk_clear(&this->shared_secret);
+			chunk_free(&this->ciphertext);
```
tobiasbrunner, line 280: You don't have to free/clear these explicitly as that's happening in `destroy()` anyway. (However, unlike with `this->pkey` there is no problem with double-frees,  because `chunk_clear/free()` sets the passed pointer to `chunk_empty`.)

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,371 @@
...
+		this->shared_secret = chunk_alloc(shared_secret_length);
+		this->ciphertext = chunk_alloc(ciphertext_length);
+		if (!EVP_PKEY_encapsulate(ctx, this->ciphertext.ptr, &this->ciphertext.len,
+								  this->shared_secret.ptr,
+								  &this->shared_secret.len))
+		{
+			EVP_PKEY_free(pkey);
+			EVP_PKEY_CTX_free(ctx);
+			chunk_clear(&this->shared_secret);
+			chunk_free(&this->ciphertext);
```
tobiasbrunner, line 303: Same as above.

src/libstrongswan/plugins/openssl/openssl_kem.c
```diff
@@ -0,0 +1,371 @@
...
+	free(this);
+}
+
+/*
+ * Described in header.
+ */
+key_exchange_t *openssl_kem_create(key_exchange_method_t method)
+{
+	private_key_exchange_t *this;
+	INIT(this,
```
tobiasbrunner, line 358: Add an empty line before this.

tobiasbrunner (changes requested):
Thanks for the updates. I commented on some remaining issues.

> This is a tricky call to make and may limit your options depending on how libraries decide to interpret NIST's "the interfaces specified in this section should not be made available to applications other than for testing purposes" guidance. At the very least, there's some risk of these APIs not being available/allowed when using a library as an approved FIPS module. At worst, these types of APIs just won't be available at all.

That's an interesting point. At least for wolfSSL and Botan the APIs don't seem dependent on whether the library is built as a FIPS module. And in both plugins I use the respective library's RNG (granted, for the `botan` plugin there is an option to use a different RNG, but that's optional), so NIST's "sampling of random values ... **shall** be performed by the cryptographic module" requirement should also be met.

But it's a good point to keep in mind and I'm fine with your approach in this plugin. Should the deterministic interface not be available, there is the issue that [runtime crypto tests](https://docs.strongswan.org/docs/5.9/plugins/test-vectors.html#_running_tests) won't work, so we might have to extend the `key_exchange_t` interface so implementations can declare whether they support them. And maybe we could extend the testing so that while we can't use the test vectors in that case, we could at least test the general functionality via some roundtrip tests (e.g. do two exchanges and verify that the shared secrets are different). Maybe the latter wouldn't be such a bad thing anyway, especially for plugins like this where the code paths are quite different.

tobiasbrunner:
Thanks again for the updates and your inputs, really appreciated.

I've applied the squashed commit with some minor changes to the commit message to the `six-release` branch. I'll add a reference to this and the other ticket again before I merge the branch to master (that's also when I'll close this ticket). I just want to avoid updates to the tickets whenever the branch is rebased (although, I just noticed that pushing the branch actually updates this ticket anyway because it's based on it - oh, well).

> On the one hand, the generic test guarantees some level of testing/viability/correctness. However, that doesn't stop a crypto module from doing something funny like exchanging keys in plaintext.

I guess the static KAT vectors don't really help in that regard as a malicious module could detect those and behave differently.

geedo0:
Thanks and good luck with the v6 release!

PR #2550, opened 2024-12-02
ha: Derive IKE SA and child SA keys correctly when multiple key exchanges are used

jfhren-stormshield:
Hello,
When adding a second key exchange (or more) to ha/active-passive test, it makes the passive crashes.
I put together a fix to handle additional key exchanges and derive the IKE and CHILD keys accordingly.
I also duplicated the ha/active-passive test as ha/active-passive-multi-ke with an additional key exchange.
Moreover there was a typo in the ha/active-passive test (`ike-id 1` in place of `ike-id 2`).
Thank you.


tobiasbrunner:
Thanks a lot for this PR. I've had a look at this and found some issues, so I reworked this based on an older patch by Thomas Egerer that was never upstreamed.

The first thing I noticed is that the proposal synchronization is incorrect. Additional key exchanges can be negotiated for each of the 7 additional transforms individually and any one of them can be omitted (e.g. `ADDKE1` and `ADDKE4` could be negotiated, while `ADDKE2`, `ADDKE3` and those above 4 might not).

I don't like adding an attribute type for each new transform type. Instead, we can add an attribute that simply transmits an array of `uint16_t` values as a `chunk_t` (each is 0 if not negotiated, the length can even be limited to the highest transform that's actually negotiated). Putting the encoding function in `ha_message.c` also allows using it for both IKE and Child SA proposals.

Another thing is that we don't need a new attribute to identify the SA during the initial exchanges. We can continue to use `HA_IKE_ID` and just first try to checkout a matching SA before creating one. We can also use the proposal as an indicator whether we handle an initial add or a key update (sending the algorithms with every update is redundant anyway as the proposal won't change).

There was also a memory leak due to the reuse of the `dh` variable for the first and the additional key exchange objects, i.e. only the latter would get destroyed. You didn't run into this in the test case because the daemon on `moon` is not properly terminated, so the leak detective won't be able to report it. Note that my changes still reuse the variable (now named `ke`), however, both objects are now destroyed together with the array.

What was missing as well is handling of multiple `HA_IKE_ADD` messages in the HA cache (to resync the SA).

I've also renamed some DH mentions to KE (except for some IKEv1 legacy stuff).

I'm afraid, eventually, there really wasn't anything left from your commit, except for the test scenario, which is appreciated. I hope that's OK.

The changes are in the `2550-ha-multi-ke` branch. Please let me know if you have any feedback.

----

By the way, what's a bit weird in the test scenario, at least in my test runs, is that the first IKE_AUTH fragment is somehow received by `moon`, while neither the IKE_SA_INIT, the IKE_INTERMEDIATE, nor the second IKE_AUTH fragment is. I currently don't understand why that happens.

However, it shows an interesting side-effect of doing the key derivation delayed (i.e. when receiving an encrypted packet instead of when sending back the response to the previous one, which versions before 5.9.7 did, see 44629bbadbd4d7f0c3ca5f8f73237b1b49b1f6f3). Which is that the passive host is unable to process any messages until the active host received the first IKE_AUTH request and synced the SA to complete the last key derivation. Because the IKE_SA is in state `IKE_CONNECTING` on the passive node (and not `IKE_PASSIVE`, which happens later), there is some processing when a packet is received. So in the test scenario, I see the following in the log when the IKE_AUTH message is received:

```
16[NET] received packet: from 192.168.0.100[4500] to 192.168.0.5[4500] (1236 bytes)
16[LIB] MAC verification failed
16[ENC] verifying encrypted fragment payload integrity failed
16[ENC] could not decrypt payloads
16[IKE] integrity check failed
16[IKE] IKE_AUTH request with message ID 2 processing failed
09[CFG] installed HA passive IKE_SA 'rw' 192.168.0.5[mars.strongswan.org]...192.168.0.100[carol@strongswan.org]
```

This means that a handover during that time wouldn't allow completing the SA, which I guess was possible in earlier versions.

jfhren-stormshield:
Thank you for your reply.
Sorry, I was not aware some additional transforms could be omitted. Maybe the test scenario could be tweaked to cover this behavior ?
Concerning the new message attributes, is there not a downward compatibility issue by adding them not a the end ?

jfhren-stormshield:
Destroying incomplete IKE_SAs might be better than letting them timeout eventually.

PR #2552, opened 2024-12-03
child-create: fix double free of labels after migrate

rshearman:
If a migrate of a child-create occurs then labels_i and labels_r are freed, but the pointers are left set. If the task is subsequently destroyed without being reused, then both of these will be double freed.

Fix this by setting labels_i and labels_r to NULL in the migrate method after freeing, similar to other fields that are freed.


tobiasbrunner:
@cla-bot check

Thanks a lot, applied to master.

PR #2580, opened 2024-12-22
nm: Expose remote-ts

SpeedGriffon:
Expose the "remote-ts" config option (04db34a3a748e74dc9199834cadad78e3c307fa3) in the GUI. It is a desired change to me.

### DIFF
--- a/src/frontends/gnome/properties/nm-strongswan-dialog.ui
+++ b/src/frontends/gnome/properties/nm-strongswan-dialog.ui
@@ -737,13 +737,40 @@
             <property name="margin">6</property>
             <property name="spacing">6</property>
             <child>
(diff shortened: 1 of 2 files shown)


tobiasbrunner:
Thanks. In order to not increase the height of the dialog, I wonder if it would make sense to add this field in a new tab (e.g. "Traffic"). We could then even add another field for the local traffic selector (via the change in the `2084-nm-local-ts` branch), which would allow negotiating site-to-site tunnels without virtual IP.

SpeedGriffon:
I had the same thought when I saw #2084. Very well, I've added commits moving the option to a new tab and exposing "local-ts" for when #2084 corrects the error message and gets merged.

tobiasbrunner:
Thanks a lot, merged to master.

Issue #1255, opened 2022-08-26
unable to install inbound and outbound IPsec SA (SAD) in kernel


gv74:
**System (please complete the following information):**
 - OS: Rocky Linux 8.6 (a.k.a. RHEL 8.6)
 - Kernel version (if applicable): 4.18.0-372.19.1.el8_6.x86_64
 - strongSwan version(s): 5.9.6, 5.9.7
 - Tested/confirmed with the latest version: yes

**Describe the bug**
Up to strongswan version 5.9.5 I had no issues. The tunnes are up for months. 
Starting with 5.9.6, strongswan start fine but after about 1 hour it start rekeying in a loop and fail. After a while the tunnel is dropped. I use the same configuration file with 5.9.5, 5.9.6, 5.9.7. During rekeying the following message is issued: "unable to install inbound and outbound IPsec SA (SAD) in kernel" (not sure if is relevant). This also look strange:

```
Fri, 2022-08-26, 17:55:01 15[IKE] <v|1> CHILD_SA rekeying failed, trying again in 8 seconds
Fri, 2022-08-26, 17:55:01 15[CHD] <v|1> CHILD_SA v{1} state change: REKEYING => INSTALLED
Fri, 2022-08-26, 17:55:01 15[CHD] <v|1> CHILD_SA v{96} state change: INSTALLING => DESTROYING
Fri, 2022-08-26, 17:55:01 15[KNL] <v|1> deleting policy m.m.m.125/32 === 192.168.1.0/24 in
Fri, 2022-08-26, 17:55:01 15[KNL] <v|1> policy still used by another CHILD_SA, not removed
Fri, 2022-08-26, 17:55:01 15[KNL] <v|1> not updating policy m.m.m.125/32 === 192.168.1.0/24 in [priority 371327, refcount 1]
Fri, 2022-08-26, 17:55:01 15[KNL] <v|1> deleting policy m.m.m.125/32 === 192.168.1.0/24 fwd
Fri, 2022-08-26, 17:55:01 15[KNL] <v|1> policy still used by another CHILD_SA, not removed
Fri, 2022-08-26, 17:55:01 15[KNL] <v|1> not updating policy m.m.m.125/32 === 192.168.1.0/24 fwd [priority 371327, refcount 1]
Fri, 2022-08-26, 17:55:01 15[KNL] <v|1> deleting SAD entry with SPI c6d14020
[...]

PR #2602, opened 2025-01-08
openssl: Require minimum version 3.0.0 for HKDF

ntransw:
Context: https://github.com/strongswan/strongswan/issues/1255

OpenSSL's HKDF can't process larger MODP groups and nonces with its 1024 byte buffer size in older versions. 3.0.0+ has a sufficient buffer size of 2048 bytes, so require it for OpenSSL's HKDF to be registered over strongSwan's KDF plugin. This will prevent issues during rekeys where a shared secret and nonces are concatenated to form a seed that may potentially be too large.

### DIFF
--- a/src/libstrongswan/plugins/openssl/openssl_plugin.c
+++ b/src/libstrongswan/plugins/openssl/openssl_plugin.c
@@ -496,8 +496,10 @@ METHOD(plugin_t, get_features, int,
 			PLUGIN_PROVIDE(SIGNER, AUTH_HMAC_SHA2_512_256),
 			PLUGIN_PROVIDE(SIGNER, AUTH_HMAC_SHA2_512_512),
 #endif
-#if OPENSSL_VERSION_NUMBER >= 0x10101000L
-		/* HKDF is available since 1.1.0, expand-only mode only since 1.1.1 */
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
@cla-bot check

Thanks and sorry for the delay. Had to fix our CI first (switch to Ubuntu 24.04 on GitHub) and then some other stuff came up.

I found two issues with this. One was that AWS-LC (a fork of BoringSSL by Amazon) supports HKDF for any input (they use a dynamic buffer). The second is automatically enabling the `kdf` plugin when building against older versions of OpenSSL. Because this currently only depends on whether the `openssl` plugin is enabled or not, it won't happen when building against an older version. This was already an issue before, but OpenSSL 1.1.1 was still supported so it wasn't that much of a problem (older versions are pretty rare). However, 1.1.1 is still used on some systems (e.g. Debian Bullseye, Ubuntu 20.04 or the macOS images on GitHub, you can also see such a failure in the third AppVeyor build against OpenSSL 1.1.1 - for the other two builds, which use older OpenSSL versions, I enabled the plugin explicitly before).

I figure it would help if we checked the version and enabled the `kdf` plugin automatically if necessary. So I've added such a check to the configure script (also changed some stuff in the CI script and added a check for AWS-LC) and now pushed an updated commit to master.

Note that the `kdf` plugin is enabled by default, as are other plugins that cause it to get enabled as well, like `cmac` and `xcbc`. But when customizing the configure options, as some of our CI build do, it could be a problem.

PR #2615, opened 2025-01-13
plugin_loader: Add missing ifdef

yangn0:
If HAVE_DLADDR is not defined, all functions using the dl*() interface need to be wrapped by #ifdef

### DIFF
--- a/src/libstrongswan/plugins/plugin_loader.c
+++ b/src/libstrongswan/plugins/plugin_loader.c
@@ -202,6 +202,7 @@ struct plugin_entry_t {
 	linked_list_t *features;
 };
 
+#ifdef HAVE_DLADDR
 /**
  * Destroy a plugin entry
  */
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks, but this patch does not look correct. Only the actual function calls etc. should be wrapped, the rest of the plugin handing is still necessary even when using static plugin constructors.

tobiasbrunner:
I've pushed an alternative patch to the `2615-have-dladdr` branch. Please test it.

yangn0:
> I've pushed an alternative patch to the `2615-have-dladdr` branch. Please test it.

ok. Thanks!

yangn0:
> I've pushed an alternative patch to the `2615-have-dladdr` branch. Please test it.

It works!
I used the following command on freebsd, and it builds successfully:
```
../configure --enable-kernel-pfkey  --disable-updown  --disable-resolve  --enable-kernel-pfroute --disable-kernel-netlink --disable-scripts --disable-shared --enable-static --enable-monolithic --disable-pki  --disable-swanctl --enable-cmd --with-printf-hooks=builtin
make
```

And it also builds successfully on my target os.

tobiasbrunner:
OK, great. Pushed to master.

PR #2635, opened 2025-01-26
rfc/fix: tolerate non-default IPv4 IHLs

lowjoel:
This modifies the eBPF to load the IPv4 Header Length field rather than relying on a fixed size. I used tcpdump BPF output as reference: 

```bash
$ tcpdump -d -i br-lan.xxxx ip and '((udp port 67) or (udp port 68))'
(000) ldh      [12]
(001) jeq      #0x800           jt 2    jf 14
(002) ldb      [23]
(003) jeq      #0x11            jt 4    jf 14
(004) ldh      [20]
(005) jset     #0x1fff          jt 14   jf 6
(006) ldxb     4*([14]&0xf)
(007) ldh      [x + 14]
(008) jeq      #0x43            jt 13   jf 9
(009) jeq      #0x44            jt 13   jf 10
(010) ldh      [x + 16]
(011) jeq      #0x43            jt 13   jf 12
(012) jeq      #0x44            jt 13   jf 14
(013) ret      #262144
(014) ret      #0
```

This loads the IHL into the index register and removes the iphdr length from all other subsequent offsets.

I haven't rewrote this for the non-Linux implementation however.

Discovered as part of openwrt/packages#25801

### DIFF
--- a/src/libcharon/plugins/dhcp/dhcp_socket.c
+++ b/src/libcharon/plugins/dhcp/dhcp_socket.c
@@ -720,24 +720,24 @@ dhcp_socket_t *dhcp_socket_create()
 	int on = 1, rcvbuf = 0;
 
 #if !defined(__APPLE__) && !defined(__FreeBSD__)
-	const size_t skip_ip4 = sizeof(struct iphdr);
-	const size_t skip_udp = skip_ip4 + sizeof(struct udphdr);
+	const size_t skip_udp = sizeof(struct udphdr);
 	struct sock_filter dhcp_filter_code[] = {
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
As mentioned in the OpenWrt ticket, I'm not sure if this is necessary as I don't think many DHCP server will be using IPv4 options. If anything, both implementations should use matching filters.

lowjoel:
Yeah, I could try to match the logic in both but I don't have the means to test it. In any case, if you think this isn't necessary let's skip it then.

PR #2664, opened 2025-02-10
Float to NAT-T port early

michael-dev:
With force_encap, some Microsoft Windows RAS services behave errneous and send the ESP packets and IKE AUTH reply back to IKE SA_INIT UDP encapsulation port used instead of the port used for IKE SA_AUTH later on.

tcpdump output:
```
IP <client-ip>.47547 > <server-ip>.4500: UDP-encap: ESP(spi=0xfd4e5fc2,seq=0x139b), length 116
IP <server-ip>.4500 > <client-ip>.57962: UDP-encap: ESP(spi=0xccc5e213,seq=0x21ff), length 196
```

Avoid this by floating early to the local nat-t port if force_encap is enabled.

### DIFF
--- a/src/libcharon/sa/ikev2/tasks/ike_natd.c
+++ b/src/libcharon/sa/ikev2/tasks/ike_natd.c
@@ -330,7 +330,7 @@ METHOD(task_t, build_i, status_t,
 {
 	notify_payload_t *notify;
 	ike_cfg_t *ike_cfg;
-	host_t *host;
+	host_t *host, *dst;
 
 	if (this->hasher == NULL)
 	{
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Did you receive my emails regarding the CLA? (Just checking via this channel in case something with the email delivery went wrong.)

tobiasbrunner:
Thanks. I've pushed an alternative patch to the `2664-port-float` branch. It rewords the commit message a bit, adds a (lengthy) comment, and handles the changing source port correctly if UDP-encap is not enforced by using the new local endpoint to generate the NAT-D notify (setting the source in the message is not strictly necessary as the latest endpoints will be used when sending it, but it ensures all later tasks etc. can see the updated address on the message should they get it from there). Please let me know if that works for you.

michael-dev:
Thanks. Branch 2664-port-float works fine for me as well.

tobiasbrunner:
Great, thanks for testing.

PR #2708, opened 2025-03-05
wolfssl: Correctly unlock/lock key when running in FIPS mode

julek-wolfssl:
When running in FIPS configuration, `wc_ecc_shared_secret` may return `-287` (`FIPS_PRIVATE_KEY_LOCKED_E`) which indicates that the private key was not unlocked for reading. This patch solves this error and allows StrongSwan to correctly establish a connection.

### DIFF
--- a/src/libstrongswan/plugins/wolfssl/wolfssl_ec_diffie_hellman.c
+++ b/src/libstrongswan/plugins/wolfssl/wolfssl_ec_diffie_hellman.c
@@ -215,6 +215,7 @@ METHOD(key_exchange_t, set_seed, bool,
 static bool compute_shared_key(private_wolfssl_ec_diffie_hellman_t *this)
 {
 	word32 len;
+	int ret;
 #ifdef USE_RNG_FOR_TIMING_RESISTANCE
 	WC_RNG rng;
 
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks. But that's apparently not the only function that may return that error (e.g. the `wc_HKDF*` functions inexplicably do so as well, even though the key is passed right with the rest of the parameters). Since it's not documented which functions do, would you be so kind and please use your insight as a wolfSSL employee and add them around all affected functions our `wolfssl` plugin calls. If the CLA is an issue, you could also provide a list of affected functions and I can prepare a patch.

I also pushed a possible solution (although far broader, [but documented](https://www.wolfssl.com/documentation/manuals/wolfssl-fips-ready/section03.html)) to the `wolfssl-fips` branch (https://github.com/strongswan/strongswan/commit/f63906ab8876f8ef39832e6e8dc4b67f9ea40768), where `PRIVATE_KEY_UNLOCK()` is called once when the plugin is initialized, and `PRIVATE_KEY_LOCK()` once when the plugin is deinitialized. Is that a viable approach?

tobiasbrunner:
Alright. Please be aware that it looks like there might be a bug related to the `wc_EccPublicKeyToDer()` function. According to the documentation I linked above, the `wc_*` functions don't call the locking functions. However, the above function internally calls `SetEccPublicKey`, which in turn calls them. I feel it shouldn't do that if the entry point is via `wc_*` functions (directly or indirectly e.g. via `wc_MakeCert`, which we currently don't use, though).

PR #2725, opened 2025-03-16
agent: Add missing prefixes for ECDSA schemes

ntransw:
Added missing `prefix` assignments when checking ECDSA schemes.


tobiasbrunner:
These prefixes are not necessary. They are only used to check if the signature we get back matches a particular type, which is primarily an issue for RSA with SHA-2 as that's triggered by a flag that older ssh-agents might not support (probably not that much of an issue nowadays), so we'd get back a signature with SHA-1 and `ssh-rsa` prefix. ECDSA signatures always use SHA-2 and if ssh-agent doesn't support them, we simply wouldn't get a signature back. I suppose the same is true for EdDSA, so a prefix isn't really necessary there either.

PR #2765, opened 2025-04-15
Add new counters to the ENV variables in the updown plugin

artraf:
I've added the packet and byte counters to the environment variables in the updown plugin. This feature is quite useful for statistics collection scripts.

### DIFF
--- a/src/libcharon/plugins/updown/updown_listener.c
+++ b/src/libcharon/plugins/updown/updown_listener.c
@@ -19,6 +19,7 @@
 #define _GNU_SOURCE
 #include <stdio.h>
 #include <unistd.h>
+#include <inttypes.h>
 
 #include "updown_listener.h"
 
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks. But that's not a good idea, as it would give users of the plugin/script the wrong impression. The script is not called during rekeyings, so it can only report the numbers of the "last" Child SA. Any traffic that occurred on Child SAs before that won't be seen/reported. Additionally, the script is called for every combination of local and remote traffic selector (not just a single time for the whole Child SA), so the numbers would be reported for every call (and they might not even be the same if the SAs/policies are queried for every call - which also adds some overhead, by the way, as it requires calls into the kernel).

As an alternative, you could use [vici events](https://docs.strongswan.org/docs/latest/plugins/vici.html) (`child-updown` and `child-rekey`) that allow you to keep track of the numbers in an external program/script. Or you use [RADIUS accounting](https://docs.strongswan.org/docs/latest/plugins/eap-radius.html#_accounting), which keeps track of the numbers and reports the complete total over all Child SAs of an IKE SA in the `Accounting-Stop` message. It doesn't require authentication via RADIUS, and with interim updates you'd even get live updates on the usage.

artraf:
Thank you, I appreciate your feedback. I’ve realized that my approach doesn’t provide correct values due to rekeyings, which I hadn’t taken into consideration. I’ll use Vici events instead.

PR #2771, opened 2025-04-23
wolfssl FIPS: unlock the key when necessary

julek-wolfssl:
Wrap the API that requires it in PRIVATE_KEY_UNLOCK/PRIVATE_KEY_LOCK. This can't be done at plugin initialization because it needs to be done for every thread. Strongswan doesn't have on thread creation callbacks for plugins so we need to wrap each direct call.

Tested with this. `test_threading.c` creates new threads so we are testing a cryptographic operation in a new thread. It fails with wolfSSL FIPS without this commit (with FIPS_PRIVATE_KEY_LOCKED_E).
```
diff --git a/src/libstrongswan/tests/suites/test_threading.c b/src/libstrongswan/tests/suites/test_threading.c
index cd0101a51..70fd15faf 100644
--- a/src/libstrongswan/tests/suites/test_threading.c
+++ b/src/libstrongswan/tests/suites/test_threading.c
@@ -546,6 +546,34 @@ static void *rwlock_run(refcount_t *refs)
        }

        rwlock->write_lock(rwlock);
+       {
+           key_exchange_t *a = NULL, *b = NULL;
+           u_char seed[96];
+           chunk_t a_priv, b_priv, a_pub, b_pub, a_sec, b_sec;
+           a_pub = b_pub = a_sec = b_sec = chunk_empty;
+           memset(seed, 12, sizeof(seed));
+           a = lib->crypto->create_ke(lib->crypto, MODP_2048_BIT);
+           ck_assert(a != NULL);
+           b = lib->crypto->create_ke(lib->crypto, MODP_2048_BIT);
+           ck_assert(b != NULL);
+        /* the seed is the concatenation of both DH private keys */
+        a_priv = chunk_create(seed, sizeof(seed)/2);
+        b_priv = chunk_create(seed + sizeof(seed)/2, sizeof(seed)/2);
+        ck_assert(a->set_seed(a, a_priv, NULL));
+        ck_assert(b->set_seed(b, b_priv, NULL));
+        ck_assert(a->get_public_key(a, &a_pub));
+        ck_assert(b->set_public_key(b, a_pub));
+        ck_assert(b->get_shared_secret(b, &b_sec));
+        ck_assert(b->get_public_key(b, &b_pub));
+        ck_assert(a->set_public_key(a, b_pub));
+        ck_assert(a->get_shared_secret(a, &a_sec));
+        DESTROY_IF(a);
+        DESTROY_IF(b);
+        chunk_free(&a_pub);
+        chunk_free(&b_pub);
+        chunk_free(&a_sec);
+        chunk_free(&b_sec);
+       }
        ck_assert_int_eq(*refs, 0);
        sched_yield();
        rwlock->unlock(rwlock);
```

Also contains a small fix for correct ecc initialization. This passes all tests when configured with `--enable-pki --enable-wolfssl --enable-pem --enable-pkcs1 --enable-pkcs8 --enable-x509 --enable-constraints --enable-test-vectors --disable-openssl`.

### DIFF
--- a/src/libstrongswan/plugins/wolfssl/wolfssl_diffie_hellman.c
+++ b/src/libstrongswan/plugins/wolfssl/wolfssl_diffie_hellman.c
@@ -93,9 +93,14 @@ METHOD(key_exchange_t, get_shared_secret, bool,
 
 	if (!this->shared_secret.len)
 	{
+        int ret;
 		this->shared_secret = chunk_alloc(this->len);
-		if (wc_DhAgree(&this->dh, this->shared_secret.ptr, &len, this->priv.ptr,
-					   this->priv.len, this->other.ptr, this->other.len) != 0)
+		PRIVATE_KEY_UNLOCK();
+		ret = wc_DhAgree(&this->dh, this->shared_secret.ptr, &len,
+		        this->priv.ptr, this->priv.len, this->other.ptr,
+		        this->other.len);
+		PRIVATE_KEY_LOCK();
+		if (ret != 0)
 		{
 			DBG1(DBG_LIB, "DH shared secret computation failed");
 			chunk_free(&this->shared_secret);
@@ -139,9 +144,13 @@ METHOD(key_exchange_t, set_seed, bool,
 	/* calculate public value - g^priv mod p */
 	if (wolfssl_mp2chunk(&this->dh.g, &g))
 	{
+	    int ret;
 		len = this->pub.len;
-		if (wc_DhAgree(&this->dh, this->pub.ptr, &len, this->priv.ptr,
-						 this->priv.len, g.ptr, g.len) == 0)
+		PRIVATE_KEY_UNLOCK();
+		ret = wc_DhAgree(&this->dh, this->pub.ptr, &len, this->priv.ptr,
+						 this->priv.len, g.ptr, g.len);
+		PRIVATE_KEY_LOCK();
+		if (ret == 0)
 		{
 			this->pub.len = len;
 			success = TRUE;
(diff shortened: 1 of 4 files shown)


tobiasbrunner:
Thanks for this PR.

> The list of API that need the FIPS unlocking is:

What if the list changes in the future? Is this documented somewhere?

We still need a signed CLA. Or would you prefer I implement this based on the list of functions you provided?

julek-wolfssl:
> What if the list changes in the future?

The FIPS module does not change very often as its tied to the exact version what is certified by the lab.

> Is this documented somewhere?

I'm not sure. I'll check with the team.

> We still need a signed CLA.

I'll write to info@strongswan.org shortly.

tobiasbrunner:
@cla-bot check

> > What if the list changes in the future?
> 
> The FIPS module does not change very often as its tied to the exact version what is certified by the lab.

I see. But that will still change in the future (maybe even more significantly as there will be several wolfSSL versions between releases of the FIPS module).

> > Is this documented somewhere?
> 
> I'm not sure. I'll check with the team.

It would be helpful for future releases of the FIPS module (e.g. [here](https://www.wolfssl.com/documentation/manuals/wolfssl-fips-ready/section03.html), or even in the repository somewhere would be fine).

I've pushed the two commits to the `wolfssl-fips` branch, with some code style changes (see [here](https://docs.strongswan.org/docs/latest/devs/programmingStyle.html) for details) and edits of the commit messages. Let me know if that's OK for you. If you have time, please also have a look at the other commits in that branch, they were the result @volga629-1's experiments with wolfSSL's FIPS module in #2603.

tobiasbrunner:
> We have updated our documentation at https://www.wolfssl.com/documentation/manuals/wolfssl-fips-faq/section02.html#key-access-management to include a list of API that require unlocking.

Great, thanks for doing that.

> I looked at the `wolfssl-fips` branch. Everything looks good.

Appreciated, thanks.

I've now merged the changes to master.

PR #2786, opened 2025-05-12
openssl: Fix AWS-LC build

geedo0:
The `crypt` function defined in `openssl_crypter.c` conflicts with the `crypt` function defined in `unistd.h` and triggers a compilation error when building against the latest version of AWS-LC. This [PR](https://github.com/aws/aws-lc/pull/2321) introduced the new transitive of `unistd.h` include to `bio.h`.

This PR simply renames the function to avoid the error. There are identically named `crypt` functions in similar source files like `openssl_aead`. I did not rename things exhaustively, if that's preferred I can do that just let me know.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -132,7 +132,7 @@ build_openssl()
 
 build_awslc()
 {
-	LC_REV=1.48.5
+	LC_REV=1.51.2
 	LC_PKG=aws-lc-$LC_REV
 	LC_DIR=$DEPS_BUILD_DIR/$LC_PKG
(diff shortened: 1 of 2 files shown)


tobiasbrunner:
Thanks for being proactive :) I would have ran into this with the next AWS-LC version bump for our CI (I've update to 1.51.2 in the `devel` branch yesterday, now in `master`).

> There are identically named `crypt` functions in similar source files like `openssl_aead`.

It's definitely an issue there as well. I've pushed an updated commit to `master`.

PR #2825, opened 2025-07-09
Latest gcc fixes

erchaland:
Those fixes have been detected by gcc error/warning policies on latest versions. However, they correspond to real potential issues.

### DIFF
--- a/src/libcharon/encoding/payloads/encrypted_payload.c
+++ b/src/libcharon/encoding/payloads/encrypted_payload.c
@@ -1023,7 +1023,7 @@ encrypted_fragment_payload_t *encrypted_fragment_payload_create()
 				.get_length = _frag_get_length,
 				.add_payload = _frag_add_payload,
 				.remove_payload = (void*)return_null,
-				.generate_payloads = (void*)nop,
+				.generate_payloads = (void (*)(encrypted_payload_t *, generator_t *)) nop_with_args,
 				.set_transform = _frag_set_transform,
 				.get_transform = _frag_get_transform,
 				.encrypt = _frag_encrypt,
--- a/src/libstrongswan/utils/identification.c
+++ b/src/libstrongswan/utils/identification.c
@@ -1765,7 +1765,7 @@ static private_identification_t *identification_create(id_type_t type)
 			this->public.hash = _hash_binary;
 			this->public.equals = _equals_binary;
 			this->public.matches = _matches_any;
-			this->public.contains_wildcards = (void*)return_true;
+			this->public.contains_wildcards = (bool (*) (identification_t *)) return_true_with_arg;
 			break;
 		case ID_FQDN:
 		case ID_RFC822_ADDR:
@@ -1800,13 +1800,13 @@ static private_identification_t *identification_create(id_type_t type)
 			this->public.hash = _hash_binary;
 			this->public.equals = _equals_binary;
 			this->public.matches = _matches_range;
-			this->public.contains_wildcards = (void*)return_false;
+			this->public.contains_wildcards = (bool (*) (identification_t *)) return_false_with_arg;
 			break;
 		default:
 			this->public.hash = _hash_binary;
 			this->public.equals = _equals_binary;
 			this->public.matches = _matches_binary;
-			this->public.contains_wildcards = (void*)return_false;
+			this->public.contains_wildcards = (bool (*) (identification_t *)) return_false_with_arg;
 			break;
 	}
 	return this;
--- a/src/libstrongswan/utils/utils.c
+++ b/src/libstrongswan/utils/utils.c
@@ -278,6 +278,14 @@ bool return_true()
 }
 
 /**
+ * returns TRUE
+ */
+bool return_true_with_arg(void *arg)
+{
+	return TRUE;
+}
+
+/**
  * returns FALSE
  */
 bool return_false()
@@ -286,9 +294,24 @@ bool return_false()
 }
 
 /**
+ * returns FALSE
+ */
+bool return_false_with_arg(void *arg)
+{
+	return FALSE;
+}
+
+/**
  * nop operation
  */
 void nop()
+{
+}
+
+/**
+ * nop operation
+ */
+void nop_with_args(void *arg1, void *arg2)
 {
 }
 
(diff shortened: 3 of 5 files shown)


tobiasbrunner:
Regarding the casts in the second commit, see 11978ddd39e800b5f35f721d726e8a4cb7e4ec0f.

Could you clarify what the misalignment issue was that GCC reported for `memwipe_inline()`? You say say GCC 14.2.0, however, I don't get that version to complain (when manually disabling `explicit_bzero` on Linux and using `-Werror -Wall -Wextra`). What compile options did you use? On what platform/architecture did you try this?

erchaland:
> Regarding the casts in the second commit, see [11978dd](https://github.com/strongswan/strongswan/commit/11978ddd39e800b5f35f721d726e8a4cb7e4ec0f).

I did the fix on 6.0.1. This commit was not in. I tested it and strongswan compiles without the second commit.

> Could you clarify what the misalignment issue was that GCC reported for `memwipe_inline()`? You say say GCC 14.2.0, however, I don't get that version to complain (when manually disabling `explicit_bzero` on Linux and using `-Werror -Wall -Wextra`). What compile options did you use? On what platform/architecture did you try this?

I use gcc from mingw64 x86_64. I didn't use special CFLAGS. This was on 6.0. Perhaps a fix has been done elsewhere ?

tobiasbrunner:
> I did the fix on 6.0.1. This commit was not in. I tested it and strongswan compiles without the second commit.

:+1: 

> I use gcc from mingw64 x86_64. I didn't use special CFLAGS. This was on 6.0. Perhaps a fix has been done elsewhere ?

Hm, not sure. But our AppVeyor Windows builds use gcc-mingw-w64-x86-64 from MSYS2 and that works fine  ([latest master](https://ci.appveyor.com/project/tobiasbrunner/strongswan-52lo9/builds/52369413)). These builds use the same CFLAGS I mentioned above, which are the default when building from the repository or explicitly used when adding `--enable-warnings`. Depending on the image, the GCC version is either 9.1.0 or 14.2.0. What exactly was the warning/error by GCC you saw?

erchaland:
Hi Tobias

> > I use gcc from mingw64 x86_64. I didn't use special CFLAGS. This was on 6.0. Perhaps a fix has been done elsewhere ?
> 
> Hm, not sure. But our AppVeyor Windows builds use gcc-mingw-w64-x86-64 from MSYS2 and that works fine ([latest master](https://ci.appveyor.com/project/tobiasbrunner/strongswan-52lo9/builds/52369413)). These builds use the same CFLAGS I mentioned above, which are the default when building from the repository or explicitly used when adding `--enable-warnings`. Depending on the image, the GCC version is either 9.1.0 or 14.2.0. What exactly was the warning/error by GCC you saw?

I had to downgrade back my gcc to https://repo.msys2.org/mingw/mingw64/mingw-w64-x86_64-gcc-14.2.0-3-any.pkg.tar.zst manually. Even on master, I have the error in the attached file.
[gcc-14-ss-master.txt](https://github.com/user-attachments/files/21160618/gcc-14-ss-master.txt)
I don't get the error with gcc 15.

tobiasbrunner:
Hm, that looks like a compiler bug to me. It complains that `i` in the assignment here

https://github.com/strongswan/strongswan/blob/a8c2d125f1520acbd05d9de0773f7a1c92e2aeef/src/libstrongswan/utils/utils/memory.h#L135-L138

exceeds the array bounds if it's 8 and the buffer was allocated with `malloc(8)`. That's technically correct, but it can't happen because the assignment is in that `for` loop that's bound by `i < n` and `n` is 8.

Your patch does not seem to address this particular issue either. It fixes that the `n >= sizeof(long)` check is not correct if the memory area is not aligned and its size is just `sizeof(long)` or a bit larger. I'd guess that malloc'ed chunks are generally aligned, but it could possibly be a problem when wiping parts of a buffer manually (which we rarely do, though).

So that patch makes sense. One optimization would be to write `i <= n - sizeof(long)` as that avoids a potential overflow of `i` and might be a bit more efficient. Also, we don't add spaces after casts. And we also need a signed CLA (see the first comment above).

erchaland:
> Hm, that looks like a compiler bug to me. It complains that `i` in the assignment here
> 
> https://github.com/strongswan/strongswan/blob/a8c2d125f1520acbd05d9de0773f7a1c92e2aeef/src/libstrongswan/utils/utils/memory.h#L135-L138
> 
> exceeds the array bounds if it's 8 and the buffer was allocated with `malloc(8)`. That's technically correct, but it can't happen because the assignment is in that `for` loop that's bound by `i < n` and `n` is 8.
> 
> Your patch does not seem to address this particular issue either. It fixes that the `n >= sizeof(long)` check is not correct if the memory area is not aligned and its size is just `sizeof(long)` or a bit larger. I'd guess that malloc'ed chunks are generally aligned, but it could possibly be a problem when wiping parts of a buffer manually (which we rarely do, though).

I fully agree. I just tried to simplify the code to help gcc not to mistake.

> So that patch makes sense. One optimization would be to write `i <= n - sizeof(long)` as that avoids a potential overflow of `i` and might be a bit more efficient. Also, we don't add spaces after casts. And we also need a signed CLA (see the first comment above).

The optimization confuses gcc. It produces the same error. Substraction on unsigned seems to be a wrong idea. I fixed the space before casts. Do you want I remove the second commit from branch ?

Regarding the CLA, my manager CHETIBI Amine tried to contact Mr. Andreas Steffen and you by email. He is still awaiting your response.

tobiasbrunner:
> The optimization confuses gcc. It produces the same error.

Since it's a compiler bug, definitely possible.

> Substraction on unsigned seems to be a wrong idea.

No, but it could cause a similar issue (underflow) if the `if` statement is removed and `n < sizeof(long)`.

But actually, I was wrong with my assessment before. The current code, while a bit weird, is correct even with small, unaligned buffers (`m` properly acts as an upper bound). So I don't think any changes are required (we'll just have to live with the bug in that particular GCC version).

> Do you want I remove the second commit from branch ?

That's not necessary either, yes.

> Regarding the CLA, my manager CHETIBI Amine tried to contact Mr. Andreas Steffen and you by email. He is still awaiting your response.

Hm, I don't remember any emails regarding CLAs from him. There is an entity CLA (in addition or instead of individual CLAs), which could be an option should there be more PRs.

PR #2827, opened 2025-07-10
whitelist: use a watcher for control socket reading rather than blocking

rshearman:
Performing a stream read_all call (which is a blocking read) from within the accept callback has the issue that if a whitelist client is still connected whilst a shutdown of the charon deamon is triggered then that shutdown won't complete gracefully due to the accept task never exiting.

So fix shutting down gracefully by using a socket watcher rather than a blocking read upon connection accept. Fall back to a blocking read for partial messages to avoid the complexity associated (i.e. storing state) for incomplete reads, which shouldn't block and cause the original problem if the client only sends whole messages.

### DIFF
--- a/src/libcharon/plugins/whitelist/whitelist_control.c
+++ b/src/libcharon/plugins/whitelist/whitelist_control.c
@@ -95,13 +95,32 @@ static void list(private_whitelist_control_t *this,
 /**
  * Dispatch a received message
  */
-static bool on_accept(private_whitelist_control_t *this, stream_t *stream)
+CALLBACK(on_read, bool,
+	private_whitelist_control_t *this, stream_t *stream)
 {
 	identification_t *id;
 	whitelist_msg_t msg;
+	ssize_t n;
 
-	while (stream->read_all(stream, &msg, sizeof(msg)))
+	while (TRUE)
 	{
+		n = stream->read(stream, &msg, sizeof(msg), FALSE);
+		if (n <= 0)
+		{
+			if (errno == EWOULDBLOCK)
+			{
+				break;
+			}
+			return FALSE;
+		}
+		if (n < sizeof(msg))
+		{
+			if (!stream->read_all(stream, ((char *)&msg) + n, sizeof(msg) - n))
+			{
+				return FALSE;
+			}
+		}
+
 		msg.id[sizeof(msg.id) - 1] = 0;
 		id = identification_create_from_string(msg.id);
 		switch (ntohl(msg.type))
@@ -131,7 +150,14 @@ static bool on_accept(private_whitelist_control_t *this, stream_t *stream)
 		id->destroy(id);
 	}
 
-	return FALSE;
+	return TRUE;
+}
+
+CALLBACK(on_accept, bool,
+	private_whitelist_control_t *this, stream_t *stream)
+{
+	stream->on_read(stream, on_read, this);
+	return TRUE;
 }
 
 METHOD(whitelist_control_t, destroy, void,


tobiasbrunner:
Thanks.

> Performing a stream read_all call (which is a blocking read) from within the accept callback has the issue that if a whitelist client is still connected whilst a shutdown of the charon deamon is triggered then that shutdown won't complete gracefully due to the accept task never exiting.

Could you clarify a bit what exactly happened with that client (why the read didn't get completed), how it's related to the shutdown, and why the same couldn't happen with this patch?

rshearman:
> Thanks.
> 
> > Performing a stream read_all call (which is a blocking read) from within the accept callback has the issue that if a whitelist client is still connected whilst a shutdown of the charon deamon is triggered then that shutdown won't complete gracefully due to the accept task never exiting.
> 
> Could you clarify a bit what exactly happened with that client (why the read didn't get completed), how it's related to the shutdown, and why the same couldn't happen with this patch?

We have our own application, which is a daemon, that acts as a whitelist client over a Unix domain socket to strongswan, and there's no guarantee that during system shutdown our application will get signalled to terminate first and thus close its connection. Thus without this change, whilst the stream is active (not closed) the read loop in the whitelist plugin in strongswan won't be exited. Our application also uses vici sockets in the same way works and that works without issue.

rshearman:
The same cannot happen with this patch, because whilst src/libstrongswan/networking/streams/stream.c:_read_all restarts its loop on EINTR, whereas src/libstrongswan/processing/watcher.c:watch calls thread_cancellation_point() if the thread was cancelled whilst the blocking poll() system call was in progress (i.e. poll() returns EINTR), allowing the thread to be gracefully exited.

tobiasbrunner:
> We have our own application, which is a daemon, that acts as a whitelist client over a Unix domain socket to strongswan

Oh, I see. Your daemon just stays connected and occasionally sends a message over the same connection (i.e. it doesn't create a new connection for each message like the `whitelist` command does)? Then this makes sense.

As it's not a critical fix, I'll merge this after the 6.0.2 release next week.

rshearman:
> > We have our own application, which is a daemon, that acts as a whitelist client over a Unix domain socket to strongswan
> 
> Oh, I see. Your daemon just stays connected and occasionally sends a message over the same connection (i.e. it doesn't create a new connection for each message like the `whitelist` command does)? Then this makes sense.

Correct.

> As it's not a critical fix, I'll merge this after the 6.0.2 release next week.

👍

tobiasbrunner:
Looking at this a bit closer, I wonder if we couldn't simplify it some. As `on_read()` is only called if there is data to read without blocking (or an error occurred on the socket), instead of looping in `on_read()`, we could just return `TRUE` after reading one message with `read_all()` (would avoid the first `read()` call and its error handling). Looks cleaner but depending on how many messages the client sends in a row it could be slightly less efficient (probably not noticeable). There also was a memory leak once a client disconnects and `FALSE` is returned from `on_read()`. Please see the `2827-whitelist-watcher` branch.

rshearman:
Our use case involves sending up to 10000 whitelist messages in one go, where there's a chance the additional overhead might be noticeable. However, I don't have any data that supports this, so I'm happy with your changes and agree that they simplify the implementation a lot. Thank you for also finding and fixing the memory leak.

I've tested your changes in the `2827-whitelist-watcher` branch and can confirm that, functionally, they look good. Appreciate you taking the time to review and fix/build upon my proposed changes.

tobiasbrunner:
Thanks for your feedback. 



> Our use case involves sending up to 10000 whitelist messages in one go, where there's a chance the additional overhead might be noticeable.

Yeah, it definitely makes a difference for such workloads. A quick test with `whitelist add-from` and a list of 10000 random email addresses shows a slowdown by a factor of 10 (up to 1 second). Not sure if that's a problem in practice, but doesn't look ideal in a direct comparison. Because I don't like the multiple error and cleanup paths, I wondered how it would look like if we added state and just used `read()`. And I think it's not too bad, actually. I've pushed that to the branch. Let me know what you think.

rshearman:
Those new changes are ideal - properly handling the partial reads turns out to be much easier than I feared. I've tested them and can confirm that functionally they look good. Thank you once again.

tobiasbrunner:
Great, thanks for testing again. I've merged the changes to master.

PR #2829, opened 2025-07-10
openssl: Init salt for KDF check too

hendrikdonner:
Fixes

charon: 11[IKE] KDF_PRF with PRF_HMAC_SHA2_256 not supported
charon: 11[IKE] key derivation failed

during IKE handshake and

charon: 00[LIB] enabled  KDF_PRF[openssl]: unable to apply any available test vectors

during running the test vectors from the test vector plugin.


hendrikdonner:
Looks like this was already fixed today, missed the discussion in https://github.com/strongswan/strongswan/discussions/2828. I debugged the same issue around the same time.

PR #2842, opened 2025-07-17
whitelist: doubling the id length

seantywork:
Hello!

I've recently run into a situation where it's very likely that the current 128 bytes limit on whitelist id field might not be sufficient.

As an internal logic dictates that I have to deal with automatically generated certs with its subject attributes being something similar to:
C=KR,ST=Seoul,O=Some Alien Institution,OU=Area 51,CN=some user identification hex string
where CN hexstring could extend to maximum 64 bytes, which makes it 128 characters for CN value alone.

Would it be okay to double it?

### DIFF
--- a/src/libcharon/plugins/whitelist/whitelist_msg.h
+++ b/src/libcharon/plugins/whitelist/whitelist_msg.h
@@ -53,7 +53,7 @@ struct whitelist_msg_t {
 	/** message type */
 	int type;
 	/** null terminated identity */
-	char id[128];
+	char id[256];
 } __attribute__((packed));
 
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
@cla-bot check

Thanks. I'm fine with the change (we already use the same limit in the `error-notify` and `lookip` plugins). However, the protocol is simple and has no version number. So it could be a bit tricky changing it. But I'd assume people will generally use the header file from the current version when they compile their own clients. If they wrote a client in a different language, that could be an issue. But I'll definitely mention this change in the developer notes for the next release.

PR #2844, opened 2025-07-18
ml: disable versioning for shared object

rshearman:
Avoid generating versioned shared objects which would need to be installed along with the version-independent symlink by specifying "-avoid-version" in the libtool LDFLAGS for the plugin. Avoid any unwanted surprises by also specifying the "-module" option, making the LDFLAGS consistent with all other libstrongswan plugins.


tobiasbrunner:
Thanks, applied to master. I wonder how this got lost :thinking:

PR #2869, opened 2025-08-19
list-conns: Include local and remote ports

mathiasaerts:
In order to get the **configured** local and remote ports for a connection, which may differ from the actually used ports in the SA, they need to be exposed in the `list-conns` data.

This PR adds these items from the connection in the `vici` plugin and `swanctl` output.

The output of `swanctl --list-conns` will include them in square brackets with the local and remote addresses:
```
  local:  10.133.2.1[4500]
  remote: 10.133.6.2[4500]
```

### DIFF
--- a/src/libcharon/plugins/vici/README.md
+++ b/src/libcharon/plugins/vici/README.md
@@ -882,6 +882,8 @@ _list-conns_ command.
 			remote_addrs = [
 				<list of valid remote IKE endpoint addresses>
 			]
+			local_port = <local IKE endpoint port>
+			remote_port = <remote IKE endpoint port>
 			version = <IKE version as string, IKEv1|IKEv2 or 0 for any>
 			reauth_time = <IKE_SA reauthentication interval in seconds>
 			rekey_time = <IKE_SA rekeying interval in seconds>
--- a/src/libcharon/plugins/vici/vici_query.c
+++ b/src/libcharon/plugins/vici/vici_query.c
@@ -956,6 +956,11 @@ CALLBACK(list_conns, vici_message_t*,
 		tokens->destroy(tokens);
 		b->end_list(b);
 
+		b->add_kv(b, "local_port", "%u",
+			ike_cfg->get_my_port(ike_cfg));
+		b->add_kv(b, "remote_port", "%u",
+			ike_cfg->get_other_port(ike_cfg));
+
 		b->add_kv(b, "version", "%N", ike_version_names,
 			peer_cfg->get_ike_version(peer_cfg));
 		b->add_kv(b, "reauth_time", "%u",
--- a/src/swanctl/commands/list_conns.c
+++ b/src/swanctl/commands/list_conns.c
@@ -223,17 +223,17 @@ CALLBACK(conn_sn, int,
 }
 
 CALLBACK(conn_list, int,
-	hashtable_t *sa, vici_res_t *res, char *name, void *value, int len)
+	hashtable_t *ike, vici_res_t *res, char *name, void *value, int len)
 {
 	if (chunk_printable(chunk_create(value, len), NULL, ' '))
 	{
 		if (streq(name, "local_addrs"))
 		{
-			printf("  local:  %.*s\n", len, value);
+			printf("  local:  %.*s[%s]\n", len, value, ike->get(ike, "local_port"));
 		}
 		if (streq(name, "remote_addrs"))
 		{
-			printf("  remote: %.*s\n", len, value);
+			printf("  remote: %.*s[%s]\n", len, value, ike->get(ike, "remote_port"));
 		}
 	}
 	return 0;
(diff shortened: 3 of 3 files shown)


tobiasbrunner (changes requested):
@cla-bot check

> In order to get the **configured** local and remote ports for a connection, which may differ from the actually used ports in the SA, they need to be exposed in the `list-conns` data.

Could you please clarify what you mean by that.

I've added some comments. And please also document the two new properties after this: https://github.com/strongswan/strongswan/blob/2560146204f81da1a3f2c6dd2aa38f681ac05758/src/libcharon/plugins/vici/README.md?plain=1#L882-L884

mathiasaerts:
> > In order to get the **configured** local and remote ports for a connection, which may differ from the actually used ports in the SA, they need to be exposed in the `list-conns` data.
> 
> Could you please clarify what you mean by that.

Definitely. We are using `encap = yes` and in some networks, we've seen NAT rewriting of the source port by an in-between router/firewall, causing the `remote_port` (source port) in the SA to differ from the configured remote port. For example:

Host A:
```
root@host-A:~# ip -d xfrm state | grep -A 3 0xc756462a | grep sport
	proto esp spi 0xc756462a reqid 2 mode transport
	encap type espinudp sport 4500 dport 4500 addr 0.0.0.0
```

Host B:
```
root@host-B:~# ip -d xfrm state | grep -A 3 0xc756462a | grep sport
	proto esp spi 0xc756462a reqid 1 mode transport
	encap type espinudp sport 62995 dport 4500 addr 0.0.0.0
```

Which is also reflected in the `list-sa` command on host B:
```
root@host-B:~# swanctl --list-sa
298d7e19-1639-XXXXXXXX: #358, ESTABLISHED, IKEv2, 26c27d32c8c1eeff_i* 8fca4cd6b66405b0_r
  local  'ef052683-9544-XXXXXXXX' @ 141.xxx.xxx.xxx[4500]
  remote '298d7e19-1639-XXXXXXXX' @ 85.xxx.xxx.xxx[62995]
  ...
```

For our specific setup, we are also interested in the port that is set in the config.

> I've added some comments. And please also document the two new properties after this:
> 
> https://github.com/strongswan/strongswan/blob/2560146204f81da1a3f2c6dd2aa38f681ac05758/src/libcharon/plugins/vici/README.md?plain=1#L882-L884

Thanks, that simplifies the change indeed. At some point in our testing they appeared to be empty, but I've updated the PR and tested this build, which is working as expected. Also updated the README for the vici plugin as requested.

tobiasbrunner:
> We are using `encap = yes` and in some networks, we've seen NAT rewriting of the source port by an in-between router/firewall, causing the `remote_port` (source port) in the SA to differ from the configured remote port.

Hm, that has nothing to do with `encap = yes`. It's absolutely normal behavior for IKE. The source port can be completely arbitrary and even move around. I mean the Android app and the NetworkManager plugin even use ephemeral source ports (i.e. even without any NAT the source port will not be 500 or 4500). And a switch to 4500 is also pretty much always  the case (due to NAT or because of MOBIKE), so it's unlikely that you see the default remote port of 500 (as reported by `list-conns` with these changes) in any established SA. Might be a bit confusing for users, but I guess it could still be useful to list the ports in `list-conns` if anything particular is configured (e.g. when using `socket-dynamic`). But just be aware that the configured ports are never enforced. The ports are considered when matching configs as responder, but they can always diverge from the config due to the protocol's agility.

> At some point in our testing they appeared to be empty

Do you have any details on this? Because I checked all uses of `ike_cfg_create_t` and both `local_port` and `remote_port` are always set. However, the local port can technically be 0 as it's usually determined via socket (if not configured), and e.g. the `socket-dynamic` plugin always returns 0 (it expects that the port is configured, though, so the value should not be 0 in `list-conns`).

> but I've updated the PR and tested this build, which is working as expected

Can you please squash the commits together. Thanks!

mathiasaerts:
> Hm, that has nothing to do with `encap = yes`

I thought I'd mention it as this would always lead to port 4500 being chosen over 500 right? In any case I understand this behaviour is not specific to having `encap = yes` in the configuration.

> Might be a bit confusing for users

I agree. We actually only need the configured ports to be present in the `list-conns` vici message, but figured it would be best to also include them in the `swanctl --list-conns` output.

> Do you have any details on this?

This is probably a non-issue, at some point I think we were using mixed builds of `strongswan` (or `charon-systemd`) and `swanctl` binaries, where strongswan was not running with the updated vici plugin, so the values were never set in the `list-conns` message. `vici_find_str` would then return the default value of `0`, hence the additional validation we had before.

> Can you please squash the commits together. Thanks!

Done!

tobiasbrunner:
> I thought I'd mention it as this would always lead to port 4500 being chosen over 500 right?

Yes, it fakes a NAT situation to force UDP encapsulation.

> We actually only need the configured ports to be present in the `list-conns` vici message, but figured it would be best to also include them in the `swanctl --list-conns` output.

What exactly is your use case?

> This is probably a non-issue, at some point I think we were using mixed builds of `strongswan` (or `charon-systemd`) and `swanctl` binaries, where strongswan was not running with the updated vici plugin, so the values were never set in the `list-conns` message. `vici_find_str` would then return the default value of `0`, hence the additional validation we had before.

Ah, I see.

mathiasaerts:
> What exactly is your use case?

Without going into too much detail, we are dynamically building tunnel configurations and need to detect if source port remapping is occurring in the remote network to alter the tunnel configuration accordingly. Comparing the `remote_port` from the SA and the configured port in the connection provides an easy way to achieve this. Since the `remote_port` was not included in `list-conns`, the alternative would have been to parse the config file ourselves, which would be less reliable.

tobiasbrunner:
OK, thanks. So the idea is to later use the updated config to initiate a connection to the mapped port?

mathiasaerts:
> So the idea is to later use the updated config to initiate a connection to the mapped port?

Not exactly. The Strongswan connection will work either way (as long as the remote side initiates it) and doesn't necessarily need the changed port in the config, although that is an interesting thought as well :smile: 

By "tunnel configurations" I did not mean swanctl configs (although these are being generated automatically as well), but rather configuration of overlay tunneling protocols, which we would change based on the configured ports, not the mapped ports.

Thanks for merging these changes!

tobiasbrunner:
> By "tunnel configurations" I did not mean swanctl configs (although these are being generated automatically as well), but rather configuration of overlay tunneling protocols, which we would change based on the configured ports, not the mapped ports.

Ah. And these configs need to know the UDP port used for encapsulated ESP? Shouldn't that be completely transparent to the overlay (or that there even is UDP encap, or that ESP is used at all)? Or is it actually ESP (in UDP) that's running on top of an overlay network?

mathiasaerts:
> Shouldn't that be completely transparent to the overlay (or that there even is UDP encap, or that ESP is used at all)?

It should, but having multiple remote endpoints behind the same NAT router is causing issues matching the overlay tunnel with the traffic selector and firewall state. Assigning unique ports helps to resolve this, both for the overlay as for the NAT-T port. So in our case, the overlay configuration needs to know when the swanctl config is non-default.

PR #2882, opened 2025-09-08
Prevent DNS resolution for mismatched %any4/%any6 macros (fixes #2880)

### LINKED ISSUE #2880: IKE config matching: %any6 triggers unwanted DNS resolution for IPv4 candidates
**System**:
OS: Ubuntu 22.04
strongSwan version(s): 5.9.13
Tested/confirmed with the latest version: yes

**Description**
When an incoming IKE_SA_INIT request is received from an IPv4 peer, and one of the left/right addresses in the added configurations is set to %any6, strongSwan attempts to resolve the literal string %any6 via DNS for the IPv4 family. This results in unnecessary DNS resolution attempts for configuration macros that are not hostnames.

**Steps to reproduce the behavior:**

- Configure multiple connections with mixed IPv4 and IPv6 addresses. For example, one connection with left address=%any6 and another with left address=1.2.3.4.
- Initiate an IKE_SA_INIT from an IPv4 peer.
- Observe that during ike_cfg matching, host_create_from_dns("%any6", AF_INET, ...) is called.
- Since host_create_from_string_and_family() does not handle %any6 for AF_INET, strongSwan falls back to lib->hosts->resolve(), which attempts to DNS-resolve %any6 as if it were a hostname.

**Expected behavior**
Configuration macros like %any, %any4, %any6 should be recognized and skipped without attempting DNS resolution. DNS lookups should only be performed for actual hostnames.
At present %any is skipped unconditionally. But %any4, %any6 are skipped only when the incoming IKE_SA_INIT is of same type

**Logs/Backtraces**
Sep  6 23:51:06 (none) daemon.info charon: 08[CFG] looking for an IKEv2 config for <my IPv4>...<Initiator IPv4>
Sep  6 23:51:11 (none) daemon.info charon: 25[LIB] resolving '%any6' failed: Name or service not known
Sep  6 23:51:11 (none) daemon.info charon: 08[CFG] ike config match: 28 (%any6,...<IPv6 removed orignal IP due to privacy>/0 IKEv2)

**Additional context**
- This behavior is caused by host_create_from_dns(), which unconditionally calls lib->hosts->resolve() when host_create_from_string_and_family() returns NULL.
- For %any6 with AF_INET, the parser rejects it, and the resolver is unnecessarily invoked.
- This issue becomes noticeable when many connections are configured (mix of IPv4/IPv6). Matching for IPv4 peers can trigger multiple pointless DNS lookups.
- This unnecessary DNS resolution could add delay and so could cause failure in establishing IPsec tunnels.

-Minimal fix suggestion: skip resolver call in host_create_from_dns() if the string begins with %.

manjitduhan:
This PR adds a check in host_create_from_dns() to skip host creation and DNS resolution for %any4/%any6 macros when the address family does not match. This prevents unnecessary DNS lookups and log noise.

### DIFF
--- a/src/libstrongswan/networking/host.c
+++ b/src/libstrongswan/networking/host.c
@@ -460,7 +460,14 @@ host_t *host_create_from_dns(char *string, int af, uint16_t port)
 {
 	host_t *this;
 
-	this = host_create_from_string_and_family(string, af, port);
+	if ((af == AF_INET && streq(string, "%any6")) ||
+ 	   (af == AF_INET6 && streq(string, "%any4")))
+	{
+    		return NULL;
+	} else
+	{
+		this = host_create_from_string_and_family(string, af, port);
+	}
 	if (!this)
 	{
 		this = lib->hosts->resolve(lib->hosts, string, af);


manjitduhan:
I think, the fix raised by @tobiasbrunner via `2880-ignore-keywords-dns` branch should be taken. So, I am closing this pull request.

PR #2918, opened 2025-10-30
Avoid thread exhaustion caused by blocking CRL lookups

martinwilli:
CRL fetching uses a default timeout of 10s. If it does not complete due to DNS or TCP/HTTP timeouts, each thread performing a CRL lookup is blocked for this time. When no fresh CRL is available, many client connections can easily exhaust the thread pool, making charon unresponsive.

While reducing the timeout, increasing the tread pool size or reserving threads by priority can help, this is often insufficient. Even with a 3s timeout and 300 threads, connection setup is limited to 100 per second. When exceeding that, connection attempts may stack up and result in insufficient IKE_AUTH processing times. This may not be acceptable for a setup that uses a relaxed revocation policy and many clients.

These changes synchronize CRL fetches, so that multiple threads trying to fetch the same CRL wait for the first thread to complete the fetch. Then a negative result is cached, so that threads can fail early and setup connections with a relaxed revocation policy.

### DIFF
--- a/src/libstrongswan/plugins/revocation/revocation_fetcher.c
+++ b/src/libstrongswan/plugins/revocation/revocation_fetcher.c
@@ -26,6 +26,9 @@
 #include <credentials/certificates/ocsp_request.h>
 #include <credentials/certificates/ocsp_response.h>
 
+/* number of fetch timeouts to degrade a CRL fetch after a failure */
+#define CRL_DEGRADATION_TIMES 3
+
 typedef struct private_revocation_fetcher_t private_revocation_fetcher_t;
 
 /**
@@ -70,6 +73,11 @@ struct crl_fetch_t {
 	 * Number of threads fetching this CRL.
 	 */
 	u_int fetchers;
+
+	/**
+	 * Has the previous fetch failed, until when is this URL degraded?
+	 */
+	time_t failing;
 
 	/**
 	 * CRL received in the currently active fetch.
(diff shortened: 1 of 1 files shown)


src/libstrongswan/plugins/revocation/revocation_fetcher.c
```diff
@@ -0,0 +1,298 @@
...
+	fetch->fetchers--;
+	if (fetch->crl)
+	{
+		crl = fetch->crl->get_ref(fetch->crl);
+	}
+	fetch->condvar->signal(fetch->condvar);
+	return crl;
+}
+
+METHOD(revocation_fetcher_t, fetch_crl, certificate_t *,
```
tobiasbrunner:
```suggestion
METHOD(revocation_fetcher_t, fetch_crl, certificate_t*,
```

src/libstrongswan/plugins/revocation/revocation_fetcher.h
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Certificate fetcher performing the CRL/OCSP transfer.
+ */
+struct revocation_fetcher_t {
+
+	/**
+	 * Fetch a CRL from given URL.
+	 *
+	 * @param this		revocation fetcher
+	 * @param url		URL to retreive the CRL from
```
tobiasbrunner:
```suggestion
	 * @param url		URL to retrieve the CRL from
```

src/libstrongswan/plugins/revocation/revocation_fetcher.c
```diff
@@ -0,0 +1,298 @@
...
+	}
+	else
+	{
+		fetch->failing = time_monotonic(NULL) + timeout * CRL_DEGRATION_TIMES;
+	}
+	while (fetch->fetchers)
+	{
+		fetch->condvar->signal(fetch->condvar);
+		fetch->condvar->wait(fetch->condvar, this->mutex);
+	}
```
tobiasbrunner: Hm, this doesn't look ideal. It could potentially block this original fetcher thread indefinitely if there is a continuous stream of threads fetching the same CRL.  For negative results we're OK due to the shortcut in `wait_for_crl()` so those threads won't keep this one blocked (it's unlikely cleaning out the waiting fetchers takes longer than the timeout). Maybe a similar shortcut could be added if `this->crl` is set?

tobiasbrunner (changes requested):
Nice idea, thanks! I've found some minor style issues and typos. And it looks like there are  some potential race conditions.

PR #2954, opened 2025-12-01
credential-manage: fix leaked signature params

jfhren-stormshield:
A cloned `signature_params_t` structure was not destroyed when not added to the authentication.


tobiasbrunner:
Thanks, applied to master with updated commit message.

PR #2975, opened 2026-01-05
trap-manager: Use latest seq when creating an acquire entry

jfhren-stormshield:
When using PFKey and receiving an acquire, `data->seq` is set to `0` and a non-zero reference is fetched as a sequence number. However the fetch is done after cloning `data` leading to the sequence number within the new acquire entry to be `0`. Later when `complete()` is called, the acquire is not found by its sequence number leading to a growing number of stale acquire entries.


tobiasbrunner:
Thanks for the fix, applied to master.

PR #2982, opened 2026-01-14
vici: add configurable socket permissions and group ownership

polaz:
This patch adds support for configuring VICI socket permissions and group ownership through strongswan.conf:

```
charon {
  plugins {
    vici {
      group = vpn
      permissions = 0660
    }
  }
}
```

**Settings:**
- `charon.plugins.vici.group` - group name to own the socket (optional)
- `charon.plugins.vici.permissions` - octal permissions string (default: "0660")

**Rootless support:**
When running as non-root without CAP_CHOWN, the group setting is skipped and only permissions are applied. This enables rootless deployments where chown is not possible.

**Use case:**
Allows non-root applications (like swanctl or custom management tools) to communicate with charon via VICI without requiring root privileges or manual permission changes after daemon startup.

Related discussion: https://github.com/strongswan/strongswan/discussions/737

### DIFF
--- a/src/libcharon/plugins/vici/vici_socket.c
+++ b/src/libcharon/plugins/vici/vici_socket.c
@@ -716,7 +716,7 @@ METHOD(vici_socket_t, destroy, void,
  *
  * Reads the following settings from strongswan.conf:
  *   charon.plugins.vici.group       - group name to own the socket (optional)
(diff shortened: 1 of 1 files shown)


polaz:
Closed in favor of #2983

PR #3003, opened 2026-02-11
child-delete: Prevent redundant down event on rekeyed CHILD_SA delete timeout

martinwilli:
If a CHILD_SA is rekeyed using a CREATE_CHILD_SA request, a subsequent DELETE for the old CHILD_SA may time out. Before sending this DELETE, a CHILD_REKEYED state CHILD_SA set from `child_rekey::process_i()` is immediately set to CHILD_DELETING from `child_delete::build_i()`. If the IKE_SA dies due to a retransmission timeout of this DELETE, a redundant child-down event is issued for the rekeyed CHILD_SA that has already seen a child-rekey event.

A reproducer shows the following log and events:

     [CFG] vici rekey CHILD_SA #533
     [IKE] establishing CHILD_SA XXX{534} reqid 20
     [ENC] generating CREATE_CHILD_SA request 0 [ N(REKEY_SA) SA No KE TSi TSr ]
     [ENC] parsed CREATE_CHILD_SA response 0 [ SA No TSi TSr ]
     [IKE] rekeyed CHILD_SA XXX{533} with SPIs ca997de6_i cd27d4fe_o with XXX{534} with SPIs ced1cd01_i c460a7c9_o
     Event: child-rekey
       [OLD SA] state: REKEYING, spi-in: ca997de6
       [NEW SA] state: INSTALLED, spi-in: ced1cd01
     [IKE] closing CHILD_SA XXX{533} with SPIs ca997de6_i (352 bytes) cd27d4fe_o (264 bytes) and TS 0.0.0.0/0 === 10.11.9.40/29
     [IKE] sending DELETE for ESP CHILD_SA with SPI ca997de6
     [ENC] generating INFORMATIONAL request 1 [ D ]
     [IKE] retransmit 1 of request with message ID 1
     [IKE] retransmit 2 of request with message ID 1
     [IKE] retransmit 3 of request with message ID 1
     [IKE] retransmit 4 of request with message ID 1
     [IKE] giving up after 4 retransmits
     Event: child-updown
       [SA] state: DELETING, spi-in: ca997de6
     Event: child-updown
       [SA] state: INSTALLED, spi-in: ced1cd01

To avoid the redundant child-down event for the successfully rekeyed CHILD_SA, keep a CHILD_REKEYED CHILD_SA state during DELETE. Adjust tests for the new state and handle a DELETE for CHILD_REKEYED SAs with a TEMPORARY_FAILURE.

A new exchange test exercises that a delete timeout after rekeying does not cause a duplicate child-down event.

### DIFF
--- a/src/libcharon/bus/bus.c
+++ b/src/libcharon/bus/bus.c
@@ -830,11 +830,23 @@ METHOD(bus_t, ike_updown, void,
 		enumerator = ike_sa->create_child_sa_enumerator(ike_sa);
 		while (enumerator->enumerate(enumerator, (void**)&child_sa))
 		{
-			if (child_sa->get_state(child_sa) != CHILD_REKEYED &&
-				child_sa->get_state(child_sa) != CHILD_DELETED)
+			switch (child_sa->get_state(child_sa))
 			{
-				child_updown(this, child_sa, FALSE);
+				case CHILD_REKEYED:
+				case CHILD_DELETED:
+					continue;
+				case CHILD_DELETING:
+					if (child_sa->get_outbound_state(child_sa) ==
+						CHILD_OUTBOUND_NONE)
+					{
+						/* deleting CHILD_SA has been rekeyed, omit event */
+						continue;
+					}
+					break;
+				default:
+					break;
 			}
+			child_updown(this, child_sa, FALSE);
 		}
 		enumerator->destroy(enumerator);
 	}
--- a/src/libcharon/tests/suites/test_child_rekey.c
+++ b/src/libcharon/tests/suites/test_child_rekey.c
@@ -1090,6 +1090,64 @@ START_TEST(test_regular_responder_incorrect_delete)
 END_TEST
 
 /**
+ * Check that a delete timeout after rekeying does not cause a duplicate
+ * child-down event for the rekeyed SA.
+ */
+START_TEST(test_regular_delete_timeout)
+{
+	ike_sa_t *a, *b;
+	message_t *msg;
+	status_t s;
+
+	exchange_test_helper->establish_sa(exchange_test_helper, &a, &b, NULL);
+	initiate_rekey(a, 1);
+	assert_ipsec_sas_installed(a, 1, 2);
+
+	/* this should never get called as this results in a successful rekeying */
+	assert_hook_not_called(child_updown);
+
+	/* CREATE_CHILD_SA { N(REKEY_SA), SA, Ni, [KEi,] TSi, TSr } --> */
+	assert_hook_not_called(child_rekey);
+	assert_notify(IN, REKEY_SA);
+	exchange_test_helper->process_message(exchange_test_helper, b, NULL);
+	assert_child_sa_state(b, 2, CHILD_REKEYED, CHILD_OUTBOUND_INSTALLED);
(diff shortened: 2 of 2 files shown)


tobiasbrunner:
Nice catch, thanks!

I wonder if this actually would cause issues in practice. For one, the complete IKE SA is torn down afterwards anyway. And when tracking Child SAs wouldn't one ignore the old SA because there was a switch to the new one with the `child-rekey` event? Did it actually cause problems in your case?

One minor detail regarding the changes is that we'd send back a DELETE if the peer deletes the same SA (it shouldn't of course), which we wouldn't do previously due to the change to `CHILD_DELETING`.

A bit trickier could be that  `CHILD_DELETING` is handled specially in some situations (e.g. when deciding whether to reject a concurrent IKE rekeying or for start actions). Maybe an alternative fix would be to change the `if`-statement in `bus_t::ike_updown` like this:

```diff
diff --git a/src/libcharon/bus/bus.c b/src/libcharon/bus/bus.c
index 99387d5e06f8..8ebc9fc93c61 100644
--- a/src/libcharon/bus/bus.c
+++ b/src/libcharon/bus/bus.c
@@ -830,11 +830,21 @@ METHOD(bus_t, ike_updown, void,
 		enumerator = ike_sa->create_child_sa_enumerator(ike_sa);
 		while (enumerator->enumerate(enumerator, (void**)&child_sa))
 		{
-			if (child_sa->get_state(child_sa) != CHILD_REKEYED &&
-				child_sa->get_state(child_sa) != CHILD_DELETED)
+			switch (child_sa->get_state(child_sa))
 			{
-				child_updown(this, child_sa, FALSE);
+				case CHILD_DELETING:
+					if (child_sa->get_outbound_state(child_sa) != CHILD_OUTBOUND_NONE)
+					{
+						break;
+					}
+					/* fall-through only for rekeyed SAs we are deleting */
+				case CHILD_REKEYED:
+				case CHILD_DELETED:
+					continue;
+				default:
+					break;
 			}
+			child_updown(this, child_sa, FALSE);
 		}
 		enumerator->destroy(enumerator);
 	}
```

martinwilli:
> I wonder if this actually would cause issues in practice. [...] Did it actually cause problems in your case?

In our specific setup, it did. In an external application, using vici we track:

* Remote subnets of established tunnels for propagation to BGP
* Multiple tunnels with the same selector (but different marks) are very common, hence subnets are refcounted

An earlier version of this application just tracked subnets but not individual tunnels, and the extra child-down event broke the refcounting, withdrawing BGP prefixes while tunnels are active. The application has been improved, but I guess the additional event can be problematic in specific cases.

> Maybe an alternative fix would be to change the `if`-statement in `bus_t::ike_updown` like this [...]

Your approach looks much simpler, and certainly has less side effects. I'll give it a try, thanks!

tobiasbrunner:
> An earlier version of this application just tracked subnets but not individual tunnels, and the extra child-down event broke the refcounting

I see. Thanks for the clarification.

tobiasbrunner:
Thanks, looks good. Just a nit: the commit messages still says `child-delete: ...` while the main change is now in `bus_t`, so that should probably be `bus: ...`.

martinwilli:
> Just a nit: the commit messages still says `child-delete: ...` while the main change is now in `bus_t`, so that should probably be `bus: ...`.

I've fixed that, thanks.

I did some more testing and all looks fine, so from my end this is good to go...

PR #3004, opened 2026-02-11
oss-fuzz: Add new fuzzer targeting libtls

arthurscchan:
This PR adds a new fuzzer targeting libtls implementation. There is also a set of corpora added in the fuzzing-corpora repository for this new fuzzer (https://github.com/strongswan/fuzzing-corpora/pull/3).

Remark: This is the third in a series of PRs intended to expand fuzzing coverage for the StrongSwan project. Subsequent PRs will continue to address further coverage gaps and enhance fuzzing effectiveness, thereby improving the detection of issues across the codebase.

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -12,3 +12,4 @@ fuzz_pa_tnc
 fuzz_pb_tnc
 fuzz_radius
 fuzz_vici
+fuzz_tls
--- a/fuzz/Makefile.am
+++ b/fuzz/Makefile.am
@@ -5,6 +5,7 @@ AM_CPPFLAGS = @CPPFLAGS@ \
 	-I$(top_srcdir)/src/libimcv \
 	-I$(top_srcdir)/src/libtncif \
 	-I$(top_srcdir)/src/libtpmtss \
+	-I$(top_srcdir)/src/libtls \
 	-I$(top_srcdir)/src/libtnccs \
 	-I$(top_srcdir)/src/libtnccs/plugins/tnccs_20 \
 	-I$(top_srcdir)/src/libradius
@@ -47,6 +48,10 @@ radius_ldflags = \
 	$(top_builddir)/src/libradius/.libs/libradius.a \
 	$(fuzz_ldflags)
 
+tls_ldflags = \
+	$(top_builddir)/src/libtls/.libs/libtls.a \
+	$(fuzz_ldflags)
+
 # fuzzers that use crypto plugins in a significant way
 fuzzers_with_plugins = \
 	fuzz_certs fuzz_crls fuzz_ocsp_req fuzz_ocsp_rsp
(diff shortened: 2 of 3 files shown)


tobiasbrunner (changes requested):
Thanks, I've added some comments.

tobiasbrunner:
Thanks, interesting approach with the combinations. I wonder if `build()` should be called first in the `!is_server && test_build` case. Then the input would be treated as response to the initial request. However, that input could then never be valid for the `is_server` case, but that's also true vice-versa - so maybe marking inputs as request/response with a flag, which is then mapped to `is_server`, might actually make sense. Or how do you see that? And does `test_build` even have to be a flag? Couldn't `build()` just get called depending on the status returned by `process()` (or if  `!is_server`).

arthurscchan:
Hi, sorry for the long hanging of this PR because I was stuck on other things.

Yes, I do agree that the build/process approach should be more controlled for server and client side handling which make less false cases that returns early. I will change the approach and make that flow more coherence. And the test_build may not be necessary at all. I have changed the code a bit and now doing testing. Will ping you again once it is good in my end. Thanks again for reviewing the fuzzers.

arthurscchan:
@tobiasbrunner I think the CI is failing on the configure.ac version check because the jumping to 6.0.6 does not have any tag yet.

tobiasbrunner:
The tag has definitely been there since the release. And that this check won't work when cloning with `--depth 1` is also normal. What's weird, however, is that this does not result in the expected `checking version from Git repository... UNKNOWN (could be due to insufficient clone depth)` message but this error. I think the reason is that you merged the current master into your branch (which was based on a commit between 6.0.5dr1 and 6.0.5) instead of rebasing it to master.  Could you try the latter?

arthurscchan:
Thanks for the advice, I will try and fix it on my side.

arthurscchan:
I thinks thats good now. Sorry for the merging and rebase issue.

tobiasbrunner:
Great, thanks for the update. I've pushed a modified patch to the `3004-fuzz-tls` branch. Besides some small code style changes, the most significant update is that I added a server certificate. That allows the server to fully process a valid request and produce a response.

Without a more advanced fuzzing approach that incorporated crypto and protocol details (i.e. would perform an actual key exchange and react to the messages), we can't expect much more I think. During testing I noticed that the client in particular rejects responses if the public DH value in the response doesn't match the expected length for the method it selected for its request (seems to default to ECP-256). I guess the fuzzer will notice that and find a way around it (maybe it even triggers a restart by requesting a different DH method from the client, TLS 1.3 has such a mechanism). I guess we'll have to see.

Please let me know what you think. And could you perhaps update https://github.com/strongswan/fuzzing-corpora/pull/3 to match the new logic?

arthurscchan:
Yeah, I think that is good enough approach to have a fixed server certificate. I think it is good to merge in as is right now. I will handle the corpora update and let you know on that PR. In general, OSS-Fuzz uses coverage guided approach so if that rejecting path is not triggering more coverage, OSS-Fuzz should able to tune the random input to eventually allow that to get past the blockers. I will see if my changes to the corpora can help OSS-Fuzz to go around that much faster.

tobiasbrunner:
OK, thanks. I've pushed the patch to master.

PR #3005, opened 2026-02-15
Fix: whitelist plugin improperly handling client disconnection, causing DoS

seantywork:
# Environment
- os: Ubuntu 24.04
- compiler: gcc(Ubuntu 13.3.0-6ubuntu2~24.04)
# Strongswan version: 
- 6.0.4
# Build configuration
```shell
./configure --prefix=/usr --sysconfdir=/etc --enable-charon --enable-systemd --disable-defaults --enable-static --enable-test-vectors --enable-pki --enable-ikev2 --enable-vici --enable-swanctl --enable-ldap --enable-pkcs11 --enable-tpm --enable-aesni --enable-aes --enable-rc2 --enable-sha2 --enable-sha1 --enable-md5 --enable-mgf1 --enable-rdrand --enable-random --enable-nonce --enable-x509 --enable-revocation --enable-constraints --enable-pubkey --enable-pkcs1 --enable-pkcs7 --enable-pkcs8 --enable-pkcs12 --enable-pgp --enable-dnskey --enable-sshkey --enable-pem --enable-openssl --enable-gcrypt --enable-af-alg --enable-fips-prf  --enable-gmp  --enable-curve25519 --enable-agent --enable-chapoly --enable-xcbc --enable-cmac --enable-hmac --enable-ctr --enable-ccm --enable-gcm --enable-ntru --enable-drbg --enable-curl --enable-attr --enable-kernel-netlink --enable-resolve --enable-socket-default --enable-connmark --enable-forecast --enable-farp --enable-stroke --enable-vici --enable-updown --enable-eap-identity --enable-eap-aka --enable-eap-md5 --enable-eap-gtc --enable-eap-mschapv2 --enable-eap-dynamic --enable-eap-radius --enable-eap-tls --enable-eap-ttls --enable-eap-peap --enable-eap-tnc --enable-xauth-generic --enable-xauth-eap --enable-xauth-pam --enable-tnc-tnccs --enable-dhcp --enable-lookip --enable-error-notify --enable-certexpire --enable-led --enable-addrblock --enable-unity --enable-counters --enable-whitelist --enable-sqlite --enable-attr-sql
```
# Strongswan configuration
- everything default

# Issue
In whitelist plugin, client disconnection is improperly handled, resulting in no file descriptor left for new connection on whitelist socket or any other sockets.

With no previous condition setting, just run the below script to observe the problem.

```shell
#!/bin/bash
for i in $(seq 1 10000);
do
        sudo /usr/libexec/ipsec/whitelist enable
        echo "iter: $i"
done
```
It will stuck at some point around 100X, mine being constantly 1007.

At first, as I attached `gdb` to figure out what was going on, it seemed some sort of a race condition. 

But it turned out, `notify_end` fails to remove entry after `notify_async` because `on_read` called by `notify_async` doesn't return FALSE on client disconnection as it hangs on condvar->wait in watcher->remove , which resulting in pc never hitting close(fd) within destory() in stream.c. 

```shell
# breakpoints on watcher remove and close(stream fd) within destroy() in stream.c
(gdb) br stream.c:262
Breakpoint 1 at 0x742c18af9d04: file networking/streams/stream.c, line 262.
(gdb) br stream.c:263
Breakpoint 2 at 0x742c18af9d22: file networking/streams/stream.c, line 263.
(gdb) c
Continuing.
# on another terminal
@seantywork ➜ ~/box $ sudo /usr/libexec/ipsec/whitelist enable
# watcher remove hit, but close(fd) not getting hit
[Switching to Thread 0x742c101206c0 (LWP 130764)]

Thread 4 "charon-systemd" hit Breakpoint 1, destroy (this=0x742bec000c80) at networking/streams/stream.c:262
262             lib->watcher->remove(lib->watcher, this->fd);
(gdb) c
Continuing.

``` 
Ultimately, fd limit per process will be reached, leaving no fd for charon process to take up.

# Solution
Use the same `lib->processor->queue_job` approach as vici socket plugin does to asynchronously handle stream closure and immediately return FALSE from `on_read`, which prevent detroy() from hanging because of infinite is_in_callback == true status.
Tested with the same script above.

# Further consideration
Might need to apply this to other plugins where direct calls to destroy within `notify_async` are used

### DIFF
--- a/src/libcharon/plugins/whitelist/whitelist_control.c
+++ b/src/libcharon/plugins/whitelist/whitelist_control.c
@@ -25,6 +25,7 @@
 
 #include <daemon.h>
 #include <collections/linked_list.h>
+#include <processing/jobs/callback_job.h>
 
 #include "whitelist_msg.h"
 
@@ -50,6 +51,50 @@ struct private_whitelist_control_t {
 	 */
 	stream_service_t *service;
 };
+
+/**
+ * Information about a client connection.
+ */
+typedef struct {
+	private_whitelist_control_t *this;
+	whitelist_msg_t msg;
+	size_t read;
+} whitelist_conn_t;
+
+
+/**
+ * Information needed to delete client connection.
+ */
+typedef struct {
+	whitelist_conn_t *conn;
+	stream_t* stream;
+} whitelist_conn_del_t;
+
+
+
+/**
+ * Asynchronous callback to disconnect client
+ */
+CALLBACK(disconnect_async, job_requeue_t,
+	whitelist_conn_del_t *wcd)
+{
+	wcd->stream->destroy(wcd->stream);
+	free(wcd->conn);
+	return JOB_REQUEUE_NONE;
+}
+
+/**
+ * Disconnect a connected client
+ */
+static void disconnect(whitelist_conn_t *conn, stream_t *stream){
+	whitelist_conn_del_t *wcd;
+	INIT(wcd,
+		.conn = conn,
+		.stream = stream,
+	);
+	lib->processor->queue_job(lib->processor,
+		(job_t*)callback_job_create(disconnect_async, wcd, free, NULL));
+}
 
 /*
  * List whitelist entries using a read-copy
@@ -92,14 +137,6 @@ static void list(private_whitelist_control_t *this,
 	stream->write_all(stream, &msg, sizeof(msg));
 }
 
-/**
- * Information about a client connection.
- */
-typedef struct {
-	private_whitelist_control_t *this;
-	whitelist_msg_t msg;
-	size_t read;
-} whitelist_conn_t;
 
 /**
  * Dispatch a received message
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
Thanks. Looks like the same issue was fixed for the `lookip` plugin with 961409b66858facdad00e4bf38f534ced3c1255e over ten years ago. When refactoring the error handling in the `whitelist` plugin with 85ebf6abd4412d241fd55bc076b60f0a87931540 last year, I missed that calling `stream_t::destroy` from the callback is problematic.  I've pushed your commit with some modifications to the `3005-whitelist-disconnect` branch (updated commit message, some code style issues and alignment with the code in `lookip`, also added a commit with some documentation changes for `stream_t`).

> Ultimately, fd limit per process will be reached, leaving no fd for charon process to take up.

That's not actually the issue. The threads deadlock way before that.

taehyy00nn:
Thanks for the clarification, I've failed to mention that `no fds available` as a side effect of the deadlock

tobiasbrunner:
Changes were merged with 14cbe0bf242319ec468b7b857030ef1095ad8af7 and 589c74009b6a88c39f29c5980b9068f4731d9aae.

PR #3007, opened 2026-02-16
Support remote-psk option for charon-nm.

phlibi:
NetworkManager .nmconnection profiles can now contain a "remote-psk" option, which, if set to some non-empty password/key, configures a preshared key to use to authenticate the gateway independently of the client authentication.

Consider this plain strongswan/charon configuration:

```
  connections {
 
    swivpn {
      remote_addrs = 1.2.3.4
      vips = 0.0.0.0
 
      local {   
        auth = eap-mschapv2
        eap_id = phip
      }
      remote {    
        auth = psk
        id = 1.2.3.4
      }
      children {
        swivpn {
          remote_ts  = 0.0.0.0/0  
          start_action = start   
        }  
      }  
    }  
  }  
 
  secrets {
    ike-swivpn {
      id = 1.2.3.4
      secret = TheVPNPSK
    }
    eap-phip { 
      id = phip
      secret = "TheUserPassword"
    }
  }
```

Local (client) authentication is done using EAP-MSCHAPv2, while the gateway is authenticated using a preshared key (PSK). This PR allows the same configuration to be used with a NetworkManager profile. The matching `/etc/NetworkManager/system-connections/swivpn.nmconnection` would look like:

```
[connection]
id=swivpn
uuid=aaa-bbb-ccc
type=vpn

[vpn]
address=1.2.3.4
encap=yes
ipcomp=no
remote-psk=TheVPNPSK
method=eap
password-flags=0
proposal=no
user=phip
virtual=yes
service-type=org.freedesktop.NetworkManager.strongswan

[vpn-secrets]
password=TheUserPassword

[ipv4]
method=auto

[ipv6]
addr-gen-mode=default
method=auto

[proxy]
```

Please note that this PR *only adds support to charon-nm* to process such extended profile files. It does *not add an option to the GUI to graphically create such a profile*.

I made this extension based on an immediate need by myself, so it is a rather rigid solution. I'd also be happy to see something more flexible implemented, as other people might have different requirements for asymmetric authentication. This change only supports the case of a PSK for the gateway and some (other) authentication for the client/user/roadwarrior, which is the default on FortiGate firewalls for their FortiClient software. See also this discussion: https://github.com/strongswan/strongswan/discussions/3000

Side note: An initial version of this update was generated using an AI tool, with subsequent review and refinement. This process allowed me to implement the change much more quickly than if I first needed to go through the whole source code to find the right spots to change.

### DIFF
--- a/src/charon-nm/nm/nm_creds.c
+++ b/src/charon-nm/nm/nm_creds.c
@@ -60,6 +60,16 @@ struct private_nm_creds_t {
 	chunk_t keyid;
 
 	/**
+	 * Remote peer ID for PSK authentication
+	 */
+	identification_t *remote_id;
+
+	/**
+	 * Remote peer PSK
+	 */
+	char *remote_psk;
+
+	/**
 	 * users certificate
 	 */
 	certificate_t *usercert;
@@ -238,8 +248,17 @@ METHOD(credential_set_t, create_shared_enumerator, enumerator_t*,
 
 	switch (type)
 	{
-		case SHARED_EAP:
 		case SHARED_IKE:
+			/* Check for remote PSK */
+			if (this->remote_psk && this->remote_id &&
+				other && this->remote_id->equals(this->remote_id, other))
+			{
+				key = chunk_create(this->remote_psk, strlen(this->remote_psk));
+				break;
+			}
+			/* Fall-through is intentional, to use password also as PSK if no
+			 * dedicated remote PSK has been set */
+		case SHARED_EAP:
 			if (!this->pass || !this->user)
 			{
 				goto no_secret;
@@ -379,6 +398,17 @@ METHOD(nm_creds_t, set_pin, void,
 	this->lock->unlock(this->lock);
 }
 
+METHOD(nm_creds_t, set_remote_psk, void,
+	private_nm_creds_t *this, identification_t *id, char *psk)
+{
+	this->lock->write_lock(this->lock);
+	DESTROY_IF(this->remote_id);
+	this->remote_id = id ? id->clone(id) : NULL;
+	free(this->remote_psk);
+	this->remote_psk = strdupnull(psk);
+	this->lock->unlock(this->lock);
+}
+
 METHOD(nm_creds_t, set_cert_and_key, void,
 	private_nm_creds_t *this, certificate_t *cert, private_key_t *key)
 {
@@ -403,6 +433,8 @@ METHOD(nm_creds_t, clear, void,
 	free(this->pass);
 	free(this->keypass);
 	free(this->keyid.ptr);
+	free(this->remote_psk);
+	DESTROY_IF(this->remote_id);
 	DESTROY_IF(this->usercert);
 	DESTROY_IF(this->key);
 	this->key = NULL;
@@ -411,6 +443,8 @@ METHOD(nm_creds_t, clear, void,
 	this->user = NULL;
 	this->keypass = NULL;
 	this->keyid = chunk_empty;
+	this->remote_psk = NULL;
+	this->remote_id = NULL;
 }
 
 METHOD(nm_creds_t, destroy, void,
@@ -443,6 +477,7 @@ nm_creds_t *nm_creds_create()
 			.set_username_password = _set_username_password,
 			.set_key_password = _set_key_password,
 			.set_pin = _set_pin,
+			.set_remote_psk = _set_remote_psk,
 			.set_cert_and_key = _set_cert_and_key,
 			.clear = _clear,
 			.destroy = _destroy,
@@ -450,6 +485,8 @@ nm_creds_t *nm_creds_create()
 		.lock = rwlock_create(RWLOCK_TYPE_DEFAULT),
 		.certs = linked_list_create(),
 	);
+	this->remote_id = NULL;
+	this->remote_psk = NULL;
 	return &this->public;
 }
 
--- a/src/charon-nm/nm/nm_creds.h
+++ b/src/charon-nm/nm/nm_creds.h
@@ -74,6 +74,14 @@ struct nm_creds_t {
 	void (*set_pin)(nm_creds_t *this, chunk_t keyid, char *pin);
 
 	/**
+	 * Set the pre-shared key for remote peer authentication.
+	 *
+	 * @param id		ID of the remote peer
+	 * @param psk		pre-shared key
+	 */
+	void (*set_remote_psk)(nm_creds_t *this, identification_t *id, char *psk);
+
+	/**
 	 * Set the certificate and private key to use for client authentication.
 	 *
 	 * @param cert		client certificate
--- a/src/charon-nm/nm/nm_service.c
+++ b/src/charon-nm/nm/nm_service.c
@@ -850,7 +850,7 @@ static gboolean connect_(NMVpnServicePlugin *plugin, NMConnection *connection,
 	NMSettingVpn *vpn;
 	enumerator_t *enumerator;
 	identification_t *gateway = NULL;
-	const char *str, *method, *user;
+	const char *str, *remote_psk, *method, *user;
 	chunk_t safe_file;
 	bool virtual, proposal;
 	proposal_t *prop;
@@ -931,6 +931,7 @@ static gboolean connect_(NMVpnServicePlugin *plugin, NMConnection *connection,
 	 * Register credentials
 	 */
 	priv->creds->clear(priv->creds);
+	remote_psk = nm_setting_vpn_get_data_item(vpn, "remote-psk");
 
 	/* gateway/CA cert */
 	str = nm_setting_vpn_get_data_item(vpn, "certificate");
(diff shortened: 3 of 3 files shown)


tobiasbrunner:
As stated before, what FortiGate is doing is not standard-compliant. EAP client authentication MUST be combined with public-key server authentication in IKEv2 (see [RFC 7296, section 2.16](https://datatracker.ietf.org/doc/html/rfc7296#section-2.16)). strongSwan allows configuring it due to it's flexibility. But it's not something we'd ever recommend doing due to its inherent risks. Also see [this FAQ entry](https://docs.strongswan.org/docs/latest/support/faq.html#_pre_shared_key_authentication) for why using PSK authentication isn't recommended in general (but especially not in roadwarrior scenarios where the PSK is shared between multiple clients).

phlibi:
@tobiasbrunner Fair point. However, I guess it would be rather hard to convince Fortinet that they're doing things wrong, so we'll need to somehow work with what we have... Additionally, a PSK (as long as it's not publicly known) can still serve as a good first line of defense, both against weak user credentials and also against EAP DoS attacks. But of course it's not ideal needing to share a secret across many different parties.

tobiasbrunner:
> However, I guess it would be rather hard to convince Fortinet that they're doing things wrong

You won't be able to use it without their proprietary app (or manually configured strongSwan) as no other IKEv2 client (Windows, iOS/macOS, Android) supports this scheme.

> so we'll need to somehow work with what we have...

Then use a certificate. They support ACME to generate a certificate automatically as far as I saw (and even if that doesn't work for IKEv2, you should be able to load a server certificate into the box).


> Additionally, a PSK (as long as it's not publicly known) can still serve as a good first line of defense, both against weak user credentials

Depends on the strength of the PSK. Because any random client can easily get a hash just by connecting to the server and then attack it offline.

> and also against EAP DoS attacks.

It does not. The initiator doesn't need to know the PSK, it can just ignore the server's AUTH payload in the IKE_AUTH response and continue with the EAP authentication the server initiated. That's different in IKEv1 where the PSK was mixed into the IKE keys and the server wouldn't be able to decrypt the request, that's not the case with IKEv2.

phlibi:
> You won't be able to use it without their proprietary app (or manually configured strongSwan) as no other IKEv2 client (Windows, iOS/macOS, Android) supports this scheme.

Probably. But again, I can't change their product.

> Then use a certificate. They support ACME to generate a certificate automatically as far as I saw (and even if that doesn't work for IKEv2, you should be able to load a server certificate into the box).

I checked that as well and indeed I've now found a way to connect both with their FortiClient (wasn't able to do it with the Windows native client yet) and strongswan using a server and client certificate and no other credentials. That could be an option, but now I'm struggling with the identification of the user on the firewall, so that different users would get access to different parts of the system via policies. It looks that this is only supported using some kind of EAP, even though every single user is authenticated using a separate client certificate. I've really gone through quite a lot of variants now, many of them looking promising, only to fail at some point :(

> It does not. The initiator doesn't need to know the PSK, it can just ignore the server's AUTH payload in the IKE_AUTH response and continue with the EAP authentication the server initiated. That's different in IKEv1 where the PSK was mixed into the IKE keys and the server wouldn't be able to decrypt the request, that's not the case with IKEv2.

Thanks, that's good to know. I know from OpenVPN that they recommend to use a shared key (actually TLS-AUTH or something) to keep the bad guys from even trying out passwords and figured that IKEv2 would use the PSK in the same way on the server. Well, apparently not, so that makes the PSK+EAP approach definitely less appealing.

I really haven't decided about what to do about it now. I'm currently using my certificate-based access to test the IPSec VPN in general, finding out if it really provides a good enough user experience from the point of stability and mobility. So far it looks good, but you never know what will come.

If Fortinet would just support WireGuard, that would make life so much easier...

tobiasbrunner:
> It looks that this is only supported using some kind of EAP, even though every single user is authenticated using a separate client certificate.

At least via RADIUS it would seem to be possible [to use EAP-TLS](https://docs.fortinet.com/document/fortigate/7.6.6/administration-guide/726232/windows-ikev2-native-vpn-with-user-certificate) and assign users to groups for firewalling. Their local RADIUS server might not support that though.

They also seem to support multiple authentication rounds (first a client certificate, then EAP). They have something like that [described in their docs](https://docs.fortinet.com/document/fortigate/7.6.6/administration-guide/298520/ipsec-ikev2-vpn-2fa-with-eap-and-certificate-authentication). But not sure if they actually implement [RFC 4739](https://datatracker.ietf.org/doc/html/rfc4739), which strongSwan supports, or use some proprietary extension. None of the built-in clients in major OSes support that RFC anyway (and our NM plugin currently doesn't either, but our Android app does and you can configure it via swanctl).

> Well, apparently not, so that makes the PSK+EAP approach definitely less appealing.

Note that such active attacks against EAP could also be an issue when using a server certificate with client-side EAP. You can really only avoid issues with weak passwords by using client certificates (or a strong, client-specific PSK without EAP).

phlibi:
> At least via RADIUS it would seem to be possible [to use EAP-TLS](https://docs.fortinet.com/document/fortigate/7.6.6/administration-guide/726232/windows-ikev2-native-vpn-with-user-certificate) and assign users to groups for firewalling. Their local RADIUS server might not support that though.

Indeed, that's what I've found so far. Since we already use cloud-based SSO for a lot of services, spinning up another RADIUS server appears to only add a lot of complexity that we actually don't want. And yes, the internal one doesn't appear to support EAP-TLS, though I haven't found a clear statement about that yet.

> They also seem to support multiple authentication rounds (first a client certificate, then EAP). They have something like that [described in their docs](https://docs.fortinet.com/document/fortigate/7.6.6/administration-guide/298520/ipsec-ikev2-vpn-2fa-with-eap-and-certificate-authentication). But not sure if they actually implement [RFC 4739](https://datatracker.ietf.org/doc/html/rfc4739), which strongSwan supports, or use some proprietary extension. None of the built-in clients in major OSes support that RFC anyway (and our NM plugin currently doesn't either, but our Android app does and you can configure it via swanctl).

That's also something I've tried, but to no avail. And as you wrote, it isn't supported by most clients (including charon-nm), so it's also not really a viable option.

> Note that such active attacks against EAP could also be an issue when using a server certificate with client-side EAP. You can really only avoid issues with weak passwords by using client certificates (or a strong, client-specific PSK without EAP).

Absolutely. That's why I'm currently trying to find a working solution without EAP, which I've basically found with certificates, except for the noted downside of not being able to match firewall policies to individual users. If this would work, I'd probably have found a solution for all our VPN needs. I've already asked for this in the Fortinet forum, but so far without a positive response: https://community.fortinet.com/t5/Support-Forum/Firewall-policy-match-based-on-IPSec-user-PKI/td-p/431252
I think I'll ask their support directly, if not for a solution, then at least for a clear statement that it doesn't (currently) work so that I can stop wasting time on this approach.

My last resort would be to deploy a small external server having a static IPSec tunnel to our firewall and then letting users connect through WireGuard. Policy matching could then be done based on the user's assigned IP address. But that's also only a workaround and I'd really prefer to have a VPN solution to the firewall directly, also to avoid another point of failure, administration need, etc...

Anyway, thanks for your support. As I understand, this PR won't be merged anyways, so should I just close it?

tobiasbrunner:
> And as you wrote, it isn't supported by most clients (including charon-nm)

I'd be open to add a certificate+EAP option to the NM plugin (analogous to the Android app). Could be a bit tricky because multiple secrets could be necessary via auth-dialog, though.

By the way, if the box actually supported RFC 4739, you'd see `N(MULT_AUTH) ` in the IKE_SA_INIT response in the strongSwan log.


> I've already asked for this in the Fortinet forum, but so far without a positive response:

I see. Looks relatively straight forward. Would be unfortunate if that doesn't work somehow.



> As I understand, this PR won't be merged anyways, so should I just close it?

Yes, I don't intend to merge it. Thanks anyway.

phlibi:
> By the way, if the box actually supported RFC 4739, you'd see N(MULT_AUTH) in the IKE_SA_INIT response in the strongSwan log.

I'm still investigating about this possibility, but at least in a quick test, this capability (or however this is called for IPSec), is not presented when the firewall is configured according to the guide for Cert+EAP authentication:

```
[IKE] initiating IKE_SA swivpn[1] to 1.2.3.4
[ENC] generating IKE_SA_INIT request 0 [ SA KE No N(NATD_S_IP) N(NATD_D_IP) N(FRAG_SUP) N(HASH_ALG) N(REDIR_SUP) ]
[NET] sending packet: from 192.168.1.13[500] to 1.2.3.4[500] (304 bytes)
[NET] received packet: from 1.2.3.4[500] to 192.168.1.13[500] (289 bytes)
[ENC] parsed IKE_SA_INIT response 0 [ SA KE No N(NATD_S_IP) N(NATD_D_IP) CERTREQ N(FRAG_SUP) ]
[CFG] selected proposal: IKE:AES_CBC_128/HMAC_SHA2_256_128/PRF_HMAC_SHA2_256/ECP_384
[IKE] local host is behind NAT, sending keep alives
[IKE] received cert request for "CN=CA1"
[IKE] sending cert request for "CN=CA1"
[IKE] establishing CHILD_SA swivpn{1}
[ENC] generating IKE_AUTH request 1 [ IDi N(INIT_CONTACT) CERTREQ IDr CPRQ(ADDR DNS) SA TSi TSr N(MOBIKE_SUP) N(ADD_4_ADDR) N(ADD_4_ADDR) N(ADD_4_ADDR) N(ADD_4_ADDR) N(EAP_ONLY) N(MSG_ID_SYN_SUP) ]
```

So I suppose they don't do it according to RFC 4739.

tobiasbrunner:
> So I suppose they don't do it according to RFC 4739.

Yeah, doesn't look like they do.

PR #3027, opened 2026-03-11
OSS-Fuzz: Add fuzzer targeting Radius messages

arthurscchan:
This PR creates a new fuzzer targeting Radius messages processing in the strongswan project.

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -6,3 +6,4 @@ fuzz_ocsp_req
 fuzz_ocsp_rsp
 fuzz_pa_tnc
 fuzz_pb_tnc
+fuzz_radius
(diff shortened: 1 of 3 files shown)


tobiasbrunner (changes requested):
Thanks, looks pretty good.

tobiasbrunner:
Thanks for the updates. I've pushed a modified version of this patch to the `3027-fuzz-radius` branch. It's rebased to the `3026-fuzz-vici` branch. Similar changes here (`radius_message_parse()` does also not parse anything yet, but it does some checks, so it can fail). Note that calling `verify()` requires loading plugins and creating an MD5 hasher and an HMAC-MD5 signer, otherwise it crashes pretty much immediately (or a bit later).

PR #3043, opened 2026-04-08
fix: UB in cpu_feature_t enum from signed left shift into sign bit

PatriciaBucataru:
CPU_FEATURE_PADLOCK_PMM_ENABLED uses (1 << 31) where 1 is a signed int. Per C11 6.5.7, shifting a signed value such that the result is not representable in the result type is undefined behaviour, since 1 * 2^31 exceeds INT_MAX.

Changed to (1u << 31).


tobiasbrunner:
Thanks. While this might technically be undefined behavior, I don't think it is in practice as reasonable compilers will use unsigned integers for enumerations if no signed value is defined. The [GCC docs e.g. state](https://gcc.gnu.org/onlinedocs/gcc-11.4.0/gcc/Structures-unions-enumerations-and-bit-fields-implementation.html):

>  Normally, the type is unsigned int if there are no negative values in the enumeration, otherwise int.

I assume clang does the same. Please also note that we use this in other enums as well (`ike_extension_t`, `ike_condition_t` and `pb_tnc_mutual_protocol_type_t`) and no compiler in our CI ever complained when compiling with `-Wall -Wextra`.

PR #3044, opened 2026-04-08
fix: null dereference in asn1_build_object() when malloc fails

PatriciaBucataru:
malloc() return value was not checked. If allocation fails, pos is set to NULL and the subsequent *pos++ = type dereferences it, crashing the process.

Added a null check that sets the object to chunk_empty and returns NULL on allocation failure.


tobiasbrunner:
Thanks, but since we don't check the return value of `malloc()` anywhere in our code base, I don't see why we should make an exception in this particular case (and callers of `asn1_build_object()` don't expect a NULL pointer to get returned either).

PR #3067, opened 2026-05-20
vici: Include proposals in connection listings

bk203:
## Summary
This adds configured proposal output to VICI `list-conns` / `list-conn` responses.

Our workflow validates received configuration by comparing it with the runtime configuration that was already committed to VICI via `load-conn`. Previously, fields such as IKE `proposals` and child `esp_proposals` / `ah_proposals` could be provided to VICI, but could not be read back from the connection listing API.

With this change, the information submitted through VICI can also be retrieved through VICI, making round-trip config validation possible.

## Changes
- Emit IKE `proposals` in `list-conn` output.
- Emit child `esp_proposals` and `ah_proposals` in `list-conn` output.
- Document the new fields in the VICI README.

## Note
Any feedback is appreciated. I have compiled and tested this locally to the best of my ability.

Looking at the contributing guidelines, I expect a bot will tell me how to proceed with the signed CLA.

### DIFF
--- a/src/libcharon/plugins/vici/README.md
+++ b/src/libcharon/plugins/vici/README.md
@@ -893,6 +893,9 @@ _list-conns_ command.
 			version = <IKE version as string, IKEv1|IKEv2 or 0 for any>
 			reauth_time = <IKE_SA reauthentication interval in seconds>
 			rekey_time = <IKE_SA rekeying interval in seconds>
+			proposals = [
+				<list of configured IKE proposals>
+			]
 
 			local*, remote* = { # multiple local and remote auth sections
 				class = <authentication type>
@@ -921,6 +924,12 @@ _list-conns_ command.
 					rekey_time = <CHILD_SA rekeying interval in seconds>
 					rekey_bytes = <CHILD_SA rekeying interval in bytes>
 					rekey_packets = <CHILD_SA rekeying interval in packets>
+					esp_proposals = [
+						<list of configured ESP proposals>
+					]
+					ah_proposals = [
+						<list of configured AH proposals>
+					]
 					local-ts = [
 						<list of local traffic selectors>
 					]
--- a/src/libcharon/plugins/vici/vici_query.c
+++ b/src/libcharon/plugins/vici/vici_query.c
@@ -209,6 +209,29 @@ static void list_label(vici_builder_t *b, child_sa_t *child, child_cfg_t *cfg)
 }
 
 /**
+ * List proposals for a config
+ */
+static void list_proposals(vici_builder_t *b, linked_list_t *proposals,
+						   char *label, protocol_id_t protocol)
+{
+	enumerator_t *enumerator;
+	proposal_t *proposal;
+
+	b->begin_list(b, label);
+	enumerator = proposals->create_enumerator(proposals);
+	while (enumerator->enumerate(enumerator, &proposal))
+	{
+		if (protocol == PROTO_NONE ||
+			proposal->get_protocol(proposal) == protocol)
+		{
+			b->add_li(b, "%P", proposal);
+		}
+	}
+	enumerator->destroy(enumerator);
+	b->end_list(b);
+}
+
+/**
  * List additional key exchanges
  */
 static void list_ake(vici_builder_t *b, proposal_t *proposal)
@@ -1006,6 +1029,10 @@ CALLBACK(list_conns, vici_message_t*,
 		b->add_kv(b, "unique", "%N", unique_policy_names,
 				  peer_cfg->get_unique_policy(peer_cfg));
 
+		list = ike_cfg->get_proposals(ike_cfg);
+		list_proposals(b, list, "proposals", PROTO_NONE);
+		list->destroy_offset(list, offsetof(proposal_t, destroy));
+
 		dpd_delay = peer_cfg->get_dpd(peer_cfg);
 		if (dpd_delay)
 		{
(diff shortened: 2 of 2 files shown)


bk203:
Ok, I triple-checked my email, unchecked the “private” option, added my real name to my account, and set my primary email to public visibility, but the CLA bot is still failing.

Does anyone know what I might be doing wrong?

Also, right now I’m force-pushing just to retrigger the bot — is there a better way to do this?

tobiasbrunner:
The committer and author email addresses in the commit are different. Make sure it's the same and the one associated with your account.

bk203:
@tobiasbrunner, thank you!

tobiasbrunner:
@cla-bot check 

Thanks. There are few concerns I have with this.

One is that the two `get_proposals()` methods log a `configured proposals: ...` message on level 2. So every `list-conns` request would produce these log messages for every config, which is not ideal. We had a similar issue with the traffic selectors, where at some point there was a flag to suppress the logging. I changed that last year so that the getter really just returns the configured value without logging and the "select" method is used to get the initiator's values during the negotiation. The same approach won't work here because only a single proposal can get selected (unlike traffic selectors), so the return values of the functions don't match up. But I suppose we could add a flag (would also allow suppressing the log message in other places e.g. when installing trap policies or triggering alerts). I've done so in the `3067-proposals` branch.

The next issue is the actual output this produces in the response. This will just be a list of strings, each of the form `<proto>:<transform1>/<transform2>/...`. These are intended for the log, so I'm not sure that's ideal for an API. Unfortunately, VICI currently doesn't allow nested lists or anonymous structures. So doing this differently is a bit tricky. One option could be to put each proposal in a numbered section and likewise each transform in one so that each could contain the type (which is now only guessable e.g. via prefix of the identifier or by knowing the identifiers and mapping them), the identifier (could be numeric if there is no string for it, but maybe the numeric ID could always be provided and the name, in a separate key/value pair, only if available - having the numeric IDs would actually make this more generic as these are standardized unlike the strings, on the other hand, we currently don't have that either for the negotiated proposals) and the optional key size. To avoid encoding the transform type redundantly, the numbered transform sections could be put in named sections (e.g. `encr`, `integ`, `prf`, etc.) within the numbered proposal sections. Alternatively, we could again put each proposal in a numbered section, but instead of sub-sections each would contain lists for each transform type (named as above). Contained in the lists would be the strings produced for the log message, i.e. `<name>[_keysize]` or `UNKNOWN_<numeric id>[_keysize]` (but without the `KE#_` prefix for additional KEs as they would be in their separate lists). That would be more compact than the first suggestion above (sections and key-value pairs add a lot more overhead than lists/list items) but still more structured than just strings and closer to how single proposals for established connections are returned (each algorithm in a key/value pair like `encr-alg`, `integ-alg` etc.). What do you think? How exactly are you using this? How do you process the returned proposals/strings?

Of course, that individual transform strings don't match the configured values is an issue here as well. But that's a general problem with VICI and also affects the output for the negotiated proposals (but as mentioned, that is encoded more structured as each transform gets its own key-value pair - that's obviously easier to do with a single proposal that only has one value per type).

The last thing I want to mention is the `default` keyword, which is the default value for these settings. As that gets expanded when the config is loaded, the user technically won't get back what they configured. However, that could also be an interesting use for this feature as it allows to learn the actual default proposal without having to check the log once the connection is initiated.

bk203:
Thanks for the detailed feedback and insights. I am still fairly new to the project, so my changes were intentionally scoped quite narrowly around this specific use case, and I definitely appreciate the additional context and design considerations you provided.  

The logging concern definitely makes sense. I agree that `list-conns` should not produce `configured proposals` log messages just by introspecting configuration state. The suppression flag approach sounds reasonable to me, especially since there are apparently other contexts where proposal retrieval is needed without logging side effects. I'll checkout the `3067-proposals` branch, to see it for myself.  

Regarding the encoding itself, my primary use case is configuration validation/reconciliation from an external NaaS controller. I intentionally try to keep as little VICI state as possible on my side, so when new configuration is applied I query `list-conn` to determine whether changes actually need to be pushed to VICI again. Because of that, even the current string representation is already very useful, since proposal information is currently not available at all via `list-conn`.  

I am not particularly troubled by additional parsing on my end, but I do understand the concern that the current representation is essentially log-oriented serialization exposed through the API and that it does not necessarily map cleanly back to the original configured input.  

Of the two alternatives you mentioned, I think I slightly prefer the grouped-list approach (named sections like `encr, integ, prf, dh, etc`. containing transform strings). It seems like a good compromise between structure and VICI overhead/limitations, while still being much easier to process reliably than the current `<proto>:<transform>/<transform>/...` strings.  

Regarding `default`, for my use case the expanded/effective proposal is actually useful and probably preferable to the literal keyword, since I am mainly interested in validating the effective configuration currently stored internally.  

I'm also fine if the initial version remains simpler and evolves later, but I agree it's worth thinking about the API shape now before clients start depending on the exact format.

tobiasbrunner:
Thanks for your feedback. I agree that the list approach is probably a good compromise.  I've pushed such a change on top of your (slightly adapted) commit to the `3067-proposals` branch.  Please let me know what you think.

bk203:
Just had the time to review and test the changes in branch `3067-proposals` — this looks good to me.

I think the grouped-list representation is a nice improvement over the original string format and strikes a good balance between structure and VICI's current limitations. It is also straightforward to consume from my side.

Thanks again for the detailed feedback and for taking the time to iterate on the API design and implementation.

tobiasbrunner:
Thanks for reviewing and testing the branch. I've merged the changes to master.

PR #3075, opened 2026-06-01
OSS-Fuzz: Add fuzzer targets ssh key parsing

arthurscchan:
This PR adds a new fuzzer targets sshkeys parsing, with fc and fd plugin lists.

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -11,5 +11,7 @@ fuzz_ocsp_rsp_def
 fuzz_pa_tnc
 fuzz_pb_tnc
 fuzz_radius
+fuzz_sshkey_cus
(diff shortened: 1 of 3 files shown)


tobiasbrunner:
I don't think this fuzzer has to be built twice. There is only the `sshkey` plugin that can parse these keys. The plugin should just be added to one of the lists (probably the custom one) and the Makefile updated accordingly.

Actually, I don't think it's necessary to have this separate fuzzer. This can just be handled by the generic pubkey fuzzer you added in #3078 because the `sshkey` plugin supports `BUILD_BLOB`. See there for more.

arthurscchan:
Yes. This should be closed then.

PR #3076, opened 2026-06-01
OSS-Fuzz: Add fuzzer for pkcs crypto plugin

arthurscchan:
This PR adds 3 fuzzers for different pkcs crypto plugin.


tobiasbrunner:
For PKCS#8, we only need a custom version. Because without `pkcs8` plugin, we'll just fuzz the generic key loading in the `openssl` plugin (i.e. we'd fuzz OpenSSL). For the other two structures, there is more code around the OpenSSL API.

After parsing a valid PKCS#7 or PKCS#12 container, it might make sense to call some of the methods on the `container_t`, `pkcs7_t` and `pkcs12_t` interfaces.

PKCS#7 is a bit tricky as it can be nested and data can be signed and/or encrypted (symmetrically or asymmetrically). So some parts of the code will never be reached without credentials that match such input. Maybe matching input for the corpus could be created with `pki`.

Similarly, both PKCS#8 and PKCS#12 can be encrypted (various algorithms supported), which requires a shared secret to decrypt (I guess we could load one and put some matching input in the corpus).

Also, I was wondering why you didn't use `LLVMFuzzerInitialize` for these fuzzers. Wouldn't it be more efficient to do repeated stuff only once there?

arthurscchan:
I have done the following changes.
1) Add corpora for these pkcs crypto fuzzers in https://github.com/strongswan/fuzzing-corpora/pull/4
2) As PKCS#8 has no fuzzing value with default plugin, I have revamp the build to avoid building the default plguin version of it.
3) I have changed to use fuzzerInitialize to decrease overhead of plugin loading.
4) I have added more logic post parsing for these fuzzers.

Do let me know if you think more should be changes on these fuzzers.

PR #3077, opened 2026-06-01
OSS-Fuzz: Add fuzzer for pgp plugins

arthurscchan:
This PR adds fuzzers for gpg plugins.


tobiasbrunner:
This only needs custom versions as only the `pgp` plugin can load such credentials. And it might be possible to handle this generically with a single fuzzer (see #3078) and `fuzz_cert` (see #3080).

arthurscchan:
Closing this prefer #3078.

PR #3078, opened 2026-06-01
OSS-Fuzz: Add fuzzers for some key plugins

arthurscchan:
This PR adds new fuzzers targeting some key plugins.

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -10,6 +10,10 @@ fuzz_ocsp_rsp_cus
 fuzz_ocsp_rsp_def
 fuzz_pa_tnc
 fuzz_pb_tnc
+fuzz_pubkey_cus
+fuzz_pubkey_def
 fuzz_radius
 fuzz_tls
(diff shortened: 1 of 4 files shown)


tobiasbrunner:
Only the `dnskey` plugin supports DNSKEYs. So that's not required twice. Just add the plugin to `fc`. Unlike `sshkey`, the `dnskey` plugin does not support `BUILD_BLOB`.

However, it might still be possible to handle this with a single "pubkey" fuzzer. Just use the first byte as encoding type  (probably should be a custom mapping that maps to `BUILD_BLOB_DNSKEY`, `BUILD_BLOB_SSHKEY`, `BUILD_BLOB_PGP`, or `BUILD_BLOB`). A second byte could even indicate the actual key type (`KEY_ANY`, `KEY_RSA` etc.), although the code paths might not be that different for a specific type.

arthurscchan:
I have fixed the fuzzers and combine related fuzzers to a single one with less build.

PR #3079, opened 2026-06-01
OSS-Fuzz: Add fuzzer for proposal parsing

arthurscchan:
This PR creates new fuzzer for proposal parsing.


tobiasbrunner:
I don't think this is necessary. This parses config strings (i.e. user input). It's very simple (splits by dashes and does a static lookup) and not security relevant.

arthurscchan:
OK. I will close this.

PR #3080, opened 2026-06-01
OSS-Fuzz: Add fuzzer for acert

arthurscchan:
This PR adds new fuzzer for x509 acert processing.

### DIFF
--- a/fuzz/fuzz_certs.c
+++ b/fuzz/fuzz_certs.c
@@ -17,24 +17,49 @@
 #include <library.h>
 #include <utils/debug.h>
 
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
I wonder if this could not just be handled by `fuzz_certs`. We could extend that to first try parsing the input with `CERT_X509` and if that failed try with `CERT_X509_AC` (and maybe even add `CERT_PGP` as a second fallback so we don't need a separate fuzzer for that either).

arthurscchan:
I do agree. I have now remove the new fuzzers and modify the existing fuzz_certs to accommodate that.

PR #3081, opened 2026-06-01
OSS-Fuzz: Add fuzzer for simaka parsing

arthurscchan:
This PR adds fuzzing for simaka parsing


tobiasbrunner:
Without a `simaka_crypto_t` object this will crash if parsing the message succeeds as there will be an attempt to decrypt it. Since `simaka_crypto_create()` takes a protocol, maybe the first input byte should indicate SIM or AKA (mapped to `EAP_SIM` or `EAP_AKA`).

Another crash will be caused because there is no SIM/AKA manager when `call_hook()` is called. These could be created and registered with something like this:

```c
simaka_manager_t *mgr = simaka_manager_create();
lib->set(lib, "sim-manager", mgr);
mgr = simaka_manager_create();
lib->set(lib, "aka-manager", mgr);
```

Both will be required as even with the flag above, the necessary manager depends on the header of the parsed message, which might not match.

For the crypto to work, a key derivation would be necessary, which technically requires a SIM card that provides triplets/quintuplets. But I guess `derive_keys_full()` could be called with some static input. On the other hand, it's very unlikely to be successful unless the corpus contained some matching input that can actually get decrypted.

arthurscchan:
Have fixed accordingly.

PR #3093, opened 2026-06-11
fix: Select the CA matching the 4th packet and place it into the 5th …

w348917281:
Match against the CA from the certreq in the 4th packet, and place only the matching corresponding CA in the 5th packet. This avoids injecting all CAs, which would cause validation failures on third-party vendors due to extraneous CA entries.


tobiasbrunner:
First, why are you still using IKEv1?

Anyway, I don't think this is a good idea. The CA required to verify the responder's certificate might not be the same as the one that issued the initiator's certificate, so only sending a CERTREQ for the latter is not necessarily correct. If you need to send/enforce a specific CA for the responder, configure it as such on the initiator, then only that will be sent (or disable sending certificate requests if they cause problems, they are optional). Feel free to apply the patch locally, but I don't think it makes sense to upstream it.

PR #3126, opened 2026-08-12
Couple of test suite fixes specific to windows

djlucas:
Looks like Linux and Mac clients accepted the malformed header but WinHTTP rejected it. Simple fix for that in src/libstrongswan/tests/suites/test_fetch_http.c

Next was Windows filesystem permissions. By default, unprivileged users cannot list files in the windows temp directory. There really isn't a reason to try and use this path. I suppose one could get the user's temp directory and use that, but there is really no reason not to just do locally.

Both tests should be unaffected on Linux/Mac.


djlucas:
Sorry about the mess, I was working across 2 machines and didn't switch branches. I'll close and reopen so you need not see that mess.

Issue #3134, opened 2026-08-27
[charon-nm / nm-strongswan] Add EAP-TTLS support to NetworkManager plugin


wyllianbs:
### Summary

The NetworkManager strongSwan plugin (`network-manager-strongswan`) does not expose EAP-TTLS as an authentication method in its GUI. Many enterprise/university VPN gateways (e.g., using FreeRADIUS) require EAP-PEAP or EAP-TTLS with MSCHAPv2 as the inner method. Since `charon-nm` and the NM editor plugin only offer "EAP" (which maps to direct EAP-MSCHAPv2), these servers reject the connection.

### Current behavior

1. The `method-combo` in `nm-strongswan` offers: EAP (Username/Password), Certificate, EAP-TLS, Pre-shared key
2. Selecting "EAP" sends `method=eap`, which triggers direct EAP-MSCHAPv2
3. The server requests EAP-PEAP, `charon-nm` responds with EAP_NAK, server falls back to EAP-MD5, which fails

### Workaround

Using `charon-systemd` + `swanctl` with `auth = eap-ttls` works perfectly — the `eap-ttls` and `eap-mschapv2` plugins handle the tunnel and inner auth correctly. However, this bypasses NetworkManager entirely, losing desktop integration (system tray status, automatic connect, etc.).

### Proposed change

1. **nm-strongswan editor plugin**: Add "EAP-TTLS" as a 5th option in `method-combo`, saving `method=eap-ttls`. The UI would show Username/Password fields (same as current "EAP" option).
2. **charon-nm**: Recognize `method=eap-ttls` and configure the connection to use the `eap-ttls` plugin with MSCHAPv2 as the inner method.

### Environment

- Debian 13 (trixie), strongSwan 6.0.1, network-manager-strongswan 1.6.2
- Server: vpn.ufsc.br (IKEv2, EAP-PEAP/EAP-TTLS + MSCHAPv2)

wyllianbs:
Here's a draft patch implementing this feature. The change spans two components:

### 1. charon-nm (`src/charon-nm/nm/nm_service.c`)

Three changes:
- `add_auth_cfg_pw()`: when `method=eap-ttls`, add `AUTH_RULE_EAP_TYPE = EAP_TTLS` so strongSwan negotiates EAP-TTLS (with NAK for unsupported methods like EAP-PEAP) instead of accepting whatever the server proposes
[...]

PR #3135, opened 2026-08-27
charon-nm: Add EAP-TTLS authentication method

wyllianbs:
## Summary

Add support for the `eap-ttls` method string in the NetworkManager IKEv2 backend (`charon-nm`), mapping it to `AUTH_RULE_EAP_TYPE=EAP_TTLS` instead of plain EAP.

## Reason for the change

Several university and enterprise VPN gateways (e.g. UFSC's `vpn.ufsc.br`) require IKEv2 with an **EAP-TTLS** outer method wrapping an **MSCHAPv2** inner authentication, distinct from plain EAP. `charon-nm` currently only exposes a generic `eap` method (which maps to `eap-mschapv2`), so NetworkManager users with such gateways have no way to configure the connection and must fall back to manual `swanctl` configuration.

This is part of a 3-layer patch set — the other two layers add the corresponding UI option to the desktop VPN editors:

- **plasma-nm** (KDE): ✅ already merged, see [plasma/plasma-nm!636](https://invent.kde.org/plasma/plasma-nm/-/merge_requests/636), targeting Plasma 6.8
- **network-manager-strongswan** (GTK, GNOME/XFCE/etc): patch ready, not yet submitted

All patches, plus build instructions for any desktop environment, are consolidated at https://github.com/wyllianbs/vpn-eap-ttls-linux while reviews are pending. This continues the discussion from #3134.

## Changes

- `src/charon-nm/nm/nm_service.c`: recognize `method == "eap-ttls"` and set `AUTH_RULE_EAP_TYPE` to `EAP_TTLS` on the peer's auth config; treat `eap-ttls` the same as `eap`/`psk` for username/password handling.

## Test plan

- Built strongSwan with `--enable-nm` on Debian 13 (trixie)
- Configured a NetworkManager IKEv2 connection with `method=eap-ttls`, username and password
- Verified the connection successfully negotiates EAP-TTLS/MSCHAPv2 against a real gateway (UFSC university VPN)

### DIFF
--- a/src/charon-nm/nm/nm_service.c
+++ b/src/charon-nm/nm/nm_service.c
@@ -789,6 +789,10 @@ static bool add_auth_cfg_pw(NMStrongswanPluginPrivate *priv,
 	auth = auth_cfg_create();
 	auth->add(auth, AUTH_RULE_AUTH_CLASS,
 			  streq(method, "psk") ? AUTH_CLASS_PSK : AUTH_CLASS_EAP);
+	if (streq(method, "eap-ttls"))
+	{
+		auth->add(auth, AUTH_RULE_EAP_TYPE, EAP_TTLS);
+	}
 	/* in case EAP-PEAP or EAP-TTLS is used we currently accept any identity */
 	auth->add(auth, AUTH_RULE_AAA_IDENTITY,
 			  identification_create_from_string("%any"));
@@ -1044,6 +1048,7 @@ static gboolean connect_(NMVpnServicePlugin *plugin, NMConnection *connection,
 		}
 	}
 	else if (streq(method, "eap") ||
+			 streq(method, "eap-ttls") ||
 			 streq(method, "psk"))
 	{
 		if (!add_auth_cfg_pw(priv, vpn, peer_cfg, err))
@@ -1248,6 +1253,7 @@ static gboolean need_secrets(NMVpnServicePlugin *plugin, NMConnection *connectio
 			}
 		}
 		else if (streq(method, "eap") ||
+				 streq(method, "eap-ttls") ||
 				 streq(method, "psk"))
 		{
 			need_secret = !nm_setting_vpn_get_secret(settings, "password");


tobiasbrunner:
This seems unnecessary. If the server requests EAP-TTLS and the plugin is loaded it will be used.

wyllianbs:
Thanks for taking a look! I actually tested this exact scenario before opening the PR, and wanted to share the result rather than assume.

Against the same gateway (`vpn.ufsc.br`, UFSC's IKEv2 VPN), with an unpatched `eap-ttls` plugin loaded on the client:

- Selecting the generic **"EAP"** method (plain, no `AUTH_RULE_EAP_TYPE` forced) → connection **fails to activate**.
- Selecting **"EAP-TTLS"** (which requires this patch to set `AUTH_RULE_EAP_TYPE=EAP_TTLS`) → connection **activates successfully**.

So in practice, at least against this gateway, charon-nm does not appear to auto-negotiate EAP-TTLS from the server's EAP-Request when the method is left as plain "eap" — the client seems to need the type pinned explicitly. I don't have visibility into charon's internals to say why (maybe something about how the NM backend builds the peer config vs. the swanctl backend, where I never needed this pinning), but I can share `charon-nm` debug logs for both attempts if that would help pin down whether this is a bug elsewhere (in which case fixing that would be preferable to this patch) or expected behavior for the NM backend specifically.

Happy to test any suggested config too.

tobiasbrunner:
> * Selecting the generic **"EAP"** method (plain, no `AUTH_RULE_EAP_TYPE` forced) → connection **fails to activate**.

I suggest you read the logs and determine why that is.

wyllianbs:
I dug deeper and reproduced this with full charon-nm debug logging (also independently confirmed on a fresh Ubuntu 26.04 LTS live image, unrelated to my own machine's setup). This makes the actual mechanism clear.

**With the unpatched charon-nm and the connection's `method` still set to `eap-ttls`:** the connection fails instantly, before any IKE packet is even sent:
```
05[CFG] using gateway identity 'vpn.ufsc.br'
<warn> vpn[...]: failed to connect: 'Configuration parameters missing.'
```
This matches the code — `nm_service.c`'s method dispatch (`cert`/`eap-tls`/`key`/`agent`/`smartcard` vs `eap`/`psk` vs else-error) doesn't recognize `"eap-ttls"` at all without this patch, so it's rejected before any negotiation.

**With the unpatched charon-nm and `method` set back to the generic `eap`** (which *is* recognized), the connection gets much further — full IKE_SA_INIT, cert validation, IKE_AUTH — but then:
```
07[IKE] server requested EAP_IDENTITY (id 0x00), sending 'wyllian.bs@ufsc.br'
14[IKE] server requested EAP_PEAP authentication (id 0x01)
14[IKE] EAP method not supported, sending EAP_NAK
13[IKE] server requested EAP_MD5 authentication (id 0x02)
[...responds with password...]
16[IKE] received EAP_FAILURE, EAP authentication failed
```

So in practice, this gateway's EAP cascade is: **EAP-PEAP first** (NAK'd, no eap-peap plugin loaded), then **EAP-MD5** (accepted by the loaded `eap-md5` plugin, but authentication then fails since the account's credentials are MSCHAPv2-inside-TTLS, not compatible with plain MD5-Challenge). The server's automatic cascade never reaches EAP-TTLS in this flow — it fails out at EAP-MD5 first.

With this patch (forcing `AUTH_RULE_EAP_TYPE=EAP_TTLS`), the client explicitly requests/accepts only EAP-TTLS from the start, and the connection succeeds.

So at least against this specific gateway, the plugin being loaded is not sufficient — the server doesn't offer EAP-TTLS as part of its default negotiation order, so the client needs to explicitly pin the desired EAP type to get there. Happy to provide the full debug log if useful, or test any other configuration you'd like to see.

tobiasbrunner:
> So in practice, this gateway's EAP cascade is: **EAP-PEAP first** (NAK'd, no eap-peap plugin loaded)

Then load that plugin to use that method (EAP-PEAP with EAP-MSCHAPv2 inside is very similar to EAP-TTLS with EAP-MSCHAPv2 inside).

> then **EAP-MD5** (accepted by the loaded `eap-md5` plugin, but authentication then fails since the account's credentials are MSCHAPv2-inside-TTLS, not compatible with plain MD5-Challenge)

Then the server's behavior makes no sense. Why would it request a method that's not suitable for this client's EAP-Identity?

Also, you can control which EAP methods are sent in what order in the EAP-Nak by configuring the load order for EAP plugins (via `charon-nm.plugins.<eap-...>.load` in [strongswan.conf](https://docs.strongswan.org/docs/latest/plugins/pluginLoad.html#_runtime_plugin_configuration)). So e.g. set `charon-nm.plugins.eap-md5.load` to `no` to avoid the outcome above. Or you could increase the priority of the `eap-ttls` plugin by setting `charon-nm.plugins.eap-ttls.load` to a value larger than `1` so it is proposed first in the EAP-Nak (will depend on the server if it adheres to its own ordering or the client's).

wyllianbs:
Follow-up after further testing — this patch is not needed, you were right. Sharing the full picture for the record.

**Test 1 (Ubuntu 26.04, unpatched charon-nm):** installing `libcharon-extra-plugins`/`libstrongswan-extra-plugins` (which include `eap-peap`, missing by default) was enough. With plain `method=eap`:
```
server requested EAP_PEAP authentication (id 0x01)
EAP method EAP_PEAP succeeded, MSK established
CHILD_SA established
```

**Test 2 (Debian 13, unpatched charon-nm, no eap-peap plugin available on Debian):** setting `charon-nm.plugins.eap-md5.load = no` in `/etc/strongswan.d/charon/eap-md5.conf` was enough. With plain `method=eap`:
```
server requested EAP_PEAP authentication (id 0x01)
EAP method not supported, sending EAP_NAK
server requested EAP_GTC authentication (id 0x02)
EAP method EAP_GTC succeeded, no MSK established
IKE_SA established
CHILD_SA established
```

So the account isn't actually locked to EAP-TTLS specifically — it authenticates through whatever compatible EAP type the negotiation lands on (PEAP, GTC, presumably TTLS too), as long as the client doesn't get stuck accepting an incompatible one (EAP-MD5, which was the fallback that plain "eap" kept accidentally landing on) before reaching one that works. That's a client-side EAP method availability/priority issue, not something charon-nm needs new code for — the existing `eap-*.load` config already covers it.

I'll close this PR. Thanks for pushing back and for the pointer to the load-order option — it was the right call.

Issue #3138, opened 2026-09-08
kernel-netlink: get_route() ignores the routing policy database (ip rule) and dumps all routing tables on every lookup


sithglan:
**System (please complete the following information):**
 - OS: Debian 13.6 (amd64)
 - Kernel version: 6.12.107 (production host), 7.1.12 (reproduction)
 - strongSwan version(s): 6.0.1 (Debian 6.0.1-6+deb13u7) in production, reproduced on current `master` (38c8a56ca, 6.1.0dev)
 - Tested/confirmed with the latest version: yes

**Describe the bug**

`get_route()` in `src/libcharon/plugins/kernel_netlink/kernel_netlink_net.c` performs its own longest-prefix match over a dump of *all* routing tables and never consults the routing policy database (`ip rule`). On a router that keeps a full BGP feed in a policy-selected table, charon picks a source address the kernel would never use for that destination, so `IKE_SA_INIT` is sent from the wrong address and no tunnel can be established.

The same code path also dumps the entire RIB (1.09 million routes here) on every single source-address lookup. Under a retransmit loop this drives charon into a complete lockup.

Impact on the affected host: complete and silent loss of all IPsec connectivity, not self-healing, requiring manual intervention. It stayed unnoticed for three hours, because the daemon still reported `active (running)` to systemd throughout.

---

### Part 1 — the route lookup ignores the routing policy database

The affected host is a router. Its WAN is a PPPoE link, and it carries a full BGP feed from an upstream that is reachable over an unnumbered tunnel interface (`vultr`, IPv4 link-local 169.254.0.6/31). That feed is kept in table 100, which the RPDB restricts to one prefix:

```
# ip rule show
0:      from all lookup local
220:    from all lookup 220
32764:  from all to 100.64.0.0/24 lookup 100
32765:  from 100.64.0.0/24 lookup 100
32766:  from all lookup main
32767:  from all lookup default

# ip route show table main | wc -l
57
# ip route show table 100 | wc -l
1089264
```

The main table has a single IPv4 default route over the PPPoE link:

```
# ip route show table main | grep default
default dev ppp0 scope link
```

The kernel therefore resolves the IPsec peers correctly:

```
# ip route get 203.0.113.222
203.0.113.222 dev ppp0 src 198.51.100.41
# ip route get 192.0.2.191
192.0.2.191 dev ppp0 src 198.51.100.41
```

[...]

PR #3121, opened 2026-08-09
fix: avoid full routing table dumps for netlink route lookups

sithglan:
Only request NLM_F_DUMP in get_route() when the extra filtering is actually needed, otherwise use a plain RTM_GETROUTE lookup. With a full BGP table in the kernel, every route lookup previously dumped ~1.3M netlink messages, overflowing the receive buffer, triggering endless re-synchronization dumps, and stalling IKE/CHILD_SA establishment (100-200% CPU, "peer not responding"). A plain lookup returns the single route the kernel would use and honors policy routing rules.

The dump is retained for subnet lookups, excluded tables, virtual IPs, and regular IPsec routes (new has_ipsec_routes() helper); throw routes are skipped since lookups fall through them anyway.

PR #3139, opened 2026-09-08
kernel-netlink: Ignore source addresses with a smaller scope than the destination

sithglan:
`get_matching_address()` returns the first usable address on an interface even if its scope is too small for the destination. `is_address_better()` implements RFC 6724 rule 2 (prefer appropriate scope), but it is only consulted once an interface offers more than one candidate — so an interface that has nothing but a link-local address yields that address for global destinations.

On a router with an unnumbered tunnel interface (IPv4 link-local `169.254.0.6/31`), `get_route()` picks a route over that interface, finds no `RTA_PREFSRC`, falls through to `get_interface_address()` and gets `169.254.0.6`. charon then sends `IKE_SA_INIT` from a link-local address to a global unicast peer and no tunnel can be established. The kernel would never select such a source address itself.

With this change the address is skipped, `get_route()` continues with the next matching route, and the source address ends up the same one the kernel uses.

This is the secondary issue described in #3138. It is independent of #3121 and does not fix the underlying problem there (the userspace route lookup ignoring `ip rule`), but it removes a clearly wrong source address from the candidate set in the cases where the dump-based lookup is still used.

### Testing

Reproduction in a network namespace, where `wan` carries the default route, `tun0` is an unnumbered tunnel with only a link-local address, and table 100 holds a more specific route to the peer while the RPDB restricts that table to an unrelated prefix:

```sh
ip netns add swtest
ip -n swtest link set lo up
ip -n swtest link add wan  type dummy
ip -n swtest link add tun0 type dummy
ip -n swtest link set wan up
ip -n swtest link set tun0 up
ip -n swtest addr add 198.51.100.2/24 dev wan
ip -n swtest addr add 169.254.0.6/31  dev tun0
ip -n swtest route add default via 198.51.100.1 dev wan
ip -n swtest route add 203.0.113.0/24 dev tun0 table 100
ip -n swtest rule  add to 192.0.2.0/24 lookup 100 pref 32764
```

The kernel's own answer:

```
# ip netns exec swtest ip route get 203.0.113.5
203.0.113.5 via 198.51.100.1 dev wan src 198.51.100.2 uid 0
```

Initiating a connection to `203.0.113.5` with no `local_addrs`, before:

```
[IKE] initiating IKE_SA test[1] to 203.0.113.5
[NET] sending packet: from 169.254.0.6[500] to 203.0.113.5[500] (972 bytes)
[IKE] retransmit 1 of request with message ID 0
[NET] sending packet: from 169.254.0.6[500] to 203.0.113.5[500] (972 bytes)
```

after:

```
[IKE] initiating IKE_SA test[1] to 203.0.113.5
[NET] sending packet: from 198.51.100.2[500] to 203.0.113.5[500] (972 bytes)
[IKE] retransmit 1 of request with message ID 0
[NET] sending packet: from 198.51.100.2[500] to 203.0.113.5[500] (972 bytes)
```

A link-local destination still gets the link-local source, so reaching a peer on the unnumbered link keeps working:

```
[IKE] initiating IKE_SA test[1] to 169.254.0.7
[NET] sending packet: from 169.254.0.6[500] to 169.254.0.7[500] (972 bytes)
```

`make check` passes.

### DIFF
--- a/src/libcharon/plugins/kernel_netlink/kernel_netlink_net.c
+++ b/src/libcharon/plugins/kernel_netlink/kernel_netlink_net.c
@@ -921,6 +921,12 @@ static host_t *get_matching_address(private_kernel_netlink_net_t *this,
 	iface_entry_t *iface;
 	addr_entry_t *addr, *best = NULL;
 	bool candidate_matched = FALSE;
+	u_char scope_dest = 0;
+
+	if (dest)
+	{
+		scope_dest = get_scope(dest);
+	}
 
 	ifaces = this->ifaces->create_enumerator(this->ifaces);
 	while (ifaces->enumerate(ifaces, &iface))
@@ -933,6 +939,12 @@ static host_t *get_matching_address(private_kernel_netlink_net_t *this,
 				if (addr->refcount ||
 					addr->ip->get_family(addr->ip) != family)
 				{	/* ignore virtual IP addresses and ensure family matches */
+					continue;
+				}
+				if (dest && get_scope(addr->ip) < scope_dest)
+				{	/* ignore addresses of a smaller scope than the destination,
+					 * e.g. link-local addresses for global destinations, which
+					 * the kernel wouldn't use as source address either */
 					continue;
 				}
 				if (net.ptr && !host_in_subnet(addr->ip, net, mask))


sithglan:
Hello,

> Thank you for your pull request and welcome to our community. We
> require contributors to sign our Contributor License Agreement, and we
> don't seem to have the users @sithglan on file. In order for us to
> review and merge your code, please contact the project maintainers via
> ***@***.*** to get yourself added.

I already signed it. See:

https://github.com/strongswan/strongswan/pull/3121

Cheers,
        Thomas

tobiasbrunner:
> I already signed it

You have not! That's not how this works. Please read the instructions.

PR #3140, opened 2026-09-08
nm: Four independent fixes for charon-nm and the editor

ggiesen:
Four unrelated fixes in charon-nm and the NM editor. They do not depend on each
other -- take or drop individually.

- **Wipe secrets before freeing them.** The user password and private key
  passphrase were freed without being cleared.

- **Keep connection data items the editor has no widget for.**
  `update_connection()` rebuilds the VPN setting from the widgets, so anything
  else is silently dropped on save. Not hypothetical: `remote-ts` was readable by
  charon-nm from 2020 and only gained a field in 2025, so for that whole period an
  nmcli-configured connection lost its traffic selectors the first time anyone
  opened it in the GUI.

- **Add a per-connection `dpd-delay`.** `peer_cfg_create_t.dpd` is left
  zero-filled, so no NM connection runs a periodic liveness check despite the
  child config setting `dpd_action`; a gateway that vanishes without a DELETE is
  only noticed at the next rekeying. `charon-nm.check_current_path` does not cover
  this -- it fires once from the roam handler, catching the client moving rather
  than the gateway going away. Unset gives 0, so nothing changes unless
  configured. Keyfile/nmcli only, as `remote-ts` was.

- **Make `dispose()` idempotent.** It dereferenced `priv->builder` before checking
  it, and left both pointers set after unreffing them.

Tested on Ubuntu 24.04 and built against master. `dpd-delay` was verified in
isolation on an otherwise pristine 5.9.13 tree: with `dpd-delay=10` on a
`method=eap` connection DPD fires every 10 s; with the item absent, nothing is
sent. The editor fix was verified by driving `update_connection()` directly.

### DIFF
--- a/src/charon-nm/nm/nm_creds.c
+++ b/src/charon-nm/nm/nm_creds.c
@@ -348,13 +348,25 @@ METHOD(nm_creds_t, load_ca_dir, void,
 	}
 }
 
+/**
+ * Wipe and free a secret string
+ */
+static void wipe_and_free(char *secret)
+{
+	if (secret)
+	{
+		memwipe(secret, strlen(secret));
+		free(secret);
+	}
+}
+
 METHOD(nm_creds_t, set_username_password, void,
 	private_nm_creds_t *this, identification_t *id, char *password)
 {
 	this->lock->write_lock(this->lock);
 	DESTROY_IF(this->user);
 	this->user = id->clone(id);
-	free(this->pass);
+	wipe_and_free(this->pass);
 	this->pass = strdupnull(password);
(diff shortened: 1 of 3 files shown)


ggiesen:
These components are DCO-only per the [contribution requirements](https://docs.strongswan.org/docs/latest/devs/contributions.html) -- "NM plugin" (`src/frontends/gnome`) and "charon-nm" (`src/charon-nm`) are both listed as GPL-only. This PR touches only those two directories and all four commits are signed off, but happy to sign a CLA if you'd prefer.

Issue #3143, opened 2026-09-14
Android client choses IP family initially, gets "stuck" when that server/family becomes unreachable


eqvinox:
**System:**
 - OS: Android 17
 - strongSwan version(s): 6.0.3
 - Android strongSwan version(s): 2.6.3
 - Tested/confirmed with the latest version: yes

**Describe the bug**

The Android StrongSwan app gets stuck on the initial IP address family used to connect, even if it goes away (e.g. when roaming Wifi ↔ Mobile).

**To Reproduce**

1. have an IPsec server that supports both IPv4 and IPv6, and has both on the same hostname
2. configure that endpoint in the Android app, using the hostname
3. connect to the VPN with IPv6 available
4. roam to a network without IPv6

alternative:

1. have an IPsec server with only IPv4
1a. have an IPv6-mostly network (DHCP option 108 + NAT64 available and advertised in PREF64)
2. configure that endpoint in the Android app, using the hostname
3. connect to the VPN with IPv6 available
3a. due to NAT64, an IPv6 GUA will be used for the client, and the server will be on `64:ff9b::<ipv4>`
4. roam to a network without IPv6

**Expected behavior**

The IPsec client should switch to IPv4 (either with MOBIKE or a full reconnect).

**Actual behavior**

It keeps trying and failing on IPv6 indefinitely.

**Logs/Backtraces**

```
Sep 14 18:03:22 00[DMN] +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
Sep 14 18:03:22 00[DMN] Starting IKE service (strongSwan 6.0.3, Android 17 - CP2A.260805.005/2026-08-05, Pixel 8 - google/shiba/Google, Linux 6.1.157-android14-11-gbd23337e42e7-ab14791245, aarch64, org.strongswan.android)
Sep 14 18:03:22 00[LIB] providers loaded by OpenSSL: default legacy
Sep 14 18:03:22 00[LIB] loaded plugins: androidbridge charon android-log socket-default openssl nonce pkcs1 pem x509 xcbc kdf revocation eap-identity eap-mschapv2 eap-md5 eap-gtc eap-tls
Sep 14 18:03:22 00[JOB] spawning 16 worker threads
Sep 14 18:03:22 13[IKE] initiating IKE_SA android[64] to 2a07:server6net::7261
[...]

PR #3148, opened 2026-09-16
kernel_netlink: allow global IP addrs on loopback

eqvinox:
Linux supports 'global' IPv4/v6 addresses on loopback, and this is used with both routers (where addresses may be assigned to the system as a whole) as well as "L3 VM connectivity" setups.  Especially with IPv4 addresses being convertible into money these days...

This was already working fine, but the addresses were not getting listed in ADDR_*_ADD.  Just ditch the check that was preventing these addresses from being enumerated.

---
cf. #3143


eqvinox:
Superseded by https://github.com/strongswan/strongswan/commit/7627fd3cb4202e2370f3de886bc756f4bf793ef4

PR #3149, opened 2026-09-18
libtpmstss: Skip zero-padding of r and s when preparing EC signature

rshearman:
According to the TPM 2.0 reference implementation the r and s parameters of an EC signature are placed into a buffer sized according to the curve, and thus are padding with leading zeroes.

However, DER requires a minimal encoding, and as such there was a 1 in 256 chance of creating a signature that would not successfully validate.

Fix this by skipping the leading zeros when encoding an EC signature as DER.

Tested by doing the following with a suitablly set up swtpm and CA and leaf keys and certs:
```shell
for i in $(seq 1 500); do
	setsid -w ./test/bin/pki --issue --cakeyid 81010005 --cacert test/ca.der --in test/leaf.pub \
		--dn "CN=leaf-$i" --digest sha384 --outform der </dev/null 2>/dev/null \
		> test/leaf.der
	out=$(./test/bin/pki --verify --in test/leaf.der --cacert test/ca.der </dev/null 2>&1)
	grep -q "certificate trusted, lifetimes valid" <<<"$out" || echo "iter $i: VERIFY FAILED"
done
```
Without the fix this failed for two iterations on average and with the fix all iterations are successful.


tobiasbrunner:
Thanks. What the reference implementation does is correct, according to the [TPM 2.0 Library spec](https://trustedcomputinggroup.org/resource/tpm-library-specification/) (Part 1, 44.5.3).

> However, DER requires a minimal encoding, and as such there was a 1 in 256 chance of creating a signature that would not successfully validate.

Yeah, for each integer, there is roughly a 1/256 chance the first byte is 0 and a 1/2 chance that's the case for the second byte's MSB (otherwise, the resulting encoding was correct by accident). Whether the verification of such a signature fails depends on the implementation (e.g. the `botan` plugin parses the DER-encoded signature and skips zero bytes for each integer, but neither the `openssl`, nor the `wolfssl` plugins do that as they just pass the whole signature blob, so it further depends on what these libraries do).

I see that you basically copied this from the `pkcs11` plugin (including the comment), so there is not much to complain about the code itself. I just moved the second `asn1_integer()` line, so the code looks even closer to that. And I added the same fix to the `v1` version of the implementation.

I also changed the commit message a bit to reference the actual specs, explain the probability, fix a typo in the subsystem name, and add `Fixes:` tags (`3149-tpm-ecdsa-pad` branch). Please let me know if that's OK with you.

rshearman:
Thank you! I've taken a look at the the changes in the branch and all good with me, with just a comment on the spec reference - I did look hard at the spec to try to find a reference for the padding of the r and s values from ECDSA signing, but I wasn't able to find a specific reference. TPM 2.0 Library Spec, Part 1, 44.5.3 ("Padding"), as you reference, is the closest but the first part is specific to ECC points:

> In ECC points returned by the TPM, the x and y values, if non-empty, are required to be the size of their
associated curve (e.g., 32 bytes for NIST P-256)

However, Part 2, 11.3.2 ("TPMS_SIGNATURE_ECC") specifies that signatureR and signatureS are of type  TPM2B_ECC_PARAMETER.

Then later on in Part 1, 44.5.3 ("Padding"), it says:

> When the ECC parameters are returned by the command TPM2_ECC_Parameters(), the numeric values will
be as specified in the TCG Algorithm registry [22]. However, the size of values, other than the x and y of the
generator point, may not be the same as the registry because values may or may not be zero padded.

These aren't ECC parameters returned by TPM2_ECC_Parameters, so it doesn't apply.

However, there is a catch-all of Part 1, 2.5 ("Reference Code") which says:

> A TPM need not be implemented using the Reference Code. However, any implementation should provide
equivalent or, in most cases, identical results as observed at the TPM interface or demonstrated through
evaluation.

So anything that falls between the cracks of the spec is resolved by doing whatever the reference code says, which I'd argue is the case here.

I'm happy either way (the most important thing is that the bug gets fixed), but just wanted to make sure you had all the information to make the call on the spec reference in the commit message.

tobiasbrunner:
Thanks, those are good points. I guess I concluded too quickly that because the type is `TPM2B_ECC_PARAMETER` and section 44.5.3 covers the "padding requirements for ECC parameters", this would apply to signatures as well. But in particular the paragraph about `TPM2_ECC_Parameters()` as well as table 44, which just states signatures are "contained in structure with two sized buffers", indicate the padding might not necessarily be there. But if that was the case, the behavior for RFC4754-style signatures would not always be correct either (`SIGN_ECDSA_256/384/521`) as we do [require zero-padding](https://datatracker.ietf.org/doc/html/rfc4754#section-7) there.

So let's just hope actual TPMs mirror [the reference implementation](https://github.com/TrustedComputingGroup/TPM/blob/3b1f0e623f244abe2015df9e3ba86de5c9118524/TPMCmd/tpm/src/crypt/CryptEccSignature.c#L82-L93). But I wonder where the padding actually comes from (there are no comments and the file history isn't useful either). Since section 44.3.1 in the spec refers to ISO/IEC 14888-3, I tried to have a look to see if that clarifies things. But while it defines the signature as concatenation of R and S, which only really works for a verifier if they are the same length, it does not seem to spell this out explicitly. In the (informative) Annex G.2 it states that the signature size for ECDSA is 2β bits (β is the length of q) and S in example 3 (F.6.4) is actually zero-padded. So not really conclusive. Anyway, the commit message should be updated.

However, I wonder whether a simpler fix for this would be to just call `chunk_skip_zero()` from `asn1_integer()`. I feel passing a zero-padded value to produce a non-minimal encoding is never intended. So that seems like the right approach to fix this and avoid such issues in the future. I've pushed this to the `3149-asn1-int-pad` branch.

rshearman:
The commits on `3149-asn1-int-pad` look good and pass my pki test. I can also confirm that a real TPM (a Nuvoton NPCT) also pads the ECDSA r/s parameters with leading zeroes.

